PR #89, opened 2020-09-30
expose KSType.isMarkedNullable API

neetopia:


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSErrorType.kt
```diff
@@ -52,6 +52,8 @@ object KSErrorType : KSType {
         return this
     }
 
+    override val isMarkedNullable: Boolean = false
```
ting-yuan, line 55: Just saying, maybe we should move away from the singleton for cases `NotExistingType?` **in the future**.

ting-yuan (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSErrorType.kt
neetopia: I think one point of being singleton is that ErrorType doesn't need to be specified as nullable or non-nullable since they are unresolvable, can you give a use case where nullability matters for error types?

PR #90, opened 2020-09-30
add createKSTypeReferenceFromKSType API to Resolver

neetopia:


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -213,6 +214,10 @@ class ResolverImpl(
         return KSNameImpl.getCached(name)
     }
 
+    override fun createKSTypeReferenceFromKSType(type: KSType): KSTypeReference {
+        return KSTypeReferenceSyntheticImpl(type)
```
ting-yuan, line 218: Why not getCached?

neetopia: Agh.. totally forgot it.

ting-yuan (approved):

PR #92, opened 2020-09-30
Update contribution guide to not use TestGenerator

neetopia:


CONTRIBUTING.md
```diff
@@ -57,11 +57,7 @@ it should be extending [AbstractTestProcessor](compiler-plugin/src/main/kotlin/c
         * Then follows virtual files section till the end of test file.
         * You can use ```// FILE: <file name>``` to create files that will be available at run time of the test.
             * E.g. ```// FILE: a.kt``` will result in a file named ```a.kt``` at run time.
-* Generate test using gradle.
-    * After you have finished writing your test file, you can generate the test case in TestSuite by running 
-    ```generateTests```gradle task. It can take a while. 
-    * After generating, make sure there is no other tests generated by checking git status, you should include only tests in 
-    [KotlinKSPTestGenerated](compiler-plugin/src/test/kotlin/com/google/devtools/ksp/test/KotlinKSPTestGenerated.java) in your PR.
+* Add new test to [test suite](compiler-plugin/src/test/java/com/google/devtools/ksp/test/KotlinKSPTestGenerated.java)
 * Run generated tests with ```:compiler-plugin:test``` gradle task.
```
ting-yuan, line 61: Can you also add a reminder to `-PcompilerTestEnabled`?

ting-yuan (approved):

PR #100, opened 2020-10-02
Fix Java class constructor calculation.

neetopia:
* Add support for default constructor of Java class.
* Fixed getDeclaredFunctions which will return duplicate elements for
  all constructors.

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
@@ -64,7 +64,8 @@ fun KSClassDeclaration.getDeclaredProperties(): List<KSPropertyDeclaration> {
 }
 
 fun KSClassDeclaration.getConstructors(): List<KSFunctionDeclaration> {
-    return this.getDeclaredFunctions().filter { it.simpleName.asString() == this.simpleName.asString() }
+    val functions = if (this.origin == Origin.JAVA) this.getAllFunctions() else this.getDeclaredFunctions()
+    return functions.filter { it.simpleName.asString() == this.simpleName.asString() || it.simpleName.asString() == "<init>"}
             .let { constructors ->
                 this.primaryConstructor?.let { constructors.plus(it) } ?: constructors
             }
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSClassDeclarationJavaImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSClassDeclarationJavaImpl.kt
@@ -78,6 +78,7 @@ class KSClassDeclarationJavaImpl private constructor(val psi: PsiClass) : KSClas
             it.unsubstitutedMemberScope.getDescriptorsFiltered(DescriptorKindFilter.FUNCTIONS)
                 .toList()
                 .filter { (it as FunctionDescriptor).visibility != Visibilities.INVISIBLE_FAKE }
+                .plus(it.constructors)
                 .map { (it as FunctionDescriptor).toKSFunctionDeclaration() }
         } ?: emptyList()
     }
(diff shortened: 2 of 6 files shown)


api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -64,7 +64,8 @@ fun KSClassDeclaration.getDeclaredProperties(): List<KSPropertyDeclaration> {
 }
 
 fun KSClassDeclaration.getConstructors(): List<KSFunctionDeclaration> {
-    return this.getDeclaredFunctions().filter { it.simpleName.asString() == this.simpleName.asString() }
+    val functions = if (this.origin == Origin.JAVA) this.getAllFunctions() else this.getDeclaredFunctions()
```
ting-yuan, line 67: `getAllFunctions` sounds like an overkill.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSClassDeclarationJavaImpl.kt
```diff
@@ -92,10 +93,11 @@ class KSClassDeclarationJavaImpl private constructor(val psi: PsiClass) : KSClas
     }
 
     override val declarations: List<KSDeclaration> by lazy {
-        psi.fields.map { KSPropertyDeclarationJavaImpl.getCached(it) } +
+        (psi.fields.map { KSPropertyDeclarationJavaImpl.getCached(it) } +
                 psi.innerClasses.map { KSClassDeclarationJavaImpl.getCached(it) } +
                 psi.constructors.map { KSFunctionDeclarationJavaImpl.getCached(it) } +
-                psi.methods.map { KSFunctionDeclarationJavaImpl.getCached(it) }
+                psi.methods.map { KSFunctionDeclarationJavaImpl.getCached(it) })
+                .distinct()
```
ting-yuan, line 100: Do we know where the duplication comes from?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSClassDeclarationJavaImpl.kt
```diff
@@ -78,6 +78,7 @@ class KSClassDeclarationJavaImpl private constructor(val psi: PsiClass) : KSClas
             it.unsubstitutedMemberScope.getDescriptorsFiltered(DescriptorKindFilter.FUNCTIONS)
                 .toList()
                 .filter { (it as FunctionDescriptor).visibility != Visibilities.INVISIBLE_FAKE }
+                .plus(it.constructors)
```
ting-yuan, line 81: Do getAllFunctions() of KSClassDeclarationImpl and KSClassDeclarationDescriptorImpl contain constructors? primary, synthesized and defined.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
neetopia: This will only be for Java classes, and it can't be avoided as for Java classes, default constructor has to be looked up in descriptor. Alternatively we can add a `getConstructors` to `KSClassDeclaration` to save additional memory usage from `getAllFunctions()`, but the biggest overhead in resolving can't be avoided.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSClassDeclarationJavaImpl.kt
neetopia: Yes, see test case in allFunctions.kt

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSClassDeclarationJavaImpl.kt
neetopia: it comes from constructors and methods both returning same psi.

ting-yuan (approved):

PR #104, opened 2020-10-02
Disable KSP in KAPT tasks

ting-yuan:


neetopia (approved):
Maybe wait till next release to merge to see if a proper fix can land?

Issue #49, opened 2020-06-12
Add helper for equivlent of Types.asMemeberOf


evant:
When writing an annotation processor you can use [Types.asMemeberOf()](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/util/Types.html#asMemberOf-javax.lang.model.type.DeclaredType-javax.lang.model.element.Element-) to resolve the concrete type of a method based on the enclosing class. Ex if you had:
```kotlin
interface GenericInterface<T> {
  bar(): T
}

abstract class Foo : GenericInterface<String> {
}
```
you could get that `bar()` in the context of `Foo` returns `String`.

yigit:
FYI I started working on this: https://github.com/yigit/ksp/tree/as-member-of based on the version in Room.
No promises on the timeline though :)

PR #110, opened 2020-10-06
Implement Resolver.asMemberOf for Properties and Functions

yigit:
This PR implements functionality to see a property/function as member of a given type.

Internally, it uses the Substitutor from the compiler.

Since function declarations contain multiple types, I've added a new `KSFunctionType` which can hold this information.

It supports properties and functions declared in a class but *does not* support extension functions yet. We need a new API for them since they are not really a member of the receiver type.

When a substitution fails, we return the type or KSFunctionType inferred from the declaration of the property/function respectively.

Fixes: #49


api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
```diff
@@ -75,4 +75,55 @@ interface Resolver {
...
+     * class Base<T>(val x:T)
+     * val foo: Base<Int>
+     * val bar: Base<String>
+     * ```
+     * When `x` is viewed as member of `foo`, this method will return the [KSType] for `Int`
+     * whereas when `x` is viewed as member of `bar`, this method will return the [KSType]
+     * representing `String`.
+     *
+     * If the given type does not contain the [property], the [KSPropertyDeclaration.type] of the
+     * [property] will be returned.
```
ting-yuan, line 93: Would an explicit `KSErrorType` be less surprising to users?

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSFunctionType.kt
```diff
@@ -0,0 +1,55 @@
...
+ */
+package com.google.devtools.ksp.symbol
+import com.google.devtools.ksp.processing.Resolver
+/**
+ * Holds the information for a [KSFunctionDeclaration] where type arguments are resolved as member
+ * of a specific [KSType].
+ *
+ * @see Resolver.asMemberOf
+ */
+interface KSFunctionType {
```
ting-yuan, line 25: Have you considered using the `Function0<R>`, `Function1<P1, R>`, `Function2<P1, P2, R>`, ... family? Seems to me that they could also serve the purpose. Currently, explicit function types like `val f: (int) -> int = ...` as well as lambdas are resolved to the `FunctionN` family in the rest of the API.

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
yigit: The usage pattern we have in Room is that we don't really care whether it makes sense or not.
But also, we don't really call it in a case where Type may not be a container. I would guess it is not really a common case for processors but I don't have data to back that claim :).

If we want to detect errors, i think it is better to throw `IllegalArgumentException` instead of an Error type.

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSFunctionType.kt
yigit: Not sure what you mean, return `Function` from `asMemberOf`?
Can you link me some code where I can check how `FunctionN` is used?

I basically need a type that can hold return type, parameter types, receiver type and type parameters.

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
ting-yuan: `IllegalArgumentException` sounds good. I'm just worrying about that returning the default type might mute some problems which should be caught earlier.

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSFunctionType.kt
ting-yuan: I mean treating the function types as regular class types / `KSType`. In Kotlin/JVM, [function types]( https://github.com/JetBrains/kotlin/blob/master/spec-docs/function-types.md) are modeled in library and the `Function` family are normal interfaces. For example, the `KSType` of `x`'s type in  `val x: List<Int>.(Double) -> Float = ...` is `Function2<List<Int>, Double, Float>`, which is [defined as an interface](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/runtime/kotlin/jvm/functions/Functions.kt) in Kotlin stdlib.

To get the corresponding `KotlinType`, you may try [`KotlinBuiltIns.getFunction()`](https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/builtins/KotlinBuiltIns.java#L292)`.getDefaultType()`.

The instance of `KotlinBuiltIns` can be obtained like [this](https://github.com/google/ksp/blob/master/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt#L418).

Although it is designed around JVM, I think the idea is also applicable to other platforms. We can provide implementations for other platforms when needed in the future. The benefit is that we get an unified `KSType` for both functions and regular classes.

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
yigit: replace w/ illegal argument, added tests.

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSFunctionType.kt
yigit: as discussed offline, renamed it to `KSFunction`

yigit:
this is ready for review (besides I need to rebase w/ master).

Renamed KSFunctionType to KSFunction.
If the given KSType is an error or I cannot resolve the function declaration (not sure when it happens), I'll return a KSFunction which just returns KSError for all types but still matches the function declaration

yigit:
rebased, ready for review.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -453,6 +461,66 @@ class ResolverImpl(
...
+            if (!containing.kotlinType.isSubtypeOf(functionDeclaredIn)) {
+                throw IllegalArgumentException(
+                    "$containing is not a sub type of the class/interface that contains " +
+                        "`$function` ($functionDeclaredIn)"
+                )
+            }
+            val typeSubstitutor = containing.kotlinType.createTypeSubstitutor()
+            val substituted = declaration.substitute(typeSubstitutor)
+            return KSFunctionImpl(substituted)
+        }
```
ting-yuan, line 506: Just curious, what's the `KotlinType` returned for functions with their own type parameters? IIUC they cannot be expressed by `KotlinType`. E.g., `fun <T> g() = 0`.

On the other hand, if they are available in the form of `KotlinType`, then we probably don't need `KSFunction` and can just use `KSType`, which is underpinned by `KotlinType`.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSFunctionImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.google.devtools.ksp.symbol.KSType
+import com.google.devtools.ksp.symbol.KSTypeParameter
+import com.google.devtools.ksp.symbol.impl.binary.KSTypeParameterDescriptorImpl
+import org.jetbrains.kotlin.descriptors.CallableDescriptor
+
+class KSFunctionImpl(val descriptor: CallableDescriptor) : KSFunction {
+
+    override val isError: Boolean = false
+
+    override val returnType by lazy(LazyThreadSafetyMode.PUBLICATION) {
```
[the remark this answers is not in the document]
yigit: i didn't think much about this to be honest but the main reason is that using publication avoids any explicit locking in single threaded use (which is almost always the case for a processor).
https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/util/LazyJVM.kt#L107

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
yigit: i'm not sure how to get that. `declaration` does not have a KotlinType in it. Is there a helper method to get that?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -453,6 +461,66 @@ class ResolverImpl(
...
+        if (declaration != null && containing is KSTypeImpl && !containing.isError) {
+            if (!containing.kotlinType.isSubtypeOf(functionDeclaredIn)) {
+                throw IllegalArgumentException(
+                    "$containing is not a sub type of the class/interface that contains " +
+                        "`$function` ($functionDeclaredIn)"
+                )
+            }
+            val typeSubstitutor = containing.kotlinType.createTypeSubstitutor()
+            val substituted = declaration.substitute(typeSubstitutor)
+            return KSFunctionImpl(substituted)
```
ting-yuan, line 505: Can you cache the creation of `KSFunctionImpl` like [this](https://github.com/google/ksp/blob/master/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSAnnotationDescriptorImpl.kt#L41)?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
ting-yuan: Sorry it's my bad. Please ignore my comments. I was thinking `substituted` a `KotlinType`.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
yigit: i'm not sure if this is worth caching, the value is quite specific to an owner + declaration.
But even if I wanted to cache that, substitute returns `CallableDescriptor` which does not seem to enforse implementing proper `equals`. 

I've actually tried to run substitution twice between a `List<String>` type and `List#get` and they both returned different descriptors.

image

I can try to cache in the resolverImpl's `asMemberOf` with a composite key of `KSFunctionDeclaration,KSType` though again, i'm not sure how useful that cache would be.

ting-yuan: > with a composite key of `KSFunctionDeclaration,KSType`
Sounds good to me. Besides performance*, like what you said, it comes with proper `equals`.

* `asMemberOf` is inherently expensive. I'm a little bit surprised that the substituter returned different instances when called twice, which means that some of the computations are redundant.

yigit: i implemented caching, though i don't have KSFUnction#equals, I guess probably should implement it as well?
Also, might be better to do it as a followup to get this CL in.

ting-yuan (approved):

ting-yuan (approved):

PR #117, opened 2020-10-07
fix variance and classifier reference for Java unbounded type argument <?>

neetopia:


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSTypeArgumentJavaImpl.kt
```diff
@@ -51,6 +51,7 @@ class KSTypeArgumentJavaImpl private constructor(val psi: PsiType) : KSTypeArgum
             when {
                 psi.isExtends -> Variance.COVARIANT
                 psi.isSuper -> Variance.CONTRAVARIANT
+                psi.bound == null -> Variance.COVARIANT
                 else -> Variance.INVARIANT
```
ting-yuan, line 55: else -> throw ...

ting-yuan (approved):

Issue #123, opened 2020-10-12
baseMethod.overrides(childMethod) returns true


yigit:
Given:
```
interface MyInterface {
  fun foo():Unit
}

class MyClass : MyInterface {
   override fun foo():Unit {}
}
```

calling `resolver.overrides(myInterface.foo, myClass.foo)` returns `true` instead of `false`

This started failing after I moved room from old package name to the new one, I'm guessing it might be related to this change:
https://github.com/google/ksp/commit/a7085afddc82cb83af6ceea99923af13f30a7870

repo case: https://github.com/yigit/ksp/commit/aa08209a18e2e2ee4cb2b8687e9ce5791a36161d

neetopia:
a temporary workaround can be manually checking class hierarchy before calling `overrides`

PR #124, opened 2020-10-12
Check for type hierarchy while checking overrides

yigit:
This PR fixes a bug where override would return `true` if base method is sent as the overrider.
Similarly, it was also returning true when override is called with the same function itself.

I've also changed the override util call to also check for return type.

Fixes: #123

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -18,6 +18,7 @@
 
 package com.google.devtools.ksp.processing.impl
 
+import com.google.devtools.ksp.closestClassDeclaration
 import com.google.devtools.ksp.isOpen
 import com.google.devtools.ksp.isVisibleFrom
 import com.intellij.openapi.project.Project
@@ -59,6 +60,7 @@ import org.jetbrains.kotlin.resolve.*
 import org.jetbrains.kotlin.resolve.calls.smartcasts.DataFlowInfo
 import org.jetbrains.kotlin.resolve.constants.ConstantValue
 import org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluator
+import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperClassifiers
 import org.jetbrains.kotlin.resolve.jvm.multiplatform.JavaActualAnnotationArgumentExtractor
 import org.jetbrains.kotlin.resolve.lazy.DeclarationScopeProvider
 import org.jetbrains.kotlin.resolve.lazy.ForceResolveUtil
@@ -245,7 +247,6 @@ class ResolverImpl(
             return false
         if (!overridee.isVisibleFrom(overrider))
             return false
-
         if (!(overridee is KSFunctionDeclaration || overrider is KSFunctionDeclaration || (overridee is KSPropertyDeclaration && overrider is KSPropertyDeclaration)))
             return false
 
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -254,8 +255,22 @@ class ResolverImpl(
...
+            resolveClassDeclaration(it)
+        }
+        val superClassDescription = overridee.closestClassDeclaration()?.let {
+            resolveClassDeclaration(it)
+        }
+        if (subClassDescription == null || superClassDescription == null) return false
+        val typeOverride = subClassDescription.getAllSuperClassifiers()
+            .filter { it != subClassDescription } // exclude subclass itself as it cannot override its own methods
+            .any {
+                it == superClassDescription
```
neetopia, line 268: Can these 2 lamdas be merged into one?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/CheckOverrideProcessor.kt
```diff
@@ -69,5 +71,11 @@ class CheckOverrideProcessor : AbstractTestProcessor() {
         checkOverride(getY, myInterfaceX)
         checkOverride(getX, myInterfaceX)
         checkOverride(setY, myInterfaceY)
+        checkOverride(getY, getY)
```
neetopia, line 74: Can you add a test for return type checking here? Simply change the return type of `fun equals(other: Any?)` to another type such as `Int` shall work.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
yigit: `Sequence<ClassifierDescriptor>` returns a sequence so i kept them separate for readability (and does not have the cost of creating intermediate lists).
happy to change if you have a strong preference.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/CheckOverrideProcessor.kt
yigit: added one and also another one to check overriding val property with setter.
Appearantly that already works because we cannot find a matching property but still wanted to add this.

neetopia (approved):

neetopia:
LGTM. Can you rebase commit history a little bit or would you like me tot squash them all into one?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -254,8 +255,21 @@ class ResolverImpl(
 
         val superDescriptor = resolveForOverride(overridee) as? CallableMemberDescriptor ?: return false
         val subDescriptor = resolveForOverride(overrider) as? CallableMemberDescriptor ?: return false
+        val subClassDeclaration = overrider.closestClassDeclaration()?.let {
```
neetopia, line 258: Actually I was referring to name it to `subClassDescriptor` as it is the descriptor representation of the closest declaration.

yigit: whoops sorry didn't notice resolveClassDeclaration returns a descriptor. fixed.

yigit:
ready for merge. rebased manually with a proper single commit message.

neetopia:
Merged into master, thanks for contribution!

PR #128, opened 2020-10-23
add symbol validation API

neetopia:


api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    override fun visitTypeParameter(typeParameter: KSTypeParameter, data: Unit): Boolean {
+        if (typeParameter.bounds.any { shouldValidate(typeParameter, it) && !it.accept(this, Unit) }) {
+            return false
+        }
+        return true
+    }
+
+    override fun visitValueParameter(valueParameter: KSValueParameter, data: Unit): Boolean {
+        return valueParameter.type?.accept(this, Unit) != false
```
ting-yuan, line 76: Should this be `== true`, in case `valueParameter.type` is null?

api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt
```diff
@@ -0,0 +1,78 @@
...
+            return false
+        }
+        if (!validateDeclarations(function)) {
+            return false
+        }
+        return true
+    }
+
+    override fun visitPropertyDeclaration(property: KSPropertyDeclaration, data: Unit): Boolean {
+        if (property.type.resolve().isError) {
```
ting-yuan, line 59: Should this be `property.type.accept(this, Unit)`, so that `shouldValidate` has a chance to skip it?

ting-yuan (commented):
Have you tried on a larger project to see if the visitor is complete?

api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt
```diff
@@ -0,0 +1,78 @@
...
+    private fun validateDeclarations(declarationContainer: KSDeclarationContainer): Boolean {
+        return !declarationContainer.declarations.any { shouldValidate(declarationContainer, it) && !it.accept(this, Unit) }
+    }
+
+    private fun validateTypeParameters(declaration: KSDeclaration): Boolean {
+        return !declaration.typeParameters.any { shouldValidate(declaration, it) && !it.accept(this, Unit) }
+    }
+
+    private fun validateType(type: KSType): Boolean {
+        return !type.isError && !type.arguments.any { it.type?.accept(this, Unit) == false }
```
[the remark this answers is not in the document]
neetopia: No, see `errorTypes.kt` test case, error type argument does not cause the generic type itself to be error.

api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt
neetopia: that null case only happen for a property setter, in which case its validity depends on the property itself rather than the setter.

api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt
neetopia: The idea here is to skip before jumping into this check, since when checking a property, its type should be valid for the property to be valid.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -78,6 +79,14 @@ fun KSDeclaration.isLocal(): Boolean {
     return this.parentDeclaration != null && this.parentDeclaration !is KSClassDeclaration
 }
 
+/**
+ * Perform a validation on a given symbol to check if all interested types in symbols enclosed scope are valid, i.e. resolvable.
+ * @param predict: A lambda for filtering interested symbols for performance purpose. Default checks all.
+ */
+fun KSNode.validate(predict: (KSNode, KSNode) -> Boolean = { _, _-> true } ): Boolean {
```
ting-yuan, line 86: s/predict/predicate/g

ting-yuan (approved):

PR #130, opened 2020-10-27
Move KSP to an independent Gradle task

ting-yuan:
Also turn incremental compilation back on.
Note that incremental processing is still under construction.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -32,6 +32,7 @@ import com.google.devtools.ksp.gradle.KspGradleSubplugin.Companion.getKspKotlinO
 import com.google.devtools.ksp.gradle.KspGradleSubplugin.Companion.getKspResourceOutputDir
 import java.io.File
 import javax.inject.Inject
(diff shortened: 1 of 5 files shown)


neetopia (approved):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -114,10 +115,47 @@ class KspKotlinGradleSubplugin : KotlinGradleSubplugin<AbstractCompile> {
...
 }
+
+open class KspTask : KotlinCompile() {
+    lateinit var options: List<SubpluginOption>
+
+    init {
+        // kotlinc's incremental compilation isn't compatible with symbol processing in a few ways:
+        // * It doesn't consider private / internal changes when computing dirty sets.
+        // * It compiles iteratively; Sources can be compiled in different rounds.
+        incremental = false
```
ZacSweers, line 149: The PR description says it was reenabled but this seems to suggest it's not?

ting-yuan: Incremental compilation is back on in the compilation Gradle task, but incremental processing is off in the new ksp specific Gradle task, as the later is still under construction :-)

PR #137, opened 2020-11-01
Support multi module tests

yigit:
This PR adds support for multi-module tests.
Unfortunately, the current base class we use for tests does not support it. Instead, I've changed base ksp test to directly extend `KotlinBaseTest` and added a test file collector that supports modules.

Right now, we only run KSP in the last module. It can be changed later if necessary.

This does not have any impact on existing tests as we create a default module if none provided.


compiler-plugin/src/test/kotlin/com/google/devtools/ksp/test/AbstractKotlinKSPTest.kt
```diff
@@ -18,49 +18,69 @@
...
+        // get the main module where KSP tests will be run. If there is no module declared in the test input, we'll
+        // create one that will contain all test files
+        val mainModule: TestModule = files.findLast {
+            it.testModule != null
+        }?.testModule ?: TestModule("main", emptyList(), emptyList())
+
+        // group each test file with its module
+        val filesByModule = groupFilesByModule(mainModule, files)
+
+        // now compile each sub module, we'll compile the main module last.
```
neetopia, line 56: Just curious, how is the module dependency affecting compiling order? Do you have to write modules in the topological order? If so, it's fine to have such a requirement but worth noting in the comment.

yigit: yes. I could add a resolution to avoid that requirement but didn't think it is necessary and also creates a bit of ambiguity on where to apply KSP (if there are multiple leaf modules).

yigit: added a comment to clarify this.

yigit:
there is a relatively small problem i noticed here where the ResolverImpl do not get the correct module name.
It does not affect for now but eventually we need to pass the right module name to the TypeMapper (we pass default right now) and then it might matter.
I'll figure it out and do a followup CL for that.

yigit:
added a fix for the module name:
https://github.com/google/ksp/pull/137/commits/968812605edc93de983b2c3cbba38ecc662313fb

neetopia (approved):

PR #138, opened 2020-11-02
Include generated classes.

ting-yuan:


neetopia (approved):

ZacSweers:
Out of curiosity, how does this affect stale outputs?

ting-yuan:
The output directories are registered through `kspTask.outputs.dirs(kotlinOutputDir, javaOutputDir, classOutputDir)`. I think they are cleaned by `AbstractKotlinCompile` when `incremental` is set to false.

PR #140, opened 2020-11-04
fix property backing field annotations

yigit:
This commit fixes an issue where KSP would drop annotations
on properties if the property is read from a compiled code and
the annotation has field target. I fixed it by also reading
annotations on the backing field for compiled classes.

Unfortunately, the way KSP reports a kotlin class coming from
source vs compiled code is still not the same. In compiled code,
annotations that target getter/setters are reported via the generated
synthetic in .class files vs they are reported on the property in
source files.

So added tests with data class and changed the test to
ignore the difference between getting a targeted annotation
on a property vs on its getter/setter.

We might want to change it going forward but either case it would probably
be difficult to distinguish. This is luckily is not the case for field
targeting annotations as they will only be visible in backing field which
is not visible in .class files.

cleanup

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/AnnotationsInDependenciesProcessor.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/AnnotationsInDependenciesProcessor.kt
@@ -1,0 +1,84 @@
+package com.google.devtools.ksp.processor
+
+import com.google.devtools.ksp.getClassDeclarationByName
+import com.google.devtools.ksp.processing.Resolver
+import com.google.devtools.ksp.symbol.*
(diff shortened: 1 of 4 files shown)


neetopia:
The reason for that behavior difference for annotations on getter and setter is mainly due to kotlin files being  compiled to JVM class, and that compiled class file has to work with Java properly, i.e. Java need to be able to understand it. 

In that sense, I think it is not possible to completely reproduce what is in Kotlin source code after it is compiled to be class files. There might be some information in meta data which I can try to take a look.

neetopia (approved):
I agree that we do need to be able to get annotations from `@field` targets

yigit:
so to ensure i understand correctly, this change is fine but we don't necessarily desire a more complete 1-1 match since it is _almost_ impossible to get it right? sgtm.

neetopia:
Yes.

PR #141, opened 2020-11-04
Add APIs to get JVM names of functions and property accessors

### LINKED ISSUE #129: provide an API to get jvm names
For internal methods fields etc, Kotlin renames the element during compilation.

This becomes a problem in Room where we need to override / implement an internal class/field/method and we don't know the exact name.

For instance, for the following code: [full source](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:room/integration-tests/kotlintestapp/src/androidTest/java/androidx/room/integration/kotlintestapp/migration/MigrationDbKotlin.kt;l=43?q=MigrationDbK)
```
abstract class MigrationDbKotlin : RoomDatabase() {
    internal abstract fun dao(): MigrationDao
}
```
Room needs to generate `MigrationDbKotlin_Impl` and override `dao` method but we need to rename it to the JVM name (because room generates java code). In this particular case, the name we need is `dao$room_testapp_kotlin_debug()`. I'm not sure how kotlin calculates it but seems like it is coming from the project setup.

Would be great to have an API to do this conversion.

yigit:
This PR adds a new API `getJvmName` to get the jvm name of a method or accessor.

This is crucial for annotation processors that generate java source code as they might need to access internal methods in the same module compilation (as it works w/ KAPT).

Fixes #129

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
@@ -147,10 +147,10 @@ interface Resolver {
      *
      * Note that this might be different from the name declared in the Kotlin source code in two cases:
      * a) If the function receives or returns an internal class, its name will be mangled according to
-     * https://kotlinlang.org/docs/reference/inline-classes.html#mangling
+     * https://kotlinlang.org/docs/reference/inline-classes.html#mangling.
      * b) If the function is declared as internal, it will include a suffix with the module name.
      *
-     * NOTE: As inline classes are an experimental feature, the result of this function might changed based on the
+     * NOTE: As inline classes are an experimental feature, the result of this function might change based on the
      * kotlin version used in the project.
      */
     fun getJvmName(declaration: KSFunctionDeclaration): String
@@ -162,10 +162,10 @@ interface Resolver {
      * https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#properties.
      * Note that the result of this function might be different from that name in two cases:
      * a) If the property's type is an internal class, accessor's name will be mangled according to
-     * https://kotlinlang.org/docs/reference/inline-classes.html#mangling
+     * https://kotlinlang.org/docs/reference/inline-classes.html#mangling.
      * b) If the function is declared as internal, it will include a suffix with the module name.
      *
-     * NOTE: As inline classes are an experimental feature, the result of this function might changed based on the
+     * NOTE: As inline classes are an experimental feature, the result of this function might change based on the
      * kotlin version used in the project.
      * see: https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#properties
      */
--- a/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
@@ -145,8 +145,12 @@ interface Resolver {
     /**
      * Returns the jvm name of the given function.
      *
+     * The jvm name of a function might depend on the Kotlin Compiler version hence it is not guaranteed to be
+     * compatible between different compiler versions except for the rules outlined in the Java interoperability
+     * documentation: https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html.
+     *
      * Note that this might be different from the name declared in the Kotlin source code in two cases:
-     * a) If the function receives or returns an internal class, its name will be mangled according to
+     * a) If the function receives or returns an inline class, its name will be mangled according to
      * https://kotlinlang.org/docs/reference/inline-classes.html#mangling.
      * b) If the function is declared as internal, it will include a suffix with the module name.
      *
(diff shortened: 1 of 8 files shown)


api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
```diff
@@ -141,4 +141,33 @@ interface Resolver {
         function: KSFunctionDeclaration,
         containing: KSType
     ): KSFunction
+
+    /**
+     * Returns the jvm name of the given function.
+     *
+     * Note that this might be different from the name declared in the Kotlin source code in two cases:
+     * a) If the function receives or returns an internal class, its name will be mangled according to
```
ting-yuan, line 149: s/internal/inline

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
```diff
@@ -141,4 +141,33 @@ interface Resolver {
         function: KSFunctionDeclaration,
         containing: KSType
     ): KSFunction
+
+    /**
+     * Returns the jvm name of the given function.
```
ting-yuan, line 146: Can you also mention that this ties to Kotlin compiler's implementation and is not guaranteed to be cross-version compatible? Namely, not guaranteed to work on libraries compiled by other version of compilers.

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
```diff
@@ -141,4 +141,33 @@ interface Resolver {
...
+     */
+    fun getJvmName(declaration: KSFunctionDeclaration): String
+
+    /**
+     * Returns the jvm name of the given property accessor.
+     *
+     * By default, this name will match the name calculated according to
+     * https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#properties.
+     * Note that the result of this function might be different from that name in two cases:
+     * a) If the property's type is an internal class, accessor's name will be mangled according to
```
ting-yuan, line 164: s/internal/inline

compiler-plugin/testData/api/mangledNames.kt
```diff
@@ -0,0 +1,94 @@
...
+// get-internalInlineProp -> getInternalInlineProp-HRn7Rpw$mainModule
+// set-internalInlineProp -> setInternalInlineProp-E03SJzc$mainModule
+// normalFun -> normalFun
+// inlineReceivingFun -> inlineReceivingFun-E03SJzc
+// inlineReturningFun -> inlineReturningFun-HRn7Rpw
+// internalInlineReceivingFun -> internalInlineReceivingFun-E03SJzc$mainModule
+// internalInlineReturningFun -> internalInlineReturningFun-HRn7Rpw$mainModule
+// fileLevelInternalFun -> fileLevelInternalFun
+// fileLevelInlineReceivingFun -> fileLevelInlineReceivingFun-E03SJzc
+// fileLevelInlineReturningFun -> fileLevelInlineReturningFun
```
ting-yuan, line 37: Just curious, why this and `fileLevelInternalInlineReturningFun`  isn't mangled?

yigit: i honestly don't know 🤦 
I got these outputs from compiling the same input on a sample project and it is not mangled there.

```
➜  yigit-google-ksp git:(mangled-names) javap -p ~/src/ManglingPlayground/mainModule/build/classes/kotlin/main/mainPackage/InputKt.class
Compiled from "input.kt"
public final class mainPackage.InputKt {
  public static final void fileLevelInternalFun();
  public static final void fileLevelInlineReceivingFun-E03SJzc(java.lang.String);
  public static final java.lang.String fileLevelInlineReturningFun();
  public static final void fileLevelInternalInlineReceivingFun-E03SJzc(java.lang.String);
  public static final java.lang.String fileLevelInternalInlineReturningFun();
}
```
vs it is mangled when it is on a class:
```
➜  yigit-google-ksp git:(mangled-names) javap -p ~/src/ManglingPlayground/mainModule/build/classes/kotlin/main/mainPackage/Foo.class    
Compiled from "input.kt"
public final class mainPackage.Foo {
  private java.lang.String normalProp;
  private java.lang.String inlineProp;
  private java.lang.String internalProp;
  private java.lang.String internalInlineProp;
  public final java.lang.String getNormalProp();
  public final void setNormalProp(java.lang.String);
  public final java.lang.String getInlineProp-HRn7Rpw();
  public final void setInlineProp-E03SJzc(java.lang.String);
  public final java.lang.String getInternalProp$mainModule();
  public final void setInternalProp$mainModule(java.lang.String);
  public final java.lang.String getInternalInlineProp-HRn7Rpw$mainModule();
  public final void setInternalInlineProp-E03SJzc$mainModule(java.lang.String);
  public final void normalFun();
  public final void inlineReceivingFun-E03SJzc(java.lang.String);
  public final java.lang.String inlineReturningFun-HRn7Rpw();
  public final void internalInlineReceivingFun-E03SJzc$mainModule(java.lang.String);
  public final java.lang.String internalInlineReturningFun-HRn7Rpw$mainModule();
  public mainPackage.Foo();
}
```

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
yigit: updated. lmk.

yigit:
addressed the review comments and also added a test case with JvmName. Which is also another bonus of this CL as it was something that Room would need to handle manually.

yigit:
can we squash & merge this one? (I don't have merge access)

yigit:
added KspExperimental annotation.
btw, also noticed the comment above was for another PR :facepalm:

ting-yuan (approved):

Issue #135, opened 2020-10-30
Should annotation Array members use Array in KSP?


ZacSweers:
Right now, annotation members with `Array` members are represented by a `List` in the KSP API when reading them from a `KSAnnotation`. Should these be represented as `Array` in the API too? I ran into confusion around this because while casting the untyped `value` of the member I expected it to be an `Array` like the underlying type.

neetopia:
With #142 merged it should be parsed as Array now.

neetopia:
Actually still need some work to support array member for kotlin source. Will address it in next release.

neetopia:
I take previous comment back, it was fixed with last release.

PR #142, opened 2020-11-05
Fix java annotation name parsing

yigit:
This CL fixes a bug in java annotation parsing where it was always using
the index in java instead of the key. This change makes it default
to explicit name and only use index if name is not present.


yigit:
actually i noticed another problem where array types were being returned as individuals as repeated key-value pairs.
I'll fix that as well here as they are very related and probably also include #135

yigit:
will replace with another squashed one as this became a grab bag of fixes.

PR #145, opened 2020-11-06
remove nullable marker on KSValueParameter.type, add logic to retriev…

neetopia:
…e type information from property for property setters' value parameter


ting-yuan (approved):
LGTM. Before merging, can you add a test case for inferred type? For example,
```
var x
   get() = 1
   set(v) = Unit
```

PR #146, opened 2020-11-06
Fix various java annotation parsing issues

yigit:
This commit fixes multiple issues in annotation parsing for java code

1) When getting annotation names, it was always using the index instead
of the name. I changed it to default to name and only use index if
name is not provided.

2) Array values in annotations were repoted as a list of pairs instead
of 1 key and multiple values, fixed.

3) When the annotation value is a type, it was not returning a KSType
but instead returning a PsiType, fixed.

4) When the annotation value is another annotation, it was not returning
a KSAnnotation, fixed.

5) For expressions resolving to literals, it was returning the PsiLiteral
instead of the value of it.

6) TypeReferences (relevant in enum values) were not being resolved.

Updated/added tests for all cases above.

The enum fix for java source is not great right now because seems like
evaluator does not resolve it, hence I implemented a manual resolution that
is very similar to the implementation in KSAnnotationImpl.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSAnnotationJavaImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSAnnotationJavaImpl.kt
@@ -18,16 +18,18 @@
 
 package com.google.devtools.ksp.symbol.impl.java
 
+import com.google.devtools.ksp.getClassDeclarationByName
+import com.google.devtools.ksp.processing.impl.ResolverImpl
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.symbol.impl.KSObjectCache
 import com.google.devtools.ksp.symbol.impl.binary.getAbsentDefaultArguments
+import com.google.devtools.ksp.symbol.impl.kotlin.KSErrorType
 import com.google.devtools.ksp.symbol.impl.kotlin.KSNameImpl
 import com.google.devtools.ksp.symbol.impl.kotlin.KSTypeImpl
 import com.google.devtools.ksp.symbol.impl.toLocation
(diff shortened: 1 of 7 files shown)


yigit:
this probably needs a followup for descriptor implementation but i'll do that after #140 is merged.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSAnnotationJavaImpl.kt
```diff
@@ -67,15 +76,42 @@ class KSAnnotationJavaImpl private constructor(val psi: PsiAnnotation) : KSAnnot
...
+            return getCached(value)
+        }
+        val result = when(value) {
+            is PsiReference -> value.resolve()?.let { resolved ->
+                JavaPsiFacade.getInstance(value.project).constantEvaluationHelper.computeConstantExpression(value) ?: resolved
+            }
+            else -> value?.let { JavaPsiFacade.getInstance(value.project).constantEvaluationHelper.computeConstantExpression(value) }
+        }
+        return when(result) {
+            is PsiType -> {
```
neetopia, line 90: This seems to be only possible through `PsiReference.resolve()`, will it be better to merge it into the previous `when` statement?

neetopia (approved):
Overall looks good except for 1 minor comment.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSAnnotationJavaImpl.kt
yigit: i tested it but we also hit this code path for this test:
```
@KotlinAnnotation(
    stringArray = {"j-a", "j-b", null, "j-c"},
    classArray = {Object.class, List.class}
)
```
for the `classArray` values.

Similarly, the `argCLs` types in `annotationValue`. 
That test wasn't failing before because the output just compares the toString, not the type of the returned value.

Issue #157, opened 2020-11-18
kspTestKotlin output seems to clobber kspKotlin output


mvdbos:
I have two classes annotated with the same annotation. One in the main sourceset, one in the test sourceset.
In my processor, I use `CodeGenerator` to write the generated output files.

If I run only `kspKotlin`, the generated output is correctly placed in `build/generated/ksp/src/main/kotlin/packagename/SomeFile.kt`
If I subsequently run `compileKotlin`, the `.class` file for SomeFile.kt is correctly placed in the `build/classes` dir hierarchy. The generated source is untouched.

Now when I run `kspTestKotlin`, the generated source for the test sourceset is correctly placed in `build/generated/ksp/src/xitest/kotlin/packagename/SomeTestFile.kt`

Unfortunately the previously generated source for the main sourceset at `build/generated/ksp/src/main` is removed. Even the whole `main` subdir is removed. This not causing problems for running the code, since the `.class` files remain, but the IDE can now no longer find the source, which is very inconventient. 

I did some digging in the code of Ksp, but it is not clear to me why this happens. Could you help me find the cause?

mvdbos:
I did another test, and I can confirm that any invocation of `KspTask` clobbers `build/generated/ksp`.
When manually add a subdir, e.g. `build/generated/ksp/foo`, and run either `kspKotlin` or `kspTestKotlin`, the dir `build/generated/ksp` get recreated. 

Additionally, when I print the output dirs for the different `KspTask`s in Gradle like this, it indeed shows the output dir to be `build/generated/ksp`:
```kotlin
    withType<com.google.devtools.ksp.gradle.KspTask> {
        println(name)
        outputs.files.forEach { println(it.absolutePath) }
        println(destinationDir.absolutePath)
    }
```

mvdbos:
When I do the same for tasks of type `KotlinCompile`, I see that the output dir is suffixed with the sourceset, which would prevent this clobbering from happening. Is this perhaps an idea to solve this?

mvdbos:
Could it be that this causes it?

`InternalTrampoline.KotlinCompileTaskData_register(kspTaskName, kotlinCompilation, project.provider { destinationDir })`

https://github.com/google/ksp/blob/64e7f0f5535a25510e60a431530089146101695e/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L112

mvdbos:
[...]

PR #158, opened 2020-11-18
Set task outputs to kotlin output dir

mvdbos:
This PR is meant as a draft for discussion. I am not sure it is the correct way to do things, but it solves #157.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -109,10 +109,13 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
         assert(kotlinCompileProvider.name.startsWith("compile"))
         val kspTaskName = kotlinCompileProvider.name.replaceFirst("compile", "ksp")
         val destinationDir = File(project.buildDir, "generated/ksp")
-        InternalTrampoline.KotlinCompileTaskData_register(kspTaskName, kotlinCompilation, project.provider { destinationDir })
+        InternalTrampoline.KotlinCompileTaskData_register(
+            kspTaskName,
+            kotlinCompilation,
+            project.provider { kotlinOutputDir })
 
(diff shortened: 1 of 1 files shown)


mvdbos:
@googlebot I signed it!

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -109,10 +109,13 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
         assert(kotlinCompileProvider.name.startsWith("compile"))
         val kspTaskName = kotlinCompileProvider.name.replaceFirst("compile", "ksp")
         val destinationDir = File(project.buildDir, "generated/ksp")
```
yigit, line 111: i think the problem is here: ` val destinationDir = File(project.buildDir, "generated/ksp")`
it does not create the diretory specific to the compilation, instead it is global for KSP.
we should probably get another identifier from the kotlin compilation to disguish them from eachother.

mvdbos: Yes, that is indeed true. The challenge is to set it to something reasonable. Currently, there are two possible trees under `generated/ksp`: `src/$sourceSetName` and `classes/$sourceSetName`. (Of which it looks like the latter is never used, because the subsequent `compileKotlin` tasks create the .class files in `build/classes`). So even though it is most logical, unfortunately it is not as simple as appending the sourceset name to `generated/ksp`. Then you would get `generated/ksp/test/src/test` and `generated/ksp/test/classes/test`. 

Additionally, in this line all dirs specific to the sourceset are also set as outputs: 
https://github.com/google/ksp/blob/e893da3c09c53e1e7876fa62aea1be8ae6c619a5/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L121

@yigit Any suggestions? Happy to implement and test, but not sure how to proceed.

mvdbos:
Fixed by #162

PR #160, opened 2020-11-20
Optimize exception messages to contain helpful debug information

neetopia:


ting-yuan (approved):
nit: spacing properly. E.g.,

"unhandled type parameter bound, $ExceptionMessage"
const val ExceptionMessage = "please file a bug at https://github.com/google/ksp/issues/new"

PR #161, opened 2020-11-20
add annotation validation to validate API

neetopia:
Annotation argument values are left out for check for now.

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt
@@ -3,77 +3,88 @@ package com.google.devtools.ksp.visitor
 import com.google.devtools.ksp.ExceptionMessage
 import com.google.devtools.ksp.symbol.*
 
-class KSValidateVisitor(private val predicate: (KSNode, KSNode) -> Boolean) : KSDefaultVisitor<Unit, Boolean>() {
-    private fun validateDeclarations(declarationContainer: KSDeclarationContainer): Boolean {
-        return !declarationContainer.declarations.any { predicate(declarationContainer, it) && !it.accept(this, Unit) }
-    }
-
-    private fun validateTypeParameters(declaration: KSDeclaration): Boolean {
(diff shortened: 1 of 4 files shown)


api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt
```diff
@@ -40,6 +54,9 @@ class KSValidateVisitor(private val predicate: (KSNode, KSNode) -> Boolean) : KS
     }
```
ting-yuan, line 54: Should have noticed previously: Since this visitor extends `KSDefaultVisitor`, why not delegating to it? I.e., replace calls to `validateAnnotations()`, `validateTypeParameters()`, `validateDeclarations()` and `validateDeclaration()` with a single `return super.validateClassDeclaration()`. Then you don't need `private fun validateBlabla`; Just put them in `visitBlabla()`. For example,

```
override fun visitClassDeclaration(c: KSClassDeclaration) {
  // Only validate super types here
  if ( c.superTypes.any { it.accept(this, c) } == false )
    return false
  // This in turn calls visitDeclaration and visitDeclarationContainer
  return super.visitClassDeclaration()
}

override fun visitDeclaration(decl: KSDeclaration) {
  if (decl.typeParameters.any { it.accept(this, decl) } == false)
    return false
  return super.visitDeclaration()
}

override fun visitTypeParameter() {
  // validate a type parameter.
}

override fun visitDeclarationContainer(container: KSDeclarationContainer, ...) {
  return container.declarations.all { it.accept(this, container) }
}

```

and so on.

ting-yuan (approved):

ting-yuan:
Actually https://github.com/google/ksp/commit/d050c49d25efc617f7c31496d22247f236a8b351 looks better. May I know the reason why you decided to go with the latter approach?

neetopia:
> Actually [d050c49](https://github.com/google/ksp/commit/d050c49d25efc617f7c31496d22247f236a8b351) looks better. May I know the reason why you decided to go with the latter approach? Besides the default handler.

Because if we simply delegate to super, there is no control for us to let super return our desired result, all return result will be from defaultHandler and things will break, like the call to common logic won't affect the return result of `super.visitClassDeclaration()`

PR #162, opened 2020-11-20
Reorganize output directories.

### LINKED ISSUE #157: kspTestKotlin output seems to clobber kspKotlin output

ting-yuan:
So that they are under the same, sourceSetName based destinationDir.

Fixes #157

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -47,20 +47,24 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
         const val KSP_PLUGIN_ID = "com.google.devtools.ksp.symbol-processing"
 
         @JvmStatic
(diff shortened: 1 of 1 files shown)


neetopia (approved):

mvdbos:
Thanks!

PR #165, opened 2020-11-24
Expose logger statistics

neetopia:
~~When seeing errors or exceptions in logger event, pass the original binding context back to fail compilation.~~

~~This is quite tricky since there might be errors recorded in processor that are not related to compiler errors. A better solution need to change compiler plugin extension point to support such custom 'fail on sight' style `AnalysisResult`. Alternatively it is also doable to just throw exception in KSP to fail compilation but that might be too aggressive.~~

Since error out works now, we no longer need to fail compilation here, however it is still worth to expose logger statistics for future usage.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
@@ -37,8 +37,6 @@ import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.symbol.impl.KSObjectCacheManager
 import com.google.devtools.ksp.symbol.impl.java.KSFileJavaImpl
 import com.google.devtools.ksp.symbol.impl.kotlin.KSFileImpl
-import org.jetbrains.kotlin.incremental.isJavaFile
-import org.jetbrains.kotlin.incremental.isKotlinFile
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.resolve.BindingContext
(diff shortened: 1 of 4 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -107,7 +107,11 @@ abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions,
         bindingTrace: BindingTrace,
         files: Collection<KtFile>
     ): AnalysisResult? {
-        return AnalysisResult.success(BindingContext.EMPTY, module, shouldGenerateCode = false)
+        return if (logger.errorCount > 0 || logger.exceptionCount > 0) {
+            AnalysisResult.success(bindingTrace.bindingContext, module, shouldGenerateCode = false)
+        } else {
+            AnalysisResult.success(BindingContext.EMPTY, module, shouldGenerateCode = false)
+        }
```
ting-yuan, line 114: Just curious, why not simply changing `BindingContext.EMPTY` to `bindingTrace.bindingContext`?

neetopia: Because we don't want to fail processing in this branch.

ting-yuan: This kind of explanation isn't really helpful. Please provide more information next time. For example, there are other errors from compiler such as symbol not found that are not fatal to ksp.

ting-yuan (approved):

Issue #168, opened 2020-11-25
upgrade to kotlin 1.4.20


yigit:
Just filing this as a tracking ticket.
Looks like KSP is not compatible w/ 1.4.20 (as expected).

I hit this when trying to use KSP in room as AndroidX already uses 1.4.20. 

```
e: java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.descriptors.CallableDescriptor org.jetbrains.kotlin.resolve.calls.inference.InferenceUtilsKt.substitute(org.jetbrains.kotlin.descriptors.CallableDescriptor, org.jetbrains.kotlin.resolve.calls.inference.components.NewTypeSubstitutor)'
        at com.google.devtools.ksp.processing.impl.ResolverImpl.asMemberOf(ResolverImpl.kt:542)
        at androidx.room.compiler.processing.ksp.KSAsMemberOfKt.typeAsMemberOf(KSAsMemberOf.kt:34)
```

I took a quick look to see if ksp's own source compiles.  It seems to be affected by this CL:
https://github.com/JetBrains/kotlin/commit/a05d6da43b0b6fd3bb0f31d3603b5fd192b597b1#diff-9dce699a83c65328e3f52c6669cabb8f0e5115835ddd41e83cfeb2441c29db99

I got it to compile (https://github.com/yigit/ksp/tree/update-kotlin-1.4.20) but haven't run tests yet as it requires copying test utils but i don't have a kotlin checkout to easily try it.

yigit:
so i tried pulling in the tests but it was failing w/ missing class exceptions, looks like there were some changes around mutes.
I pulled in tests-mutes just to try as well but then it failed with the following exception.

```
'void com.intellij.core.CoreApplicationEnvironment.registerExtensionPointAndExtensions(java.nio.file.Path, java.lang.String, com.intellij.openapi.extensions.ExtensionsArea)'
java.lang.NoSuchMethodError: 'void com.intellij.core.CoreApplicationEnvironment.registerExtensionPointAndExtensions(java.nio.file.Path, java.lang.String, com.intellij.openapi.extensions.ExtensionsArea)'
	at org.jetbrains.kotlin.cli.jvm.compiler.CoreApplicationEnvironmentCompatKt.registerExtensionPointAndExtensionsEx(coreApplicationEnvironmentCompat.kt:17)
[...]

PR #169, opened 2020-11-25
Update kotlin to 1.4.20

yigit:
This PR updates kotlin base version to 1.4.20.

Visibilities has moved to DescriptorVisibilties.
JavaVisibilities has moved to JavaDescriptorVisibilities
Test configuration now requires a backend, I picked JVM
Updated intellij base version to 202.8194.7
Updated intellij gradle plugin to 0.6.4

One unexpected changes is that seems like testing prebuilt
now also requires mutes. I've added that jar as well.

Fixes: #168


gradle.properties
```diff
@@ -1,7 +1,7 @@
 # Copied from kotlinc
 org.gradle.jvmargs=-Duser.country=US -Dkotlin.daemon.jvm.options=-Xmx2200m -Dfile.encoding=UTF-8
 
-kotlinBaseVersion=1.4.10
-intellijVersion=193.6494.35
```
neetopia, line 5: Not sure if this is necessary, I have a commit without this change but still works.

neetopia (commented):
this is [my commit](https://github.com/google/ksp/commit/7aa10a69528878c8c45dca5ea1ff4ec0a5174895) on updating to latest commit HEAD, it looks to be different, could be a further change on visibility related code.

gradle.properties
yigit: my tests were failing until i updated intellijVersion.

yigit: i tried reverting it but then tests fail with:

```
'void com.intellij.core.CoreApplicationEnvironment.registerExtensionPointAndExtensions(java.nio.file.Path, java.lang.String, com.intellij.openapi.extensions.ExtensionsArea)'
java.lang.NoSuchMethodError: 'void com.intellij.core.CoreApplicationEnvironment.registerExtensionPointAndExtensions(java.nio.file.Path, java.lang.String, com.intellij.openapi.extensions.ExtensionsArea)'
	at org.jetbrains.kotlin.cli.jvm.compiler.CoreApplicationEnvironmentCompatKt.registerExtensionPointAndExtensionsEx(coreApplicationEnvironmentCompat.kt:17)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.registerApplicationExtensionPointsAndExtensionsFrom(KotlinCoreEnvironment.kt:537)
	
```

so your tests pass w/o this change?

neetopia: I see, my change is without test enabled due to test-mutes module issue, then it makes sense to upgrade intellij version.

yigit: :+1: 
i added the test-mutes here to fix that, seems to be working fine.

yigit: i also tried breaking a test to ensure they are really run :)

neetopia:
LGTM, I'll merge it before next release.

gradle.properties
```diff
@@ -1,7 +1,7 @@
...
+kotlinBaseVersion=1.4.20
+intellijVersion=202.8194.7
```
ting-yuan, line 5: Can we follow kotlin-1.4.20? I.e., 201.7223.91

ting-yuan: https://github.com/JetBrains/kotlin/blob/master/gradle/versions.properties#L1

yigit: done.

ting-yuan (approved):

PR #171, opened 2020-11-25
Support finding descriptors for java statics

yigit:


yigit:
ugh i was passing jdk 11 path for jdk 9 and seems like it made the function visible after this fix.
with jdk 9 it does not show up, will revert the test change

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -317,17 +318,30 @@ class ResolverImpl(
...
+                            (it as? PropertyDescriptor)?.getter?.findPsi() == psi || (it as? PropertyDescriptor)?.setter?.findPsi() == psi
+                        }
                 } else {
-                    moduleClassResolver.resolveClass(JavaMethodImpl(psi).containingClass)
-                            ?.unsubstitutedMemberScope!!.getDescriptorsFiltered().singleOrNull { it.findPsi() == psi }
+                    moduleClassResolver
+                        .resolveClass(JavaMethodImpl(psi).containingClass)
+                        ?.findEnclosedDescriptor(
+                            kindFilter = DescriptorKindFilter.FUNCTIONS,
+                            psi = psi
```
neetopia, line 333: It looks no difference from just pass the filter lambda at this call site.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -697,3 +718,38 @@ private fun Name.getNonSpecialIdentifier() :String {
...
+        kindFilter = kindFilter,
+        filter = filter
+    )
+}
+
+private fun ClassDescriptor.findEnclosedDescriptor(
+    kindFilter: DescriptorKindFilter,
+    psi: PsiElement
+): DeclarationDescriptor? {
+    return this.unsubstitutedMemberScope.findEnclosedDescriptor(
```
neetopia, line 748: Since you are using `firstOrNull` it should make no performance impact if we just combine 2 lists together and call the filter.

neetopia (approved):
LGTM, only have 2 minor comments.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
yigit: yea you are right, i first had the psi version then added the filter version hence this unnecessary duplication.
removed the psi version to always require the lambda.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
yigit: this method is going away (and it was wrong by not delegating to the ClassDescriptor filter version) so i'm assuming this comment applies to the other one as well?
in that case though i have member scopes that i cannot combine.
I can combine the results of getContributedDescriptors but that would have two unnecessary things:
* it would query staticScope regardless even when we find the result in member scope
* it would create another list to combine the two.

hence, i think it is better to keep the `ClassDescriptor.findEnclosedDescriptor` above as is? (this function is going away)

Issue #94, opened 2020-09-30
crash in overrides check for java classes that implement kotlin interfaces


yigit:
I'm getting a crash with overrides check for java methods.

ksp version: `1.4.0-dev-experimental-20200914`

Interestingly, it only happens if java class implements a kotlin interface. If i remove the `implements MyInterface`, it works fine.

repro test:
```
    @Test
    fun overrideCrash() {
        val kotlinSrc = Source.kotlin("Foo.kt", """
            interface MyInterface {
                val x:Int
                var y:Int
            }
        """.trimIndent())
        val javaSrc = Source.java("Subject", """
            class JavaImpl implements MyInterface {
                public int getX() {
                    return 1;
                }
                public int getY() {
                    return 1;
                }
                public void setY(int value) {
                    
                }
            }
        """.trimIndent())
        runKspTest(sources = listOf(kotlinSrc, javaSrc), succeed = true) {
            val klass = it.kspResolver.getClassDeclarationByName("JavaImpl")!!
            val getX = klass.getDeclaredFunctions().first {
                it.simpleName.asString() == "getX"
            }
            val getY = klass.getDeclaredFunctions().first {
                it.simpleName.asString() == "getY"
            }
            Truth.assertThat(getY.overrides(getX)).isFalse()
        }
    }
```
it crashes with:
```
java.lang.ClassCastException: class org.jetbrains.kotlin.load.java.descriptors.JavaForKotlinOverridePropertyDescriptor cannot be cast to class org.jetbrains.kotlin.descriptors.FunctionDescriptor (org.jetbrains.kotlin.load.java.descriptors.JavaForKotlinOverridePropertyDescriptor and org.jetbrains.kotlin.descriptors.FunctionDescriptor are in unnamed module of loader 'app')

	at org.jetbrains.kotlin.ksp.processing.impl.ResolverImpl.resolveJavaDeclaration(ResolverImpl.kt:478)
	at org.jetbrains.kotlin.ksp.processing.impl.ResolverImpl.resolveFunctionDeclaration(ResolverImpl.kt:250)
	at org.jetbrains.kotlin.ksp.symbol.impl.java.KSFunctionDeclarationJavaImpl.overrides(KSFunctionDeclarationJavaImpl.kt:51)
```

PR #176, opened 2020-11-27
FIx findOverridee for properties, fixes #174

### LINKED ISSUE #174: Should KSPropertyDeclarationImpl.findOverridee use singleOrNull?
https://github.com/google/ksp/blob/638d9a19a9880056c8c95a5a2f5775045cb026ae/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSPropertyDeclarationImpl.kt#L107

This now fails for me on an overridden val. The way this code is written now, looks like the last '?' in this part of the statement `.single { it.overriddenDescriptors.isEmpty() }?` is never going to happen, since `single` never returns null.

yigit:
This CL updates KSPropertyDecl.findOverridee to use the same infra
that is used by KSFunctionDeclaration.

Fixes: #174

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSPropertyDeclaration.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSPropertyDeclaration.kt
@@ -58,8 +58,30 @@ interface KSPropertyDeclaration : KSDeclaration {
     fun isDelegated(): Boolean
 
     /**
-     * Find the original overridee of this property, if overriding.
-     * @return [KSPropertyDeclaration] for the original property, if overriding, otherwise null.
+     * Find the closest overridee of this property, if overriding.
+     *
+     * For the following input:
+     * ```
+     * abstract class A {
+     *   open val x:Int
+     *   open val y:Int
+     * }
+     * abstract class B : A() {
+     *   override val x:Int
+     * }
+     * abstract class C : B() {
+     *   override val x:Int
+     *   override val y:Int
+     * }
+     * ```
+     * Calling `findOverridee` on `C.x` will return `B.x`.
+     * Calling `findOverridee` on `C.y` will return `A.y`.
+     *
+     * When there are multiple super classes / interfaces with the property, the closest declaration
+     * to the current containing declaration is selected. If they are in the same level, the
+     * property of the first specified interface (in source) will be returned.
+     *
+     * @return [KSPropertyDeclaration] for the overridden property, if overriding, otherwise null.
      * Calling [findOverridee] is expensive and should be avoided if possible.
      */
     fun findOverridee(): KSPropertyDeclaration?
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/OverrideeProcessor.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/OverrideeProcessor.kt
@@ -22,6 +22,7 @@ import com.google.devtools.ksp.getClassDeclarationByName
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.symbol.KSClassDeclaration
 import com.google.devtools.ksp.symbol.KSFunctionDeclaration
+import com.google.devtools.ksp.symbol.KSPropertyDeclaration
 
 @Suppress("unused") // used by tests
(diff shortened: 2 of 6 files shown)


neetopia (commented):
The overriding for properties usually get complicated when it involves Java, can you add a test case where Java code is overriding a property declared in Kotlin code? See [this](https://github.com/google/ksp/blob/638d9a19a9880056c8c95a5a2f5775045cb026ae/compiler-plugin/testData/api/checkOverride.kt#L110) for example.

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSPropertyDeclaration.kt
```diff
@@ -58,8 +58,30 @@ interface KSPropertyDeclaration : KSDeclaration {
...
+     *   override val x:Int
+     *   override val y:Int
+     * }
+     * ```
+     * Calling `findOverridee` on `C.x` will return `B.x`.
+     * Calling `findOverridee` on `C.y` will return `A.y`.
+     *
+     * When there are multiple super classes / interfaces with the property, the closest declaration
+     * to the current containing declaration is selected. If they are in the same level, the
+     * property of the first specified interface (in source) will be returned.
```
neetopia, line 82: nit: `specified super class or interface` instead of `specified interface`

yigit: > The overriding for properties usually get complicated when it involves Java, can you add a test case where Java code is overriding a property declared in Kotlin code? See [this](https://github.com/google/ksp/blob/638d9a19a9880056c8c95a5a2f5775045cb026ae/compiler-plugin/testData/api/checkOverride.kt#L110) for example.

the findOverridee in the java implementation of property is not implemented (just returns null). I'm guessing that is because java fields do not override parents.

You mean there is a base kotlin class with `val x:T` and java overrides it with `getX():T`?
I'll give it a try but that makes me think, that would mean the `findOverridee` of `KSFunction` would need to return a `KSPropertyAccessor` that does not implement `KSFunction` :/. 
I'll see what happens there.

There was also some work discussed before where KSP right now does not properly resolve property accessors in java source code. If we do, then we should also implement findOverridee in KSPropertyDeclaration.

yigit: actually trying to reproduce that hits this issue:
https://github.com/google/ksp/issues/94

yigit: i'm actually able to workaround that issue by making KSJavaFileImpl use descriptors for the class instead of Java impls.
Java impls are causing lots of inconsistencies hence created this issue:
https://github.com/google/ksp/issues/177

mvdbos: Thanks for looking into this so fast. 

Question: is the merge of this PR blocked by the progress on #177? Or can this be merged already? It would help me a lot: until this is fixed, I need to maintain my fork. I use Kotlin only. :-)

yigit: @neetopia wdyt? I think this one is fine to merge as it should be an improvement over what we have (fixes kotlin properties, java is still broken).
#177 probably won't happen or if it happens it will take a long time.

neetopia: Sounds good, I'll merge this one.

neetopia (approved):

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSPropertyDeclaration.kt
mvdbos: Thanks for the merge. Is it possible to create a release that contains this? I am using KSP as a dependency.
If not, what is a likely date for a new release, given your release schedule?

neetopia: I'll do a release this Friday PST.

PR #182, opened 2020-12-04
add KSTypeAliasDescriptorImpl

neetopia:
There is implicit type aliasing happening for certain Kotlin standard library classes, if not imported explicitly for its true type, will be presented as a type alias through resolution.


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSTypeAliasDescriptorImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    override val name: KSName by lazy {
+        KSNameImpl.getCached(descriptor.name.asString())
+    }
+
+    override val modifiers: Set<Modifier> by lazy {
+        val modifiers = mutableSetOf<Modifier>()
+        modifiers.addAll(descriptor.toKSModifiers())
+        modifiers
```
ting-yuan, line 27: Why not just `descriptor.toKSModifiers()`?

ting-yuan (approved):

PR #186, opened 2020-12-08
Incremental processing for Kotlin sources

ting-yuan:
For now, changes in other modules, external libraries and Java sources will trigger a full rebuild.

It is disabled by default and can be switched on by the Gradle property `ksp.incremental=true`.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
@@ -1,0 +1,453 @@
+package com.google.devtools.ksp
+
+import com.google.devtools.ksp.symbol.*
+import com.google.devtools.ksp.symbol.impl.kotlin.KSFileImpl
+import com.google.devtools.ksp.visitor.KSDefaultVisitor
+import com.intellij.util.containers.MultiMap
+import com.intellij.util.io.DataExternalizer
+import com.intellij.util.io.IOUtil
+import com.intellij.util.io.KeyDescriptor
+import org.jetbrains.kotlin.container.ComponentProvider
+import org.jetbrains.kotlin.container.get
+import org.jetbrains.kotlin.incremental.*
+import org.jetbrains.kotlin.incremental.components.LookupTracker
+import org.jetbrains.kotlin.incremental.storage.BasicMap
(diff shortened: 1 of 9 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -0,0 +1,438 @@
...
+        logSourceToOutputs()
+
+        sourceToOutputsMap.flush(false)
+        sourceToOutputsMap.close()
+    }
+
+    // TODO: Recover if processing failed.
+    fun updateOutputs(outputs: Set<File>, cleanOutputs: Collection<File>) {
+        val outRoot = options.kspOutputDir
+        val bakRoot = File(options.cachesDir, "backups")
```
neetopia, line 356: Backups seems to be unused.

ting-yuan: Added comments in code.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -0,0 +1,438 @@
...
+        val mutableSourceToOutputs = mutableMapOf<File, MutableSet<File>>().apply {
+            sourceToOutputs.forEach {
+                set(it.key, it.value.toMutableSet())
+            }
+        }
+
+        // Associate unassociated outputs to ALL FILES.
+        val associated = sourceToOutputs.values.flatten().toSet()
+        val unassociated = outputs.filterNot { it in associated }
+        mutableSourceToOutputs.getOrPut(anyChangesWildcard) { mutableSetOf() }.addAll(unassociated)
```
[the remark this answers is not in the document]
ting-yuan: There are several choices of default association.

1. Associate with everything. This is probably safest but also a performance killer for non-cooperating processors.

2. Associate with nothing. They don't affect the dirty set. This somehow forces processors to provide associations. Valid use cases without association include log / debug files.

I much prefer 2. but 1. seems to be easier to start with. I'll pitch the question in discussion channels. Before making a decision, this implementation is easiest to adapt to either choice.

neetopia (approved):
Feel free to clean commit history and push.

PR #188, opened 2020-12-10
Add test cases for cross module typealiases.

ting-yuan:


neetopia (approved):
Can this test be merged into type alias test we already have?

Issue #195, opened 2020-12-18
resolving a class from KSType.declaration vs getClassDeclarationByName gives different results


yigit:
repro CL: https://github.com/yigit/ksp/tree/class-declaration-inconsistency-repro

for some reason, i get two different KSClassDeclaration's for the same class when it is resolved from a KSType vs getClassDeclarationByName

ting-yuan:
When applicable, builtins on jvm has a special `mockSerializableType` super type, which is hard-coded to have an empty scope:
https://github.com/JetBrains/kotlin/blob/master/core/descriptors.jvm/src/org/jetbrains/kotlin/builtins/jvm/JvmBuiltInsSettings.kt#L79

Besides fixing in upstream, there is no good way to workaround unfortunately.

ting-yuan:
Worked around in #196.

PR #196, opened 2020-12-18
Replace `mockSerializableType` with real one.

ting-yuan:
See https://github.com/google/ksp/issues/195 for details.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -19,16 +19,6 @@
 package com.google.devtools.ksp.processing.impl
 
 import com.google.devtools.ksp.*
-import com.intellij.openapi.project.Project
-import com.intellij.psi.*
-import com.intellij.psi.impl.source.PsiClassReferenceType
-import org.jetbrains.kotlin.codegen.ClassBuilderMode
-import org.jetbrains.kotlin.codegen.state.KotlinTypeMapper
-import org.jetbrains.kotlin.container.ComponentProvider
-import org.jetbrains.kotlin.container.get
-import org.jetbrains.kotlin.descriptors.*
-import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
-import org.jetbrains.kotlin.incremental.components.NoLookupLocation
 import com.google.devtools.ksp.processing.KSBuiltIns
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.symbol.*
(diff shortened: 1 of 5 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -163,8 +156,7 @@ class ResolverImpl(
     override fun getClassDeclarationByName(name: KSName): KSClassDeclaration? {
         nameToKSMap[name]?.let { return it }
 
-        return (module.resolveClassByFqName(FqName(name.asString()), NoLookupLocation.FROM_BUILTINS)
-            ?: module.resolveClassByFqName(FqName(name.asString()), NoLookupLocation.FROM_DESERIALIZATION))
```
ting-yuan, line 167: Lookup location doesn't affect resolution.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -653,6 +645,13 @@ class ResolverImpl(
             override val arrayType: KSType by lazy { getKSTypeCached(builtIns.array.defaultType.replaceArgumentsWithStarProjections()) }
         }
     }
+
+    internal val mockSerializableType = module.builtIns.numberType.supertypes().single {
```
neetopia, line 649: If I understand correctly, this should also work with other builtin types, picking `numberType` is just random, right?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
```diff
@@ -68,10 +69,14 @@ class KSClassDeclarationDescriptorImpl private constructor(val descriptor: Class
         descriptor.unsubstitutedPrimaryConstructor?.let { KSFunctionDeclarationDescriptorImpl.getCached(it) }
     }
 
+    // Workaround for https://github.com/google/ksp/issues/195
+    private val mockSerializableType = ResolverImpl.instance.mockSerializableType
```
neetopia, line 73: Since it is already a singleton, maybe just use that singleton directly?

neetopia (approved):
LGTM, just 2 minor quetions.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
ting-yuan: Kind of. Number has fewest super types. I tried String as well and it also works.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
ting-yuan: I wanted to minimize the per-call overhead here. I think a reference to a private val is faster than `ResolverImpl.instance.mockSerializableType`?

PR #197, opened 2020-12-19
Insert simulated lookups for Java sources

ting-yuan:
This is WIP but mostly done; The only remaining task is running integration tests.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
@@ -29,8 +29,10 @@ import com.intellij.util.containers.MultiMap
 import com.intellij.util.io.DataExternalizer
 import com.intellij.util.io.IOUtil
 import com.intellij.util.io.KeyDescriptor
+import org.intellij.plugins.relaxNG.compact.psi.util.PsiFunction
 import org.jetbrains.kotlin.container.ComponentProvider
 import org.jetbrains.kotlin.container.get
+import org.jetbrains.kotlin.descriptors.CallableMemberDescriptor
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.incremental.*
 import org.jetbrains.kotlin.incremental.components.LookupTracker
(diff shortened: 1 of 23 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -428,6 +453,129 @@ class IncrementalContext(
...
+            record(psiFile.packageName, name)
+            onDemandImports.forEach {
+                record(it, name)
+            }
+        }
+    }
+
+    private fun recordLookup(ref: PsiClassReferenceType, def: PsiClass) {
+        val psiFile = ref.reference.containingFile as? PsiJavaFile ?: return
+        // A type parameter doesn't have qualified name.
```
neetopia, line 478: Looks like we just don't record for type parameters, what should be the ideal behavior for a type parameter has an upper bound which is concrete?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -428,6 +453,129 @@ class IncrementalContext(
...
+        when (declaration) {
+            is KSPropertyDeclarationJavaImpl -> recordLookup(declaration.psi)
+            is KSFunctionDeclarationJavaImpl -> recordLookup(declaration.psi)
+        }
+    }
+
+    fun recordLookupForCallableMemberDescriptor(descriptor: CallableMemberDescriptor) {
+        val psi = descriptor.findPsi()
+        when (psi) {
+            is PsiMethod -> recordLookup(psi)
```
neetopia, line 535: what about just make `recordLookup` take care of type checking?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -428,6 +453,129 @@ class IncrementalContext(
...
+            it.findPsi() as? PsiClass
+        }.forEach { psiClass ->
+            psiClass.superTypes.forEach {
+                recordLookup(it)
+            }
+            doChild(psiClass)
+        }
+    }
+
+    fun dumpLookupRecords(): Map<String, List<String>> {
```
neetopia, line 569: add `TODO: remove` comment

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
ting-yuan: We don't do it here. This function is just used to record a type reference to a definition. Type parameters' bounds are recorded in other places such as `recordLookup(psi: PsiMethod)` only when necessary.

Will add a comment.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
ting-yuan: Can you give an example?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
ting-yuan: It will still be useful for debug purposes. Can you keep it?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
neetopia: make `recordLookup(psi: PsiElement)` and then handle psi type checking there.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
neetopia: Ok, then maybe make it internal and add comment on suggested usage?

compiler-plugin/testData/api/recordJavaAnnotationTypes.kt
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// WITH_RUNTIME
+// TEST PROCESSOR: RecordJavaProcessor
+// EXPECTED:
+// java.util.List: javaSrc/p1/J.java
+// p1.Anno: javaSrc/p1/J.java
+// p1.Bnno: javaSrc/p1/J.java
```
neetopia, line 23: Can you elaborate on this line? Bnno is defined in a kotlin file, since this test processor only output java files, why looking up `p1.Bnno` still recorded as dependent on `J.java`?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
ting-yuan: Will rename to recordLookupPsiMethod() and recordLookupPsiField().

neetopia (approved):

PR #199, opened 2020-12-21
Require specifying sources when creating new files

ting-yuan:


api/src/main/kotlin/com/google/devtools/ksp/processing/CodeGenerator.kt
```diff
@@ -32,22 +32,38 @@ interface CodeGenerator {
...
+     *                are required.
      * @param packageName corresponds to the relative path of the generated file; using either '.'or '/' as separator.
      * @param fileName file name
      * @param extensionName If "kt" or "java", this file will participate in subsequent compilation.
      *                      Otherwise its creation is only considered in incremental processing.
      * @return OutputStream for writing into files.
      * @see [CodeGenerator] for more details.
      */
-    fun createNewFile(packageName: String, fileName: String, extensionName: String = "kt"): OutputStream
+    fun createNewFile(sources: List<KSFile>, packageName: String, fileName: String, extensionName: String = "kt"): OutputStream
```
neetopia, line 60: This is a breaking change that is only needed when incremental is enabled, can you make it default to empty list?

yigit: actually this is a bit scary to set to empty list because that would mean the processor will be silently broken in incremental compilation.
Also, we cannot distinguish whether developer just doesn't have additional sources OR they didn't implement incremental compilation.

Should we instead provide an OPT-OUT API where processor needs to specify that it doesn't support incremental compilation and we throw an error if incremental processing is enabled? (or a warning if we can workaround it by invalidating it every time)
That being said, i'm all for making opt out harder :D.

neetopia (approved):

api/src/main/kotlin/com/google/devtools/ksp/processing/CodeGenerator.kt
ting-yuan: Yeah, this is intended to make opt-out hard :p How about a AllFilesWildcard with a scary comment "this will disable incremental processing"?

yigit: yea i like that :)
does that mean this should change to a specific class like `Dependencies` which can be
`Dependencies(files)` or `Dependencies.ALL_FILES`
Not sure if dependencies is a good name but making it another class would give developer the attention (otherwise, it is easy to pass empty list and not think about it)

ting-yuan: @yigit @neetopia Updated. Please take a look again.

api/src/main/kotlin/com/google/devtools/ksp/processing/CodeGenerator.kt
```diff
@@ -32,22 +32,38 @@ interface CodeGenerator {
...
+     * are for. If a [KSFile] is indirectly obtained through other [KSNode]s, it hasn't to be specified for the given
+     * output, even if its contents contribute to the generation of the output.
+     *
+     * For example, a processor generates an output `O` after reading class `A` in `A.kt` and class `B` in `B.kt`,
+     * where `A` extends `B`. The processor got `A` by [Resolver.getSymbolsWithAnnotation] and then got `B` by
+     * [KSClassDeclaration.superTypes] from `A`. Because the inclusion of `B` is due to `A`, `B.kt` needn't to be
+     * specified in [dependencies] for `O`. Note that specifying `B.kt` in this case doesn't hurt, it is only unnecessary.
+     *
+     * @param dependencies are [KSFile]s from which this output is built. Only those that are obtained directly
+     *                     from [Resolver] are required.
```
yigit, line 52: curious, if the generated code references the file, does it still need to be provided?

e.g. if I query `getSymboldsAnnotatedWith`, get interface `Foo` and generate `FooImpl : Foo`, do I still need to specify the source file for `Foo` or does it automatically get captured?

api/src/main/kotlin/com/google/devtools/ksp/processing/CodeGenerator.kt
```diff
@@ -56,16 +72,28 @@ interface CodeGenerator {
      * @see [CodeGenerator] for more details.
      */
     fun associate(sources: List<KSFile>, packageName: String, fileName: String, extensionName: String = "kt")
+}
+
+/**
+ * Dependencies of an output file.
+ */
+class Dependencies private constructor(val isAllSources: Boolean, val dependOnNewChanges: Boolean, val sources: List<KSFile>) {
```
yigit, line 80: OriginatingFiles might be a better name: https://docs.oracle.com/javase/7/docs/api/javax/annotation/processing/Filer.html#createClassFile(java.lang.CharSequence,%20javax.lang.model.element.Element...)

or originating nodes if the API received nodes instead of files.

yigit (commented):
overall lgtm, just a 1 question and a small name suggestion (feel free to ignore0

api/src/main/kotlin/com/google/devtools/ksp/processing/CodeGenerator.kt
ting-yuan: Yes, it still is needed. Current incremental + multiple-round plan is to leave generated code alone when computing dependencies; I.e., not rely on them at all.

api/src/main/kotlin/com/google/devtools/ksp/processing/CodeGenerator.kt
ting-yuan: Done.

PR #205, opened 2020-12-22
change override checking logic to traverse CallableMemberDescriptor.overriddenDescriptors()

neetopia:
Current override checking utilize `OverridingUtil` which appears to be unable to handle concrete types when overriding a member from a generic super type. After peeking into the structure of `CallableMemberDecriptor` for a while, a BFS looks promising.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -24,6 +24,7 @@ import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.symbol.Variance
 import com.google.devtools.ksp.symbol.impl.binary.*
+import com.google.devtools.ksp.symbol.impl.findClosestOverridee
 import com.google.devtools.ksp.symbol.impl.findPsi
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -290,9 +291,18 @@ class ResolverImpl(
...
-                superDescriptor, subDescriptor, subClassDescriptor, true
-        ).result == OverridingUtil.OverrideCompatibilityInfo.Result.OVERRIDABLE
+        var queue = listOf(subDescriptor)
+        while (queue.isNotEmpty()) {
+            val newQueue = mutableListOf<CallableMemberDescriptor>()
+            for (overridden in queue) {
+                if (overridden === superDescriptor) {
+                    return true
+                }
+                newQueue.addAll(overridden.overriddenDescriptors.map { it.original })
```
ting-yuan, line 301: I think you can get rid of the above super type check as well.

Also, `recordLookup` may need to be adjusted / inserted accordingly, if it turns out that the original override check is wrong.

Have you tried `OverridingUtil.overrides`? It also has a comment regarding with why it uses structural equivalence rather than ===.

ting-yuan (approved):
LGTM. The super type check from L289 to L294 might not be needed as well but feel free to check it later and address in another PR if needed.

Issue #203, opened 2020-12-21
[Gradle Plugin] Publish plugin marker


yigit:
Right now, we don't publish a plugin marker so developers need to update their settings file to specify the artifact.

This also makes it harder to use composite builds for testing the gradle plugin itself. (or I couldn't figure it out :) )

We should probably change the plugin to be an artifact by itself (instead of being embedded into the compiler plugin) and then also publish the marker artifact to avoid the need to change the settings.gradle.kts file.

I briefly tried this but `java-gradle-plugin` seems to have some classpath conflicts. Need further investigation.

PR #206, opened 2020-12-22
Gradle plugin publishing setup

yigit:
This CL updates the build to properly publish the gradle
plugin with an id to avoid the need to modify settings
gradle.

After this change, we'll have the following 4 artifacts:

symbol-processing-api: same as before
symbol-processing: does not include the the gradle-plugin anymore
symbol-processing-gradle-plugin: the gradle plugin
com.google.devtools.ksp.gradle.plugin: gradle plugin marker artifact

After this change, developer can enable KSP by:

id("com.google.devtools.ksp")
* Still needs to have google as a plugin repository.

The gradle plugin publishing model requires publishing a marker artifact
so that Gradle can find the main plugin artifact.
I've added `java-gradle-plugin` plugin to the gradle-plugin module to
enable publishing. Unfortunately, it does have a validation step which
fails while traversing kotlin classes in the classpath so I had to
disable it. We can enable it once gradle depends on kotlin 1.4.

As part of this, I've removed the gradle plugin from the
symbol-processing artifact. Instead, it is now published as
symbol-processing-gradle-plugin artifact (it is consistent with the
project's and androidx's naming scheme but open to changes).

For plugin id, I couldn't use symbol-processing because that would
require us to publish an artifact with symbol-processing group.
Instead, I've picked `com.google.devtools.ksp` which seems close to what
AGP does (com.android.application). This way, we can publish the marker
artifact in the KSP group.

I've made a couple more changes:
* Updated gradle to 6.6.1, which is the version kotlin uses
* Moved publishing setup to the main gradle file to configure common
things (pom, version etc)
* Moved the Jar manifest configuration to the main build file so that
all artifacts have it (was also a necessary change after unbundling the
gradle plugin since it uses that information to find the version).
Had to remove `Implementation-Title` as archiveBaseName is depreacted an
will be removed in 7.
* Added an `outRepo` command line argument to set the repository
location for publication. It is handy when testing with a local build.

Fixes: #203

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -2,11 +2,6 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 description = "Kotlin Symbol Processing API"
 
-val kspVersion: String by project
-
-group = "com.google.devtools.ksp"
-version = kspVersion
-
 tasks.withType<KotlinCompile> {
     kotlinOptions.jvmTarget = "1.8"
     kotlinOptions.freeCompilerArgs += "-Xjvm-default=compatibility"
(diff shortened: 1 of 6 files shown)


yigit:
FYI i didn't update the README since it is a major change. We should remember to update it before shipping a version with this :)

ting-yuan (approved):
Thanks! Will merge before next release.

Issue #191, opened 2020-12-16
Support members implementation parse


RinOrz:
```kotlin
object Cache {
   lateinit var member1: Handler
   lateinit var member2: Handler
}
val foo = register(source = Cache.member1)
fun bar() = register(source = Cache.member2) { ... }
...

// generated code:
object Runner {
   fun run() {
     Cache.member1.handle()
     Cache.member2.handle()
     ...
   }
}
```

Sometimes I need to do something for the incoming'source' symbol, I need to get its target type [kind: properties, type: Cache.data, returnType = Any], and then call them in the newly created source code.

Because I found that the psi element in the kotlin compiler can get these, and the symbols of ksp are also based on them, so why not built-in this feature?

In the existing ksp features, I seem to be unable to achieve the operation of the above examples, I really need this kind of parse. . .

neetopia:
Are you looking for the feature of processing the generated code? If so, we have multiple round processing feature incoming very soon, where you can process the newly generated code from previous round of processing.

ting-yuan:
@RinOrz Processing expressions isn't KSP's goal, but you can try annotating `foo` directly:

```
annotation class CacheMe(val handlerName: String)

@CacheMe("Cache.member1")
val foo = ...
```

RinOrz:
@ting-yuan Yeah sure.

But sometimes,  I may need to invoke data in the method
```kotlin
fun register(source: Any) {
   source...
}
```
In other words, it may eventually become so
```kotlin
@CacheMe("Cache.member1")
val foo = register(source = Cache.member1)

@Saved("A piece of data")
val bar = saved(data = "A piece of data")
```
Just imagine, this is very cumbersome, so for me, although ksp should not pay attention to `kt-expressions`, but it would be better if it supports this feature, because it may be useful in some places

RinOrz:
@ting-yuan Does KSP accept this kind of pr? I want to try to add this function by myself. In addition, is there standard for the `psi-element` of Kotlin? Although I am not a master in this area, I am very interested in learning about it.

ting-yuan:
We evaluated supporting expressions previously. The main concern is that KSP would be much larger and complicated. That would deviate from its goal: simple and easy to use. In fact, KSP would become pretty close to kotlinc's API and it would become another API doing the same thing :-)

ting-yuan:
If you're interested in writing a full fledged compiler plugin, here is a very good talk:
https://talkingkotlin.com/compiler-plugins/

[...]

PR #212, opened 2020-12-27
Add support for kotlin expressions.

RinOrz:
At present, I have only completed the expression parse of the Kotlin Language part (there may be some problems to be solved), java does not yet support

This is a complicated feature pr, if you agree, then I think I will continue to complete it~

Related to https://github.com/google/ksp/issues/191


RinOrz:
Oh, the situation seems a bit bad, there are some problems in the test items, I will recommit after solving.

PR #213, opened 2020-12-28
Add preliminary support for kotlin expressions

### LINKED ISSUE #191: Support members implementation parse

RinOrz:
At present, I have only completed most of the expression parse of the Kotlin Language part (it should work well), and I have used some ideas to make kotlinc's original expressions less complicated, which means that developers should use expressions easily .

This is a tedious task. Due to time and personal reasons not to use Java for the time being, the expressions of the Java language have not yet been supported. Of course, if you agree with this pr, I think I will continue to complete it~

Fixes #191  (I think even if the expression function is not the focus of KSP, it should have

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/symbol/JumpKind.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/symbol/JumpKind.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2020 Google LLC
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.google.devtools.ksp.symbol
+
+
+/**
+ * Kind of a jump keyword.
+ *
+ * @author RinOrz
+ */
+enum class JumpKind {
+    Throw, Return, Continue, Break
+}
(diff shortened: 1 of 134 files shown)


neetopia:
Hi, thanks for this PR. However, we do not plan to support expression level parsing in KSP. There are multiple reasons behind that, being performance concern, hard to maintain due to different expressions between Kotlin and Java, e.t.c.

For your case, if you really need expression level support, I would recommend to write a full compiler plugin, and you can reuse the abstraction in this PR there as well.

RinOrz:
@neetopia Oh, this is a pity, but I don’t quite understand the performance problem you mentioned. KSP is an abstract program built on top of kotlin compiler, and almost all expression properties used `by lazy`, so if the developer does not call them, the performance impact should be minimal.

Although java expressions and kotlin expressions are not exactly the same, they have similarities. Even if they are connected to a common model, I think it will not be difficult. Just like `Java annotation processors` can also parse kotlin code.

In the `JAP`, there seems to be a `TreePathScanner` that can complete the analysis of the source code. So I think KSP should also have this function, although it is not the point. Because I believe that I am not the only person who needs the function of code expression. In many excellent open source projects, I have seen that they directly call Kotlinc's api to achieve this work, I hope there is a better way to achieve this, just like [KSP makes creating lightweight compiler plugins easier](https://github.com/google/ksp#ksp-makes-creating-lightweight-compiler-plugins-easier)

In fact, I also agree with what you said about writing a complete kotlin compiler plugin. But I am not an expert in this field, and you know, after having experienced an excellent and satisfying KSP, it would be painful to abandon it and create one again. Relatively speaking, I prefer to use excellent open source projects instead of making wheels. Of course, I can also use fork branch code, but rather than often merge code from the master branch, I hope I don’t have to do anything, hahaha

neetopia:
It is more about maintenance overhead when it comes to the difference in expressions between Java and Kotlin, and we don't want to invest in a new field not supported by Java, which we also think will not be performance efficient if used by developers. As for the `TreePathScanner` you mentioned, it is actually not part of Java annotation processor API (package name not under `javax.lang.model` or `javax.annotation.processing`)

neetopia:
Closed as not in scope of KSP goal.

PR #215, opened 2020-12-29
Add generated resource dir to proper sources and outputs

ting-yuan:
So that it can be picked up by compilation and jar tasks.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -22,21 +22,17 @@ import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.UnknownTaskException
 import org.gradle.api.artifacts.Configuration
+import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.SourceSetOutput
 import org.gradle.api.tasks.TaskProvider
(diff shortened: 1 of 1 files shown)


neetopia (approved):

ZacSweers (commented):
I think the sourcing works fine, it's the task dependency that is missing. I work around this by just making KotlinCompile depend on the processResources task

ting-yuan:
Resources generated by KSP are added directly to the outputs of `kotlinCompile` and not subject to `processResources`.

On the other hand, it seems that non-generated resources are picked from `build/processedResources`, which is [hard-coded](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/kotlinCompilations.kt#L91). I thought they should be handled by kotlin-gradle-plugin automatically and it's weird to me that providing an explicit dependency affects the result.

BTW, your workaround might sometimes pick the resources generated from previous round? The dependencies look like this:
compileKotlin depends on kspKotlin
compileKotlin depends on processResources
but there's no guarantee whether processResources depends on kspKotlin or not.

ZacSweers:
true, I should update it for the new separate task. 

They might be hardcoded but that doesn't guarantee a task dependency. The fact that setting the task dependency only (and not any inputs) supports this, probably just an issue with the kotlin plugin

ting-yuan:
Ah, I didn't notice that it has been there before KSP moved to a separated task :-)

Anyway, generated resources should be picked up automatically now. Or are there still any potential issues that I missed?

Issue #216, opened 2020-12-29
Java implementation of a kotlin interface triggers illegal state exception when accessing properties


yigit:
For the given code:

```kotlin
interface MyInterface {
    var x:Int
}
```

```java
class JavaImpl implements MyInterface {
    public int getX() {
        return 1;
    }
    public void setX(int value) {
    }
}
```

Trying to get properties of the `JavaImpl` class declaration causes an exception:

```
val javaClass = invocation.kspResolver.getClassDeclarationByName("JavaImpl")!!
javaClass.getAllProperties()
```

```
unexpected psi: class org.jetbrains.kotlin.com.intellij.psi.impl.source.PsiMethodImpl
java.lang.IllegalStateException: unexpected psi: class org.jetbrains.kotlin.com.intellij.psi.impl.source.PsiMethodImpl
	at com.google.devtools.ksp.symbol.impl.UtilsKt.toKSPropertyDeclaration(utils.kt:278)
	at com.google.devtools.ksp.symbol.impl.java.KSClassDeclarationJavaImpl.getAllProperties(KSClassDeclarationJavaImpl.kt:94)
```

ting-yuan:
The fix will be available in next release, scheduled on Jan. 6.

PR #217, opened 2020-12-29
Use property descriptor impl for java properties

yigit:
When a java class implements a kotlin interface that declared a property,
the descriptor could resolve to a PsiMethod for which we don't have a
property implementation (because there is another matching method so single
method is not enough). For those cases, fall back to the descriptor implementation.

Test: allFunctions.kt
Fixes: #216


ting-yuan (approved):
Thanks for catching and fixing this non-obvious case. From Kotlin's point of view it is a property.

Issue #210, opened 2020-12-26
Incremental logging does not create a file first


ZacSweers:
If I try to enable incremental logging, it will try to append text to a non-existent file without creating it first. Here's an example enabling it via test here: https://github.com/tschuchortdev/kotlin-compile-testing/pull/96

```
e: java.io.FileNotFoundException: /var/folders/p8/cjk91csx0hn8l88w70jgph8r0000gn/T/junit15907499600375035869/ksp/build/kspDirtySet.log (No such file or directory)
	at java.base/java.io.FileOutputStream.open0(Native Method)
	at java.base/java.io.FileOutputStream.open(FileOutputStream.java:291)
	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:234)
	at kotlin.io.FilesKt__FileReadWriteKt.appendBytes(FileReadWrite.kt:115)
	at kotlin.io.FilesKt__FileReadWriteKt.appendText(FileReadWrite.kt:142)
	at kotlin.io.FilesKt__FileReadWriteKt.appendText$default(FileReadWrite.kt:142)
	at com.google.devtools.ksp.IncrementalContext.logDirtyFiles(Incremental.kt:320)
	at com.google.devtools.ksp.IncrementalContext.calcDirtyFiles(Incremental.kt:348)
```

PR #221, opened 2020-12-30
Refine incremental logs

ting-yuan:
1. mkdir() before writing logs. In some cases, `build` isn't created by
   Gradle.
2. Add a timestamp for each build.


ting-yuan:
This fixes #210

neetopia (approved):

PR #229, opened 2021-01-06
Rebuild if processing options changed

ting-yuan:


neetopia (approved):

ting-yuan:
@neetopia The original implementation doesn't work in some cases. Therefore I reimplemented it in Gradle. Please take a look again.

neetopia (approved):

PR #238, opened 2021-01-07
Model Java enum entries as KSClassDeclaration

ting-yuan:
So that it is more consistent from Kotlin's point of view.


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSClassDeclarationJavaEnumEntryImpl.kt
```diff
@@ -0,0 +1,123 @@
...
+                    .filter { (it as PropertyDescriptor).visibility != DescriptorVisibilities.INVISIBLE_FAKE }
+                    .map{ (it as PropertyDescriptor).toKSPropertyDeclaration() }
+        } ?: emptyList()
+    }
+
+    override val declarations: List<KSDeclaration> = emptyList()
+
+    override val modifiers: Set<Modifier> by lazy {
+        val modifiers = mutableSetOf<Modifier>()
+        modifiers.addAll(psi.toKSModifiers())
```
neetopia, line 90: can be replaced with simply `mutableSetOf(psi.toKSModifiers())`

neetopia (approved):

neetopia (approved):

Issue #244, opened 2021-01-11
JavaCompile task fails with no sources


ZacSweers:
I'm unable to compile with the latest version of KSP (`20200107`), it seems like it tries to force the `JavaCompile` task to run even if there are no java source files generated.

```
* What went wrong:
Execution failed for task ':moshi-ksp:moshi-ksp:compileJava'.
> no source files
```

Repro: https://github.com/ZacSweers/MoshiX/pull/75

One possible clue might be the auto-service-ksp plugin generating resources, which I know got some work in this release.

@gavra0 might have some advice for this, as he fixed a similar issue in AGP awhile back for kotlin-only Android projects.

gavra0:
This is basically the same issue as https://youtrack.jetbrains.com/issue/KT-31127.

PR #245, opened 2021-01-11
Do not register resources as Java sources

gavra0:
JavaCompile task will process only *.java files,
and resources need to registered using different
mechanism.

This fixes https://github.com/google/ksp/issues/244.


ZacSweers (approved):

ting-yuan:
Thanks, will try to add resources in a proper way.

PR #255, opened 2021-01-20
close output stream after processing for code generator

neetopia:


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -65,6 +65,9 @@ class KotlinSymbolProcessingExtension(
 
 abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions, val logger: KSPLogger, val testMode: Boolean) :
     AnalysisHandlerExtension {
+
+    lateinit var codeGenerator: CodeGeneratorImpl
+
```
ting-yuan, line 70: Files need to be closed before each round, otherwise subsequent rounds may or may not observe the contents.

neetopia: Actually the existing code still works. In multiple round, every round still exits at `analysisComplete`

ting-yuan: I see. Another problem: `incrementalContext.updateCachesAndOutputs` backs up the outputs and therefore outputs need to be closed before it.

May I know why moving `codeGenerator` from a local to a member and closing it in `analysisComplete`?

neetopia: I see, in that case I can move it back. The reason I want to put it in `analysisComplete` is that I can later try to not instantiate a code generator every round.

ting-yuan (approved):

PR #258, opened 2021-01-20
Rename reports of integration tests

ting-yuan:
Unfortunately I forgot to change the name of upload when copying the entries from test and the report of test was overwritten.


ting-yuan:
Tested in
https://github.com/ting-yuan/ksp/actions/runs/499279720

PR #260, opened 2021-01-20
Traversal order

yigit:
Prototype for keeping declaration order in .class files that come from compiled kotlin code.


yigit:
this ended up requiring producing jvm signatures, which is not a great added cost and does not feel great to turn on by default.
I thought about making it an optional parameter OR possibly, provide another method that will sort the descriptors.
e.g.

```
KSClassDeclaration.getDeclarationsInDeclarationOrder()
```
which developer needs to call? That could help avoid the cost in generic case but let processor owners call it when needed.

PR #269, opened 2021-01-25
Add integration tests for plugin

yigit:
This PR adds a testing infra for the gradle plugin and also
adds support for android variants.

PR #270, opened 2021-01-26
Copy compiler args from compileKotlin to kspKotlin

ting-yuan:
Will add a test after #269 is merged.


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -133,7 +141,10 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kspTask.setDestinationDir(destinationDir)
             kspTask.mapClasspath { kotlinCompileProvider.get().classpath }
```
yigit, line 142: should we update this to also delegate in the task?

ting-yuan: Done. Thanks!

chrisjenx:
Just wondering what your release cadence is? (looks like almost every weekish?) looking to try this fix out :)

ting-yuan:
The release is generally bi-weekly and the next one will be on Feb.3.

dtdharan (approved):

PR #275, opened 2021-01-28
Fix constructor list

yigit:
This PR ensures that, whatever we report for java/kotlin sources match what we would report if it was in a `.class` file.
Major changes:
* added `CONSTURCTOR` to function types
* changed `getConstructors` to simply use `declaredFunctions.filter { it.kind == CONSTURCTOR }` for consistency
* updated KSClassDeclarationDescriptorImpl.kt `declarations` to include constructors
* changed KSFunctionDeclarationImpl and KSFunctionDeclarationJavaImpl to return `<init>` from name if it is constructor for consistency
*


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSFunctionDeclarationJavaImpl.kt
```diff
@@ -81,15 +86,29 @@ class KSFunctionDeclarationJavaImpl private constructor(val psi: PsiMethod) : KS
...
-        } else {
-            null
+        when {
+            psi.returnType != null -> {
+                KSTypeReferenceJavaImpl.getCached(psi.returnType!!)
+            }
+            psi.isConstructor -> {
+                psi.containingClass?.let { containingClass ->
+                    KSTypeReferenceLiteJavaImpl.getCached(
+                        KSClassDeclarationJavaImpl.getCached(containingClass).asType(emptyList())
```
ting-yuan, line 96: `asType(emptyList())` is undocumented and leaks default type, which is also undocumented. We should probably block it (by checking the size) and encourage users to use `startProejctedType` when type arguments are irrelevant.

ting-yuan (commented):
Overall lgtm, but do we really need the CONSTRUCTOR enum entry? which overlaps with `MEMBER`. Would `KSFunctionDeclaration.isConstructor() = name == "<init>"` serve the purpose?

yigit:
> Overall lgtm, but do we really need the CONSTRUCTOR enum entry? which overlaps with `MEMBER`. Would `KSFunctionDeclaration.isConstructor() = name == "<init>"` serve the purpose?

yea i had no idea whether it makes sense to add the type. i'll add `isConstructor()` instead and remove the new type.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSFunctionDeclarationJavaImpl.kt
yigit: is there a different function i should use here?
I'm trying to get a type refernce to the PsiClass cheapest way possible.

ting-yuan: Please try `asStarProjectedType()`, which is slightly more expensive than `asType(emptyList()` but well defined.

yigit: thanks.

yigit:
ok updated all tests manually.
for some, i just ignored init methods.
for some, updated the test to accept them and/or also make test output more clear.

Also fixed a bug in this current PR where we were unable to resolve descriptor for synthetics created for java sourcues.

Last but not least, sincere apologies for making this PR so large. I wanted to keep it small but one thing opened another :/. Reviewing commit by commit might be easier due to large change in tests (all test changes are in 1 commit except for the java modifiers change)

ting-yuan (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/synthetic/KSConstructorSyntheticImpl.kt
```diff
@@ -68,7 +73,14 @@ class KSConstructorSyntheticImpl(val ksClassDeclaration: KSClassDeclaration) : K
         ksClassDeclaration.location
     }
 
-    override val modifiers: Set<Modifier> = emptySet()
+    override val modifiers: Set<Modifier> by lazy {
+        // add public if parent class is public
+        if (ksClassDeclaration.isPublic()) {
+            setOf(Modifier.FINAL, Modifier.PUBLIC)
```
neetopia, line 79: This need some attention. If we set a class to be private, other classes in same file can still access its constructor, but if we set its constructor to be private, then other classes in same file won't be able to access its constructor. 

i.e. constructor's `private` modifier does not work as its corresponding class'

yigit: i'm not setting it to private here though, only setting it to public if class is public and if not, just not adding anything.

neetopia: Yeah, this actually still works because kotlin treat default to be public. The point is a private class will have a different modifier behavior than its public class counterpart, while in fact both are public. A minor comment though.

yigit: this part definitely needs a followup (recalling the delegation to descriptors discussion we head earlier)

Is there any changes I should make here or is it good enough for now?

neetopia (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/synthetic/KSConstructorSyntheticImpl.kt
neetopia: Merged.

PR #278, opened 2021-01-30
Workaround for some unchecked accesses

ting-yuan:
There are some resolutions that don't lookup binding context and rewrote
the slice. Therefore, the annotation resolution is rewritten to use the
same call path as those in-compiler accesses to avoid the conflict.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -523,16 +523,16 @@ class ResolverImpl(
 
     fun resolveAnnotationEntry(ktAnnotationEntry: KtAnnotationEntry): AnnotationDescriptor? {
         bindingTrace.get(BindingContext.ANNOTATION, ktAnnotationEntry)?.let { return it }
-        val containingDeclaration = KtStubbedPsiUtil.getContainingDeclaration(ktAnnotationEntry)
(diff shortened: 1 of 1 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -523,16 +523,16 @@ class ResolverImpl(
...
-            bindingTrace.get(BindingContext.ANNOTATION, ktAnnotationEntry)
-        } else {
-            ktAnnotationEntry.findLexicalScope().let { scope ->
-                annotationResolver.resolveAnnotationsWithArguments(scope, listOf(ktAnnotationEntry), bindingTrace)
-                bindingTrace.get(BindingContext.ANNOTATION, ktAnnotationEntry)
+        KtStubbedPsiUtil.getContainingDeclaration(ktAnnotationEntry)?.let { containingDeclaration ->
+            if (KtPsiUtil.isLocal(containingDeclaration)) {
+                resolveDeclarationForLocal(containingDeclaration)
+            } else {
+                resolveSession.resolveToDescriptor(containingDeclaration).annotations.forEach {}
```
neetopia, line 530: What is this empty `forEach()` for? Is it to make the annotation entry to be looked up in binding context?

ting-yuan: yes, it is similar to `forceResolveAll()` but less expensive.

neetopia (approved):

Issue #279, opened 2021-01-30
ksp returns empty constructor for data classes


yigit:
sample input:

```
data class Subject(
    var x:Int
)
```
for some reason, this is returning only 1 constructor and it is synthetic.
this is happening on master, likely because of https://github.com/google/ksp/pull/275.
I'll try to repro in ksp

PR #280, opened 2021-01-30
return primary constructor in declarations for KSClassDeclarationImpl

yigit:
PR #275 made different class declaration backends more consistent
by adding constructor into declarations and making getConstructors
return based on that.
Unfortunately, it missed primary constructor, causing missing
constructors OR even worse, return an empty constructor
instead of the primary constructor.

This PR fixes that issue and also extends the test case to
cover it.

Fixes: #279

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSClassDeclarationImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSClassDeclarationImpl.kt
@@ -60,8 +60,13 @@ class KSClassDeclarationImpl private constructor(val ktClassOrObject: KtClassOrO
         val propertiesFromConstructor = primaryConstructor?.parameters
             ?.filter { it.isVar || it.isVal }
             ?.map { KSPropertyDeclarationParameterImpl.getCached((it as KSValueParameterImpl).ktParameter) } ?: emptyList()
-        val result = ktClassOrObject.declarations.getKSDeclarations().toMutableList()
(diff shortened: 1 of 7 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -60,8 +60,13 @@ class KSClassDeclarationImpl private constructor(val ktClassOrObject: KtClassOrO
         val propertiesFromConstructor = primaryConstructor?.parameters
             ?.filter { it.isVar || it.isVal }
             ?.map { KSPropertyDeclarationParameterImpl.getCached((it as KSValueParameterImpl).ktParameter) } ?: emptyList()
-        val result = ktClassOrObject.declarations.getKSDeclarations().toMutableList()
+        val result = ktClassOrObject.declarations.getKSDeclarations().toMutableList<KSDeclaration>()
         result.addAll(propertiesFromConstructor)
+        primaryConstructor?.let { primaryConstructor: KSDeclaration ->
```
neetopia, line 65: How about using the origin of `primaryConstructor` to determine if it needs to be added into the return result?

yigit: you mean `if (primaryConstructor.origin == Origin.KOTLIN) {` ? if so, it seems to be working fine (passes tests and behaves the same for tests we have).

neetopia (approved):

PR #281, opened 2021-02-01
Update documents

ting-yuan:
1. Add an inline link in the main doc to the merged why-ksp.md and comparisons.md.
2. Updated class diagram.
3. Various updates and cleanups including API changes and examples.


neetopia (approved):
LGTM

PR #282, opened 2021-02-02
Register generated java sources with AGP

yigit:
This PR adds code to register generated java sources with AGP so that
the IDE can discover them (useful when developer needs to debug
generated code or their code needs to access them).

I had to update AGP dependency to the plugin (instead of the API)
because the plugin does not yet provide the api to register generated
resource folders.

For some reason, the generated byte code sources gets overridden by the
kotlin path, hence, I couldn't assert them in tests.

Test: SourceSetConfigurationTest.registerJavaSourcesToAndroid

### DIFF
--- a/gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/SourceSetConfigurationsTest.kt
+++ b/gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/SourceSetConfigurationsTest.kt
@@ -17,8 +17,8 @@
 package com.google.devtools.ksp.gradle
 
 import com.google.common.truth.Truth.assertThat
-import com.google.devtools.ksp.gradle.testing.DependencyDeclaration.Companion.module
 import com.google.devtools.ksp.gradle.processor.TestSymbolProcessor
+import com.google.devtools.ksp.gradle.testing.DependencyDeclaration.Companion.module
 import com.google.devtools.ksp.gradle.testing.KspIntegrationTestRule
 import com.google.devtools.ksp.gradle.testing.PluginDeclaration
 import com.google.devtools.ksp.processing.Dependencies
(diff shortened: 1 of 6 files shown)


gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/SourceSetConfigurationsTest.kt
```diff
@@ -69,6 +69,101 @@ class SourceSetConfigurationsTest {
...
+        testRule.appModule.buildFileAdditions.add(
+            """
+            tasks.register("printSources") {
+                fun logVariantSources(variants: DomainObjectSet<out com.android.build.gradle.api.BaseVariant>) {
+                    variants.all {
+                        println("VARIANT:" + this.name)
+                        val baseVariant = (this as com.android.build.gradle.internal.api.BaseVariantImpl)
+                        val variantData = baseVariant::class.java.getMethod("getVariantData").invoke(baseVariant)
+                            as com.android.build.gradle.internal.variant.BaseVariantData
+                        variantData.extraGeneratedSourceFolders.forEach {
```
gavra0, line 91: You can get all Java sources with `BaseVariant.getSourceFolders(SourceKind.JAVA)` and java compile classpath with ``BaseVariant.getCompileClasspath(null)`. It should be more stable across versions.

yigit: when i was trying this, it was not included in `getSourceFolders` so i assumed it is only for regular sources, not generated sources.

gavra0: kk, this seems ok, and if the AGP API changes, it should be easy to migrate.

gavra0 (approved):

PR #295, opened 2021-02-06
Check KSP and kotlin-gradle-plugin versions

ting-yuan:
Each KSP release depends on a specific Kotlin compiler version.
If their versions don't match, there might be runtime errors.

### DIFF
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/VersionCheckIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/VersionCheckIT.kt
@@ -1,0 +1,22 @@
+package com.google.devtools.ksp.test
+
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+import java.io.File
+import java.util.jar.*
+
+class VersionCheckIT {
+    @Rule
+    @JvmField
+    val project: TemporaryTestProject = TemporaryTestProject("playground")
+
+    @Test
+    fun testVersion() {
+        val gradleRunner = GradleRunner.create().withProjectDir(project.root)
+        val result = gradleRunner.withArguments("-Dkotlin.compiler.execution.strategy=in-process", "-PkotlinVersion=1.4.20", "clean", "build").buildAndFail()
+        Assert.assertTrue(result.output.contains("Please pick the same version of ksp and kotlin plugins."))
+    }
+}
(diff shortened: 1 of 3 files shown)


yigit (changes requested):
Can this be a warning instead of an error? E.g. for a long time we used 1.4.21 in android X with ksp 1.4.20

neetopia:
I think this is already a warning. It is Gradle plugin returns not applicable to disable KSP.

ting-yuan:
According to [Kotlin Releases](https://kotlinlang.org/releases.html), `y` indicate feature changes and `z` indicate bug fixes in version `1.x.yz`. Let's accept different `z` and reject different `y`.

ting-yuan:
Well, I take that back. Incremental version bumps don't necessarily break KSP. Bug-fix version bumps don't guarantee the compatibility of KSP. If we are not going to enforce strict version checks, it's probably better to let users to decide.

yigit:
> Well, I take that back. Incremental version bumps don't necessarily break KSP. Bug-fix version bumps don't guarantee the compatibility either. If we are not going to enforce strict version checks, it's probably better to let users to decide.

I did something similar in AndroidX infra: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/CompilationTestCapabilities.kt;l=58?q=compilationtestcapa&ss=androidx

But yea, it is all internals so hard to ensure if it will break KSP or not (thats why it is good to have a warning though)
Maybe, we should also reach out to JB to get a KSP pre-submit in kotlin repo so they can avoid breaking it when possible ?

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/VersionCheckIT.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class VersionCheckIT {
+    @Rule
+    @JvmField
+    val project: TemporaryTestProject = TemporaryTestProject("playground")
+
+    @Test
+    fun testVersion() {
+        val gradleRunner = GradleRunner.create().withProjectDir(project.root)
+        val result = gradleRunner.withArguments("-Dkotlin.compiler.execution.strategy=in-process", "-PkotlinVersion=1.4.20", "clean", "build").buildAndFail()
```
yigit, line 19: this test needs to change to assert the warning instead of error ?

yigit (approved):

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/VersionCheckIT.kt
ting-yuan: Latest ksp (1.4.30 and later) isn't compatible with kotlin-1.4.20, so the expected build result is fail. At least the warning shows up before the failure so we can still check for that.

ting-yuan:
> Maybe, we should also reach out to JB to get a KSP pre-submit in kotlin repo so they can avoid breaking it when possible ?

We had a similar discussion with JetBrains (about upstreaming) and we agreed that it might not be the best time to do it yet because both are under massive changes / developments.

PR #302, opened 2021-02-10
add integration test for onError

neetopia:
I tried to test the `onError` behavior with logger.warn but it seems warnings are shadowed in case of errors.. Probably need another fix to make it work properly.

### DIFF
--- a/integration-tests/src/test/resources/on-error/on-error-processor/build.gradle.kts
+++ b/integration-tests/src/test/resources/on-error/on-error-processor/build.gradle.kts
@@ -1,0 +1,24 @@
+val kspVersion: String by project
+val testRepo: String by project
+
+plugins {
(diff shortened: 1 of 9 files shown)


integration-tests/src/test/resources/on-error/on-error-processor/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+
+repositories {
+    maven(testRepo)
+    mavenCentral()
+    google()
+}
+
+dependencies {
+    implementation(kotlin("stdlib"))
+    implementation("com.squareup:javapoet:1.12.1")
```
ting-yuan, line 19: Seems unnecessary.

ting-yuan (approved):
LGTM with a minor comment.

PR #309, opened 2021-02-12
Update Why KSP

dwiner:
Use Room comparison instead of Glide to align with README and other public docs.


neetopia (commented):
Should [this paragraph](https://github.com/google/ksp/blame/master/docs/why-ksp.md#L83) be changed as well in this case?

dwiner:
> Should [this paragraph](https://github.com/google/ksp/blame/master/docs/why-ksp.md#L83) be changed as well in this case?

Sorry for the delay on this. You're right. I think it may make sense to hold off on updating this until we're able to share the raw numbers from Room, though, since ideally we want a comparable paragraph with the actual numbers for the amount of time spent on each step.

jaschdoc:
Hi, I'm closing this PR as it is no longer relevant. Documentation has moved to the Kotlin website.

PR #315, opened 2021-02-14
Make finish() have a default implementation

ZacSweers:
Per chat in the kotlin-lang channel, this function often doesn't need implementing in processors and saves users a bit of boilerplate.


neetopia (commented):
Since you are also updating tests, can you also update tests under /integration-tests/src/test/resources and gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/processor/TestSymbolProcessor.kt ?

ZacSweers:
I believe I covered all those (for example - https://github.com/google/ksp/pull/315/files#diff-6d671b7a6ee4792d1ceeb0a5cd440dec5858ee9a15235825559a5905c88520a9L44). Did you see any that I missed?

neetopia:
Oh, sorry I must have been wrong. Merged, thanks for PR!

PR #316, opened 2021-02-17
Implement getSealedSubclasses API on KSClassDeclaration

### LINKED ISSUE #218: Provide getSealedSubclasses
It is fairly popular and can be optimized a little bit. May also worth integration into the whole text scan, originally for getClassDeclarationByName, when [KEEP 225](https://github.com/Kotlin/KEEP/pull/225) rolls out.

ZacSweers:
Resolves #218

Starting early for review. Next steps:
* Do we need to add in incremental tracking now or wait till Kotlin 1.5?
* Do we expose the same for Java 15 "permitted subtypes" if JVM target is set to 15+ and 1.5 language version is set?
* Tests!

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSClassDeclarationImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSClassDeclarationImpl.kt
@@ -27,9 +27,11 @@ import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.symbol.impl.*
 import com.google.devtools.ksp.symbol.impl.binary.getAllFunctions
 import com.google.devtools.ksp.symbol.impl.binary.getAllProperties
+import com.google.devtools.ksp.symbol.impl.binary.sealedSubclassesSequence
 import com.google.devtools.ksp.symbol.impl.synthetic.KSConstructorSyntheticImpl
 import org.jetbrains.kotlin.descriptors.DescriptorVisibilities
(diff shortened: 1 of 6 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -52,6 +54,10 @@ class KSClassDeclarationImpl private constructor(val ktClassOrObject: KtClassOrO
         (ktClassOrObject is KtObjectDeclaration) && (ktClassOrObject.isCompanion())
     }
 
+    override fun getSealedSubclasses(): Sequence<KSClassDeclaration> {
+        return descriptor.sealedSubclassesSequence()
```
ting-yuan, line 58: Assuming that 99% of the classes are not sealed, can we check the modifier to return early without having to resolve to descriptor? e.g.,
```
return if (ktClassOrObject.hasModifier(KtTokens.SEALED_KEYWORD) {
    descriptor.sealedSubclassesSequence()
} else {
    emptySequence()
}
```

ting-yuan (approved):
Thanks, LGTM with a minor optimization nitpicking. This may work with incremental processing out of box. Therefore I'd like to do some experiments and test it a little bit. Will try to merge this before next release.

ting-yuan:
> This may work with incremental processing out of box. 

Unfortunately the IC support in compiler is implemented explicitly rather than using lookup record, probably for better granularity.
https://github.com/JetBrains/kotlin/commit/36f99156fd140f1ef58c7763f88b5bca18926b15

We need to fix this before Kotlin-1.5. Tracking in #323.

PR #321, opened 2021-02-18
triage exception for KSP errors and processor exceptions

neetopia:


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -129,7 +132,18 @@ abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions,
...
+            } catch (e: Exception) {
+                // Throws KSP exceptions
+                if (e.stackTrace.first().className.startsWith("com.google.devtools.ksp")) {
+                    throw e
+                } else {
+                    val sw = StringWriter()
+                    e.printStackTrace(PrintWriter(sw))
+                    logger.error(sw.toString())
+                }
+            }
```
ting-yuan, line 146: We probably need to catch all calls to processors, such as `init`, `finish` and even `onError`.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -204,6 +215,21 @@ abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions,
     private fun KSPLogger.hasError(): Boolean {
         return (this as MessageCollectorBasedKSPLogger).recordedEvents.any { it.severity == CompilerMessageSeverity.ERROR || it.severity == CompilerMessageSeverity.EXCEPTION }
     }
+
+    private fun handleException(call: () -> Unit) {
+        try {
+            call()
+        } catch (e: Exception) {
+            // Throws KSP exceptions
+            if (e.stackTrace.first().className.startsWith("com.google.devtools.ksp")) {
```
ting-yuan, line 224: Can we avoid the hard coded string, or define the constant in a testable (by reflection) function / variable.

docs/multi-round.md
```diff
@@ -62,11 +62,20 @@ A processor instance is created only once, which means you can store information
 
 All KSP symbols will not be reusable across multiple rounds, as the resolution result can potentially change based on what was generated in a previous round. However, since KSP does not allow modifying existing code, some information such as the string value for a symbol name should still be reusable. To summarize, processors can store information from previous rounds but need to bear in mind that this information might be invalid in future rounds.
 
-#### Error Handling
-When an error occurs (defined by processor calling `KSPLogger.error()`), processing stops after the current round completes. All processors will call the `onError()` method and will **not** call the `finish()` method.
+#### Error and Exception Handling
+When an error(defined by processor calling `KSPLogger.error()`) or exception occurs, processing stops after the current round completes. All processors will call the `onError()` method and will **not** call the `finish()` method.
```
alexvanyo, line 66: One corner case I thought of looking at this: If `finish()` throws an exception or logs an error, what is the expected behavior?

Will `onError()` be called after `finish()` throws or logs an error?

Looking at the proposed implementation, it looks like

```kotlin
incrementalContext.updateCachesAndOutputs(dirtyFiles, codeGenerator.outputs, codeGenerator.sourceToOutputs)
```

will still be called if `finish()` throws or logs an error.

neetopia: Good catch. Currently if `finish` throws an exception, `onError` will still be called. However, regardless of whether we want this behavior or not, incremental data should not be updated upon errors.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -204,6 +223,21 @@ abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions,
     private fun KSPLogger.hasError(): Boolean {
         return (this as MessageCollectorBasedKSPLogger).recordedEvents.any { it.severity == CompilerMessageSeverity.ERROR || it.severity == CompilerMessageSeverity.EXCEPTION }
     }
+
+    private fun handleException(call: () -> Unit) {
+        try {
+            call()
+        } catch (e: Exception) {
```
ting-yuan, line 230: This alone will break things using [Kotlin Compile Testing](https://github.com/tschuchortdev/kotlin-compile-testing), where people throw `CompilationErrorException` to make `CLICompiler` return `COMPILATION_ERROR`.

Instead of returning `AnalysisResult.success`, we should return `AnalysisResult.compilationError` when there are errors.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -189,7 +204,7 @@ abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions,
         if (finished) {
             logger.reportAll()
         }
-        return AnalysisResult.EMPTY
+        return if (logger.hasError()) AnalysisResult.compilationError(BindingContext.EMPTY) else AnalysisResult.EMPTY
```
ting-yuan, line 207: `analysisCompleted` needs to be updated or removed (and move its logic here) as well.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -170,13 +179,19 @@ abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions,
...
             }
         } else {
             if (finished) {
                 processors.forEach {
                     handleException {
                         it.finish()
                     }
+                    if (hasInternalError) {
+                        return AnalysisResult.internalError(BindingContext.EMPTY, internalError)
+                    }
```
ting-yuan, line 194: `hasInternalError` can be saved by making `handleException` return. e.g.,
```
handleException {
    ...
}?.let { return it }
```

neetopia: This will result in a `return null` in handleException though, I am not sure if this is better.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -128,16 +135,23 @@ abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions,
...
+                }?.let { return it }
+                deferredSymbols[processor] = mutableListOf()
             }
             initialized = true
         }
-        processors.forEach {
-            deferredSymbols[it] = it.process(resolver)
-            if (!deferredSymbols.containsKey(it) || deferredSymbols[it]!!.isEmpty()) {
-                deferredSymbols.remove(it)
+        processors.forEach processing@{ processor ->
```
ting-yuan, line 146: Should we check `logger.hasError()` here, in case some processor fail in `init()`?

api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessor.kt
```diff
@@ -22,6 +22,16 @@ import com.google.devtools.ksp.symbol.KSAnnotated
 
 /**
  * [SymbolProcessor] is the interface used by plugins to integrate into Kotlin Symbol Processing.
+ * SymbolProcessor supports multiple round execution, a processor may return a list of deferred symbols at the end
+ * of every round, which will be passed to proceesors again in the next round, together with the newly generated symbols.
+ * Upon Exceptions, KSP will try to distinguish the exceptions from KSP and exceptions from processors.
+ * Exceptions from KSP will result in a termination of processing immediately, while exception from a processor
+ * will result in the termination of the only processor throwing exception, and be logged as an error in KSPLogger.
```
ting-yuan, line 29: Aren't all subsequent processors skipped?

api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessor.kt
```diff
@@ -22,6 +22,16 @@ import com.google.devtools.ksp.symbol.KSAnnotated
...
+ * Exceptions from KSP should be reported to KSP developers for further investigation.
+ * In the event of processor exceptions or error event (defined as processor calling KSPLogger.error(),
+ * all other processors will continue processing to complete the round of processing.
+ * At the end of the round where exceptions or errors happened, all processors will invoke onError() function to do
```
ting-yuan, line 33: and here

ting-yuan (approved):

PR #326, opened 2021-02-22
Add doc how to setup with groovy

bohsen:
# Overview
Update quickstart.md with block explaining how to setup using groovy

# Proposed change

Add collapsible blocks with explanation on how to setup the ksp plugin in a project that uses groovy


ting-yuan (approved):
Thanks!

PR #328, opened 2021-02-23
Update sourceToOutputsMap before restoring outputs

### LINKED ISSUE #327: Outputs are not deleted for corresponding deleted sources

ting-yuan:
Clean files should be calculated with the up-to-date sourceToOutputsMap.

Also add reprocessed sources and outputs in `kspSourceToOutputsMap.log`.

Fixes #327

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
@@ -380,7 +380,7 @@ class IncrementalContext(
         logSourceToOutputs()
 
         sourceToOutputsMap.flush(false)
-        sourceToOutputsMap.close()
+        // Don't close the map yet. It'll be used to calculate clean outputs.
     }
 
     // TODO: Recover if processing failed.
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -377,13 +392,12 @@ class IncrementalContext(
             sourceToOutputs[source]?.let { sourceToOutputsMap[source] = it} ?: sourceToOutputsMap.remove(source)
         }
 
-        logSourceToOutputs()
+        logSourceToOutputs(outputs, sourceToOutputs)
 
         sourceToOutputsMap.flush(false)
-        sourceToOutputsMap.close()
+        // Don't close the map yet. It'll be used to calculate clean outputs.
```
neetopia, line 398: Might not need this comment, it looks this is not the correct place to close the map.

neetopia (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
ting-yuan: Will remove in a following PR.

PR #329, opened 2021-02-23
Keep lastModified timestamp when restoring incremental backups

ting-yuan:


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/OutputDepsIt.kt
```diff
@@ -82,6 +111,16 @@ class OutputDepsIt {
                 Assert.assertEquals(TaskOutcome.SUCCESS, result.task(":workload:kspKotlin")?.outcome)
                 val dirties = result.output.split("\n").filter { it.startsWith("w: [ksp]") }.toSet()
                 Assert.assertEquals(expectedDirties, dirties)
+
+                val outputRoot = File(project.root, "workload/build/generated/ksp/main/")
+                outputRoot.walk().filter { it.isFile() }.forEach {
+                    // Allow equal in the comparison in case KSP is tooooo fast.
```
yigit, line 117: could actually make sense to add a delay in test to ensure they are never the same, so we know if it was re-written or not?

ting-yuan: Done. The precision of timestamp varies among different file systems / platforms. I guess 10ms should be a safe assumption for modern systems.

neetopia (approved):

PR #333, opened 2021-02-26
Fix delimiter used to pass changed files

ting-yuan:
WIP; test is on the way.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingPlugin.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingPlugin.kt
@@ -66,8 +66,8 @@ class KotlinSymbolProcessingCommandLineProcessor : CommandLineProcessor {
             val (k, v) = value.split('=', ignoreCase = false, limit = 2)
             processingOptions.put(k, v)
         }
-        KspCliOption.KNOWN_MODIFIED_OPTION -> knownModified.addAll(value.split('=').map { File(it) } )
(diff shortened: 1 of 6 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingPlugin.kt
```diff
@@ -66,8 +66,8 @@ class KotlinSymbolProcessingCommandLineProcessor : CommandLineProcessor {
             val (k, v) = value.split('=', ignoreCase = false, limit = 2)
             processingOptions.put(k, v)
         }
-        KspCliOption.KNOWN_MODIFIED_OPTION -> knownModified.addAll(value.split('=').map { File(it) } )
-        KspCliOption.KNOWN_REMOVED_OPTION -> knownRemoved.addAll(value.split('=').map { File(it) } )
+        KspCliOption.KNOWN_MODIFIED_OPTION -> knownModified.addAll(value.split(':').map { File(it) } )
```
yigit, line 69: is this the path separator? if so, i think we should be using: https://docs.oracle.com/javase/8/docs/api/java/io/File.html#pathSeparatorChar

neetopia: I think it is the separator used [here](https://github.com/google/ksp/blob/master/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L318-L319)

ting-yuan: Done.

neetopia (approved):

PR #338, opened 2021-03-04
prototype for relying on descriptor for property getters/setters

yigit:
This is not all roses:
* we lose synthetic origin for these cases (instead they become class)
* find annotation from use site target does not work (though they correctly get placed on accessors)
* annotations on kotlin source properties that target accessors show up in the annotations list of both property and accessor

### DIFF
--- a/compiler-plugin/testData/api/implicitElements.kt
+++ b/compiler-plugin/testData/api/implicitElements.kt
@@ -22,10 +22,10 @@
 // <null>
 // <null>
 // <init>,<init>,<init>
-// readOnly.get(): CLASS annotations from property: GetAnno
-// readOnly.getter.owner: readOnly: CLASS
+// readOnly.get(): SYNTHETIC annotations from property: GetAnno
+// readOnly.getter.owner: readOnly: KOTLIN
 // readWrite.get(): KOTLIN
-// readWrite.set(): CLASS annotations from property: SetAnno
+// readWrite.set(): SYNTHETIC annotations from property: SetAnno
 // <init>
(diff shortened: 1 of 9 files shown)


compiler-plugin/testData/api/implicitElements.kt
```diff
@@ -22,10 +22,10 @@
 // <null>
 // <null>
 // <init>,<init>,<init>
-// readOnly.get(): SYNTHETIC annotations from property: GetAnno
-// readOnly.getter.owner: readOnly: KOTLIN
+// readOnly.get(): CLASS annotations from property: GetAnno
+// readOnly.getter.owner: readOnly: CLASS
```
neetopia, line 26: This does not look right, as the owner should be `KOTLIN`

neetopia (commented):
I think the point for having synthetic is to indicate this is not directly in source code, but compiler will synthetic one which will be present in compiled code. Therefore I am fine with the source being synthetic, and it will be different when reading from a library, but we shouldn't worry too much about that, since reading from library already makes it impossible to restore the information from the source it compiled from.

compiler-plugin/testData/api/implicitElements.kt
yigit: i think i can fix this by still using the synthetic but constructing it w/ the descriptor.
It might be a bit weird because we already have the descriptor when i create it but synthetic only expects the ktProperty, i can try to see how i can avoid the re-query there.

neetopia (approved):

PR #341, opened 2021-03-05
Always include the main `ksp` configuration

ting-yuan:
With MPP, `KotlinPluginWrapper` isn't always applied and the source set
`main` may not exist.


ting-yuan:
I'm not sure whether we should change from `KotlinPluginWrapper` to `KotlinBasePluginWrapper` or not. Therefore the proposed PR is rather conservative.

gavra0 (approved):

ting-yuan:
Thanks for reviewing!

PR #343, opened 2021-03-06
remove findAnnotationFromUseSiteTarget public function, integrate rel…

neetopia:
…ated logic into KSPropertyAccessor and KSPropertyDeclaration

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSPropertyDeclarationImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSPropertyDeclarationImpl.kt
@@ -29,6 +29,9 @@ import com.google.devtools.ksp.symbol.impl.synthetic.KSPropertyGetterSyntheticIm
 import com.google.devtools.ksp.symbol.impl.synthetic.KSPropertySetterSyntheticImpl
 import org.jetbrains.kotlin.descriptors.PropertyDescriptor
 import org.jetbrains.kotlin.descriptors.VariableDescriptorWithAccessors
+import org.jetbrains.kotlin.descriptors.annotations.AnnotationUseSiteTarget
+import org.jetbrains.kotlin.psi.KtAnnotated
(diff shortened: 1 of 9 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSPropertyDeclarationImpl.kt
```diff
@@ -42,6 +43,13 @@ class KSPropertyDeclarationImpl private constructor(val ktProperty: KtProperty)
         ResolverImpl.instance.resolveDeclaration(ktProperty) as? PropertyDescriptor
     }
 
+    override val annotations: List<KSAnnotation> by lazy {
+        ktProperty.annotationEntries.filter {
+            it.useSiteTarget?.getAnnotationUseSiteTarget() != AnnotationUseSiteTarget.PROPERTY_GETTER
+                    && it.useSiteTarget?.getAnnotationUseSiteTarget() != AnnotationUseSiteTarget.PROPERTY_SETTER
```
ting-yuan, line 49: Can we call `it.useSiteTarget?.getAnnotationUseSiteTarget()` only once?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSPropertyDeclarationParameterImpl.kt
```diff
@@ -41,6 +42,13 @@ class KSPropertyDeclarationParameterImpl private constructor(val ktParameter: Kt
         fun getCached(ktParameter: KtParameter) = cache.getOrPut(ktParameter) { KSPropertyDeclarationParameterImpl(ktParameter) }
     }
 
+    override val annotations: List<KSAnnotation> by lazy {
+        ktParameter.annotationEntries.filter {
```
ting-yuan, line 46: Can we share the code with the above?

ting-yuan (approved):

PR #347, opened 2021-03-10
change getJvmName() and mapToJvmSignature() to return null when unabl…

neetopia:
…e to resolve

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/MangledNamesProcessor.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/MangledNamesProcessor.kt
@@ -13,11 +13,11 @@ class MangledNamesProcessor : AbstractTestProcessor() {
     override fun toResult() = results
 
     override fun process(resolver: Resolver): List<KSAnnotated> {
-        val mangleSourceNames = mutableMapOf<String, String>()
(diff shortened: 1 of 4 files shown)


ting-yuan:
Can you comment on situations where the resolution may fail for a function / property declaration?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processor/MangledNamesProcessor.kt
```diff
@@ -60,17 +60,17 @@ class MangledNamesProcessor : AbstractTestProcessor() {
         override fun visitFunctionDeclaration(function: KSFunctionDeclaration, data: MutableMap<String, String>) {
             if (function.simpleName.asString() in IGNORED_FUNCTIONS) return
             super.visitFunctionDeclaration(function, data)
-            data[function.simpleName.asString()] = resolver.getJvmName(function)
+            data[function.simpleName.asString()] = resolver.getJvmName(function)!!
```
yigit, line 63: might be better to just change the map to accept null values to match the method signature.

yigit (approved):
lgtm, one small nit but feel free to ignore.

also, i think this might break Room if it ever happens (idk how it can happen though). 

nevertheless, it is a good change so i'll update room once it is available in a build.

PR #352, opened 2021-03-12
Support throw list

neetopia:
* Add function to Resolver for getting checked exception from Java method

* Add a utility function for getting throw list from Kotlin function as declared in `@Throws` annotation

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
@@ -21,6 +21,7 @@ package com.google.devtools.ksp
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.visitor.KSValidateVisitor
+import kotlin.jvm.Throws
 
 /**
  * Try to resolve the [KSClassDeclaration] for a class using its fully qualified name.
--- a/compiler-plugin/testData/api/throwList.kt
+++ b/compiler-plugin/testData/api/throwList.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2021 Google LLC
+ * Copyright 2010-2021 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+// TEST PROCESSOR: ThrowListProcessor
+// EXPECTED:
+// java.io.IOException,java.util.NoSuchElementException
+// java.io.IOException,java.lang.IndexOutOfBoundsException
+// END
+// FILE: ThrowsException.java
+import java.io.IOException;
+import java.lang.IndexOutOfBoundsException;
+
+public class ThrowsException {
+    public int foo() throws IOException, IndexOutOfBoundsException{
+        return 1;
+    }
+}
+// FILE: a.kt
+class ThrowsKt {
+    @Throws(java.io.IOException::class, java.util.NoSuchElementException::class)
+    fun throwsKT()
+}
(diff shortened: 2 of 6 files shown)


api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -241,6 +242,21 @@ fun KSDeclaration.isVisibleFrom(other: KSDeclaration): Boolean {
     }
 }
 
+@Throws
```
yigit, line 245: what is this annotation for?

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -241,6 +242,21 @@ fun KSDeclaration.isVisibleFrom(other: KSDeclaration): Boolean {
...
+fun KSFunctionDeclaration.getThrowList(): Sequence<KSType> {
```
yigit, line 246: can we make this function handle both cases so that developer does not need to check the origin before calling?

compiler-plugin/testData/api/throwList.kt
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+
+// TEST PROCESSOR: ThrowListProcessor
+// EXPECTED:
+// java.io.IOException
+// java.io.IOException,java.lang.IndexOutOfBoundsException
+// END
+// FILE: ThrowsException.java
```
yigit, line 24: can we add a test with library so that we ensure the descriptor version works as well?

yigit (changes requested):
this is awesome as it will help me close a room bug easily :)

https://issuetracker.google.com/issues/171572318

compiler-plugin/testData/api/throwList.kt
neetopia: descriptor version won't work unless we read bytecode directly. Throw list does not have a place in kotlin compiler descriptors.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
neetopia: Oops, must be a mistake.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
neetopia: I can do that, but such function will have to be either a member function of `Resolver` or a utility function that takes a resolver as parameter, as resolver is needed to check throw list on Java origin.

ting-yuan: +1 on unifying them. Otherwise it would be weird to get an empty sequence when calling this on a Java function.

Putting it in `Resolver` sounds better to me, since this is more likely to be used when processing Java or Kotlin sources that need to be Java compatible. It's easier to be organized together with other Java specific functions such as `getJvmName`.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -241,6 +242,21 @@ fun KSDeclaration.isVisibleFrom(other: KSDeclaration): Boolean {
...
+    return if(this.origin == Origin.KOTLIN) {
+        this.annotations
+                .singleOrNull { it.annotationType.resolve().declaration.qualifiedName?.asString() == "kotlin.Throws" }
```
ting-yuan, line 249: Can we filter by the `annotationType`'s simple name before resolution?

compiler-plugin/testData/api/throwList.kt
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.IndexOutOfBoundsException;
+
+public class ThrowsException {
+    public int foo() throws IOException, IndexOutOfBoundsException{
+        return 1;
+    }
+}
+// FILE: a.kt
+class ThrowsKt {
+    @Throws(java.io.IOException::class)
```
ting-yuan, line 35: Can we throw one more exception to make sure multiple arguments won't be broken?

compiler-plugin/testData/api/throwList.kt
ting-yuan: Sounds like a good comment in the API :)

yigit: ugh that hurts :/.
we should document that behavior as it might matter a lot when generating java code that overrides descriptors coming from compiled code :/. Though I'm not even sure how developer can solve that kind of issue.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -21,6 +21,7 @@ package com.google.devtools.ksp
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.visitor.KSValidateVisitor
+import kotlin.jvm.Throws
```
ting-yuan, line 24: seems unused

ting-yuan (approved):
LGTM with a minor comment. Please feel free to merge after jcenter is back and CI pass.

PR #353, opened 2021-03-14
Integrate ktlint

yigit:
It is a fork of what androidX does with minor modifications.

There are 3 new tasks:

rootProject:ktlintApplyToIdea -> Applied ktlint configuration to the
current .idea project folder

<subproject>:ktlint -> Runs ktlint checks
<subproject>:ktlintFormat -> Fixes formatting for the project. Note that
this may not be able to handle all violations.

I've not modified the CI files to run this since it fails right now. One
of the core maintaners should run the format tasks to come to a clean
slate then we can add it to CI. I've not added a baseline file since
they are very brittle.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[*.{kt,kts}]
+indent_size=4
+insert_final_newline=true
+max_line_length=120
+disabled_rules=no-wildcard-imports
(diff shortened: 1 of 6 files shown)


yigit:
for reference, here is a sample format run:
https://github.com/yigit/ksp/commit/dea7ee3726f2634fd8ae1a81a0def0c3ab9389b0

this seems to fix most issues except for long lines.

ting-yuan (approved):
Thanks a lot!

PR #357, opened 2021-03-16
support annotation use site target of @setparam

neetopia:
Actually logic is becoming more complicated to the degree I am thinking of delegate to descriptors directly, but posting this solution for avoiding resolution anyway. I'm open to switch to descriptor based solution on this.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/synthetic/KSValueParameterSyntheticImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/synthetic/KSValueParameterSyntheticImpl.kt
@@ -1,6 +1,7 @@
 package com.google.devtools.ksp.symbol.impl.synthetic
 
 import com.google.devtools.ksp.ExceptionMessage
+import com.google.devtools.ksp.processing.impl.findAnnotationFromUseSiteTarget
 import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.symbol.impl.KSObjectCache
 import com.google.devtools.ksp.symbol.impl.binary.KSAnnotationDescriptorImpl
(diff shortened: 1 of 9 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/synthetic/KSValueParameterSyntheticImpl.kt
```diff
@@ -10,10 +10,10 @@ import org.jetbrains.kotlin.descriptors.ValueParameterDescriptor
 import org.jetbrains.kotlin.resolve.calls.components.hasDefaultValue
 import org.jetbrains.kotlin.resolve.calls.components.isVararg
 
-class KSValueParameterSyntheticImpl(resolve: () -> ValueParameterDescriptor?) : KSValueParameter {
+class KSValueParameterSyntheticImpl(val owner: KSAnnotated?, resolve: () -> ValueParameterDescriptor?) : KSValueParameter {
 
-    companion object : KSObjectCache<() -> ValueParameterDescriptor?, KSValueParameterSyntheticImpl>() {
-        fun getCached(resolve: () -> ValueParameterDescriptor?) = KSValueParameterSyntheticImpl.cache.getOrPut(resolve) { KSValueParameterSyntheticImpl(resolve) }
+    companion object : KSObjectCache<Pair<KSAnnotated?, () -> ValueParameterDescriptor?>, KSValueParameterSyntheticImpl>() {
+        fun getCached(owner: KSAnnotated? = null, resolve: () -> ValueParameterDescriptor?) = KSValueParameterSyntheticImpl.cache.getOrPut(Pair(owner, resolve)) { KSValueParameterSyntheticImpl(owner, resolve) }
```
ting-yuan, line 16: I cannot find a call site that supply this. Seems like a bug?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -170,6 +170,26 @@ fun FunctionDescriptor.toFunctionKSModifiers(): Set<Modifier> {
     return modifiers
 }
 
+fun PsiElement.findParentAnnotated(): KSAnnotated? {
```
ting-yuan, line 173: Would this make sense?
```
inline fun <reified T, K> PsiElement.findParent(ctor: (psi: T) -> K?): K? {
    var parent = this.parent
    while (parent != null && parent !is T)
        parent = parent.parent
    return (parent as? T)?.let(ctor)
}
```
and change the call site to
```
ktParameter.findParent { psi: KtAnnotated -> // Can be the least common super type of types of interest
    return while (psi) {
        is KtPropertyAccessor -> blabla
        else -> null
    }
}
```

ting-yuan (approved):

PR #359, opened 2021-03-18
fix type arguments for inner classes

neetopia:
Basically just use descriptors to get type arguments from parent class(es).

Also found a bug in `KSTypeArgumentJavaImpl` for `<?>` where the original implementation does not align with compiler behavior. Although with this change, source based `KSTypeArgument` implementations looks to be obsolete as we use descriptors for everything now, probably better to remove all these implementations.


ting-yuan (approved):

neetopia:
Added a fix for #360

PR #370, opened 2021-03-31
fix type reference construction logic for type parameter

neetopia:


compiler-plugin/testData/api/typeParameterReference.kt
```diff
@@ -18,17 +18,30 @@
...
 // Foo.T1: true
 // Foo.bar.T2: false
 // foo.T3: false
+// T
 // END
 
+// MODULE: lib
+// FILE: lib.kt
+interface LibFoo<T> {
+    val v: T
```
ting-yuan, line 32: This may not cover some cases; Can you add the following?
```val w: List<T>```

Also, can you think of a test case where `kotlinType` is a `FlexibleType`? Because the old code tried to handle it.

I wonder if the implementation details (e.g., `FlexibleType`) could be leaked via `toString()`.

ting-yuan (approved):

PR #377, opened 2021-04-06
Add Resolver.getDeclarationsFromPackage() API for getting declarations with a given package name.

neetopia:
TODO: Isolate symbol look up from module descriptor for symbols from source if possible. Current implementation is resolving through descriptors even for symbols from source therefore can cause unnecessary overhead due to symbol resolution.


ting-yuan (approved):
Please add `Resolver.getDeclarationsFromPackage` to `incremental.md` and feel free to merge after addressing the inline comments (if that make sense).

PR #386, opened 2021-04-11
Register generated sources in incremental caches

### LINKED ISSUE #385: Generated sources not considered when calculating dirty set
Example 1:
```
// Bar.kt
class Bar : Foo()

// Baz.kt
open class Baz

// Foo.kt is generated
open class Foo : Baz()
```

Example 2:
```
// Bar.kt
class Bar

// Baz.kt
class Baz

// Foo.kt is generated
class Foo

// FooBar.kt is generated from Foo and Bar
class FooBar

// FooBaz.kt is generated from Foo and Baz
class FooBaz
```

In both examples, touching `Baz` should have invalidated `Bar`.

ting-yuan:
Fixes #385

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
@@ -192,7 +192,7 @@ class IncrementalContext(
         options.cachesDir.deleteRecursively()
     }
 
-    private fun collectDefinedSymbols(ksFiles: List<KSFile>) {
+    private fun collectDefinedSymbols(ksFiles: Collection<KSFile>) {
         ksFiles.forEach { file ->
             file.accept(symbolCollector) {
                 updatedSymbols.putValue(file.relativeFile, it)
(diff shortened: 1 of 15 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -446,8 +454,11 @@ class IncrementalContext(
 
     // TODO: Don't do anything if processing failed.
     private fun updateCaches(dirtyFiles: Collection<File>, outputs: Set<File>, sourceToOutputs: Map<File, Set<File>>) {
-        updateSourceToOutputs(dirtyFiles, outputs, sourceToOutputs)
-        updateLookupCache(dirtyFiles)
+        // dirtyFiles may contain new files, which are unknown to sourceToOutputsMap.
+        val oldOutputs = dirtyFiles.flatMap { sourceToOutputsMap[it] ?: emptyList() }
```
neetopia, line 458: Is a `distinct()` needed in case an output file correspond to multiple sources?

neetopia (commented):
Can you elaborate on what's the role of `removedOutputs` in registering generated sources?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
ting-yuan: Done. Good catch!

neetopia (approved):

PR #387, opened 2021-04-14
Add getFunctionDeclarationByName, getPropertyDeclarationByName 

neetopia:


api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
```diff
@@ -56,6 +56,24 @@ interface Resolver {
...
     fun getClassDeclarationByName(name: KSName): KSClassDeclaration?
 
+    /**
+     * Find a property in the compilation classpath for the given name.
+     *
+     * @param name fully qualified name of the function to be loaded; using '.' as separator.
+     * @param includeTopLevel a boolean value indicate if top level functions should be searched. Default false. Note if top level functions are included, this operation can be expensive.
+     * @return a KSFunctionDeclaration, or null if not found.
+     */
+    fun getFunctionDeclarationByName(name: KSName, includeTopLevel: Boolean = false): KSFunctionDeclaration?
```
ting-yuan, line 66: There can be several functions with exactly the same name. I.e., overloading.

neetopia: This is handled by the implementation in `getAllFunctions()`

ting-yuan: When there are two functions with the same name, this will return null. Shouldn't it return a list?

neetopia: Oh you are right, I was thinking about overriding.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -177,6 +177,28 @@ class ResolverImpl(
             }
     }
 
+    override fun getFunctionDeclarationsByName(name: KSName, includeTopLevel: Boolean): Sequence<KSFunctionDeclaration> {
+        val qualifier = name.getQualifier()
+        val functionName = name.getShortName()
+        val nonTopLevelResult = this.getClassDeclarationByName(qualifier)?.getAllFunctions()?.filter{ it.simpleName.asString() == functionName }?.asSequence() ?: emptySequence()
```
ting-yuan, line 183: May I know why `getAllFucntions()` is used? Since this API is used to get *declarations*, shouldn't we include only *declared* functions?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -177,6 +177,28 @@ class ResolverImpl(
...
+            nonTopLevelResult.plus(module.getPackage(FqName(qualifier))
+                .memberScope.getContributedDescriptors(DescriptorKindFilter.FUNCTIONS) { it.asString() == functionName }
+                .filterIsInstance<MemberDescriptor>().mapNotNull {  it.toKSDeclaration() as? KSFunctionDeclaration} )
+        }
+    }
+
+    override fun getPropertyDeclarationByName(name: KSName, includeTopLevel: Boolean): KSPropertyDeclaration? {
+        val qualifier = name.getQualifier()
+        val propertyName = name.getShortName()
+        val nonTopLevelResult = this.getClassDeclarationByName(qualifier)?.getAllProperties()?.singleOrNull { it.simpleName.asString() == propertyName }
```
ting-yuan, line 194: Same for `getAllProperties()`; Inherited properties might not be needed / desired.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -177,6 +177,28 @@ class ResolverImpl(
...
+        val functionName = name.getShortName()
+        val nonTopLevelResult = this.getClassDeclarationByName(qualifier)?.getAllFunctions()?.filter{ it.simpleName.asString() == functionName }?.asSequence() ?: emptySequence()
+        return if (!includeTopLevel) nonTopLevelResult else {
+            nonTopLevelResult.plus(module.getPackage(FqName(qualifier))
+                .memberScope.getContributedDescriptors(DescriptorKindFilter.FUNCTIONS) { it.asString() == functionName }
+                .filterIsInstance<MemberDescriptor>().mapNotNull {  it.toKSDeclaration() as? KSFunctionDeclaration} )
+        }
+    }
+
+    override fun getPropertyDeclarationByName(name: KSName, includeTopLevel: Boolean): KSPropertyDeclaration? {
```
ting-yuan, line 191: This may need to return a sequence as well, because:
```
package com.example.foo
fun bar() = ...
```
```
package com.example
class foo {
  fun bar() = ...
}
```

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
neetopia: it is used to get declarations, but I think declarations does not necessary mean comes from source? Like `getClassDeclaration` can get you declarations from classpath.

ting-yuan (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
ting-yuan: Well, this isn't allowed at compile time and therefore won't be a problem in KSP.

PR #390, opened 2021-04-16
Add generated files to symbolsMap in subsequent builds

ting-yuan:
and use relative paths.

### DIFF
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/GeneratedRefsIncIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/GeneratedRefsIncIT.kt
@@ -21,14 +21,49 @@ class GeneratedRefsIncIT {
             "w: [ksp] 3: [File: Goo.kt]"
         )
 
+        val expectedBar = listOf(
+            "w: [ksp] 1: [File: Bar.kt]",
+            "w: [ksp] 2: [File: Foo.kt]",
+            "w: [ksp] 3: [File: Goo.kt]"
+        )
+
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/GeneratedRefsIncIT.kt
```diff
@@ -21,14 +21,49 @@ class GeneratedRefsIncIT {
...
+
+        // Baz doesn't depend on Bar, so touching Bar won't invalidate Baz.
+        File(project.root, "workload/src/main/kotlin/com/example/Bar.kt").appendText("\n\n")
+        gradleRunner.withArguments("assemble").build().let { result ->
+            val outputs = result.output.split("\n").filter { it.startsWith("w: [ksp]")}
+            Assert.assertEquals(expectedBar, outputs)
+        }
+
+        // Make Baz depends on Bar; Bar will be invalidated.
+        File(project.root, "workload/src/main/kotlin/com/example/Bar.kt").appendText("\nclass List<T>\n")
```
neetopia, line 49: why is adding a `class List<T>` makes `Baz` to depend on `Bar`?

ting-yuan: Because `Baz.kt` has has a reference to `List`:
```
val l: List<Int> = TODO()
```
The definition here `com.example.List` is in the same package with the reference and has a higher priority than `kotlin.collections.List` in resolution.

neetopia (approved):

PR #394, opened 2021-04-21
Add processor into ksp tasks' inputs

ting-yuan:
Previously, it's done by adding to ksp tasks' source, which incorrectly
made Gradle think it is part of the source as a side-effect.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -23,8 +23,10 @@ import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.UnknownTaskException
 import org.gradle.api.artifacts.Configuration
+import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.TaskProvider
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -223,7 +223,7 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             nonEmptyKspConfigurations.forEach {
```
gavra0, line 223: You should create a new configuration that extends from all non-empty ones. Something like
```
val processorClasspath = project.configurations.create(<unique_name>).extendsFrom(nonEmptyConfigurations)
kspTask.processorClasspath.from(processorClasspath)

class KspTask {
   @InputFiles
   @SkipWhenEmpty 
   abstract val processorClasspath: ConfigurableFileCollection // Gradle will instantiate this 
```

ting-yuan: Done. I removed `@SkipWhenEmtpy` because it seems to make Gradle treat the processorClasspath as source.

For example, when there is no source under `test/`, I think the desired behavior is always skipping the `kspTestKotlin` task, no matter whether there are processor classpath changes or not.

Also the annotations (`InputFiles` and `SkipWhenEmpty`) cannot be applied to `abstract val ...`, so I used `abstract fun getProcessorClasspath()` rather than `abstract val processorClasspath`

gavra0: SGTM. You can use `@get:InputFIles abstract val ...` to avoid creating the getter.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -220,10 +222,14 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kspTask.destination = destinationDir
             kspTask.outputs.dirs(kotlinOutputDir, javaOutputDir, classOutputDir, resourceOutputDir)
             kspTask.blockOtherCompilerPlugins = kspExtension.blockOtherCompilerPlugins
+
+            // depends on the processor; if the processor changes, it needs to be reprocessed.
+            val processorClasspath = project.configurations.maybeCreate("ProcessorClasspath")
```
gavra0, line 227: For Android projects, there will be multiple KSP tasks, so this name will not be unique. Consider using "${taskName}ProcessorsClasspath".

gavra0 (approved):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
ting-yuan: So true! Updated.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
ting-yuan: Done. Thanks!

PR #397, opened 2021-04-21
Update dependencies

Goooler:


neetopia:
Thanks for your PR, however for dependency version bump, there is additional work needs to be done on our side(as you can see the CI is already failing), therefore we want to keep such work to ourselves.

PR #406, opened 2021-04-28
Add FUN and VALUE to Modifier.kt

### LINKED ISSUE #393: Add modifier for "fun interface"
Support the language feature "function SAM interfaces" - https://kotlinlang.org/docs/fun-interfaces.html

As far as I can tell `com.google.devtools.ksp.symbol.Modifier` doesn't have an entry for this and there is currently no way to check this on an KSClassDeclaration.

It seems like this would be fairly straightforward to do?
1. Add fun entry to `Modifier` enum
2. Update KSClassDeclarationDescriptorImpl.modifiers to check `ClassDescriptor.isFun`
3. Update utils.kt `modifierMap`

ZacSweers:
This introduces support for mapping both `fun` (i.e. for `fun interface`) and `value` (i.e. `value class`) modifiers to KSP modifiers

Resolves #393 along the way. I couldn't find any existing test that covers modifiers reading, let me know if theres a good place to add it.


neetopia (approved):
change looks good to me, as for test, if there is no existing test, can you add a test covering your changes? I can add test for other modifiers later based on your test.

neetopia:
Thanks for PR! I will add tests later.

ZacSweers:
Thanks! Sorry for the delay, been a busy past couple weeks for me

PR #413, opened 2021-05-03
Introduce asMemberOf to KSPropertyDeclaration and KSFunctionDeclaration

### LINKED ISSUE #208: Make `asMemberOf` an extension function
Right now it is a member of `Resolver`:
```
fun asMemberOf(
    property: KSPropertyDeclaration,
    containing: KSType
): KSType
```
Should be more *Kotlin* if:
```
fun KSPropertyDeclaration.asMemberOf(
    containing: KSType,
): KSType
```

ting-yuan:
They are more convenient and object oriented.

fixes #208

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
@@ -116,64 +116,6 @@ interface Resolver {
     fun overrides(overrider: KSDeclaration, overridee: KSDeclaration): Boolean
 
     /**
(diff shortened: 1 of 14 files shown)


neetopia:
should the original extension function be removed then?

neetopia (approved):

ting-yuan:
> should the original extension function be removed then?

which extension function?

neetopia:
I mean `Resolver.asMemberOf()`, could make it internal since it is same as the member functions from this PR.

ting-yuan:
Make sense. Please take a look again.

neetopia (approved):

PR #414, opened 2021-05-04
Introduce mapJavaToKotlin and mapKotlinToJava to Resolver

### LINKED ISSUE #126: Should `getClassDeclarationByName` mimic Kotlin Compiler?
Right now, if you request a java type explicitly from `getClassDeclarationByName`, it returns a KSType that specifically represnts the java type instead of its Kotlin counterpart.

e.g.
```
val kotlinType = resolver.getClassDeclarationByName("kotlin.String")
val javaType = resolver.getClassDeclarationByName("java.lang.String")
kotlinType == javaType // false
kotlinType.isAssignableFrom(javaType) // false
javaType.isAssingnableFrom(kotlinType) // false
```
On the other hand, if `java.lang.String` is every used in kotlin code or java code; it is always resolved to `kotlin.String`

e.g.

```
val x : String // type is kotlin.String!
val y : java.lang.String // type is kotlin.String!

// java
static String foo; // type is kotlin.String..kotlinString?
```

I'm wondering if it would be better (more consistent) if KSP behaved like the kotlin compiler where `getClassDeclarationByName` would return the kotlin type.
Otherwise, for an annotation processor that wants to support both java and kotlin, certain common functionalities becomes harder to write. e..g if it wants to check if a type is a `String`, it must check for `kotlin.String` AND `java.lang.String`. Even though, for individual cases, it seems feasible, it adds a burden on annotation processors to be aware of all types that have kotlin counterparts.

Here are some possible options:

a) Provide a mapper in resolver that receives the java name and returns the kotlin name (maybe even both ways). 
* this would be enough for room where we would always call it before getting the type to ensure Room is always in 1 realm (e.g. uses Kotlin types).

b) Change behavior of `getClassDeclarationByName` to always return the kotlin type, add another function to get java declaration.
* Even though this makes it more consistent with the kotlin compiler, it has the downside of the inconsistency where the qualified name of the returned class may not match the qualified name of the argument

c) Add an additional parameter to `getClassDeclarationByName` to specify the desired behavior.
* Makes the API more explicit but also more verbose.

d) Change all type methods to handle these cases. e.g. `isAssignableFrom` will return `true` for the case above (#125)
* I'm not actually sure if this is even feasible as it would also require equality to possibly return true or some other cases I cannot think of right now, hence could easily become inconsistent.

Creating this issue to discuss a path forward.

ting-yuan:
They map Java classes to Kotlin classes and vice versa. For example,
    java.lang.String <-> kotlin.String
    java.lang.Integer <-> kotlin.Int

See https://kotlinlang.org/docs/java-interop.html#mapped-types for
details.

Also introduced two extension functions that loads Java/Kotlin as Kotlin/Java classes when applicable.

Fixes #126

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
@@ -51,6 +51,15 @@ interface Resolver {
     /**
      * Find a class in the compilation classpath for the given name.
      *
+     * This returns the exact platform class when given a platform name. Note that java.lang.String isn't compatible
+     * with kotlin.String in the type system. Therefore, processors need to use mapJavaNameToKotlin() and mapKotlinNameToJava()
+     * explicitly to find the corresponding class names before calling getClassDeclarationByName if type checking
+     * is needed for the classes loaded by this.
+     *
+     * This behavior is limited to getClassDeclarationByName; When processors get a class or type from a Java source
+     * file, the conversion is done automatically. E.g., a java.lang.String in a Java source file is loaded as
+     * kotlin.String in KSP.
+     *
      * @param name fully qualified name of the class to be loaded; using '.' as separator.
      * @return a KSClassDeclaration, or null if not found.
      */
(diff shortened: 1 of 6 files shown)


api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
```diff
@@ -212,4 +221,35 @@ interface Resolver {
...
+     * java.lang.Integer -> kotlin.Int
+     * java.util.List -> kotlin.List
+     * java.util.Map.Entry -> kotlin.Map.Entry
+     * java.lang.Void -> null
+     *
+     * @param javaName a Java class name
+     * @return corresponding Kotlin class name or null
+     */
+    @KspExperimental
+    fun mapJavaToKotlin(javaName: KSName): KSName?
```
neetopia, line 239: maybe change name to `mapJavaNameToKotlin`? Same for the other one.

neetopia (approved):

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
ting-yuan: Done.

PR #417, opened 2021-05-07
Split Origin.CLASS into KOTLIN_LIB and JAVA_LIB

ting-yuan:
For KSAnnotationDescriptor, check whether the underlying descripter is
JavaAnnotationDescriptor.

For KSDeclarationDescriptorImpl, KSPropertyAccessorDescriptorImpl and
KSValueParameterDescriptorImpl, check whether they are contained by
JavaClassDescriptor.

For KSTypeReferenceDescriptorImpl, KSClassifierReferenceDescriptorImpl
and KSTypeArgumentDescriptorImpl, specify the origin in their creation
appropriately. Most of time it is the same as where they are created.
The only exception is when it's created for type arguemnts from KotlinType.
In that case it is SYNTHETIC.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
@@ -59,8 +59,9 @@ class KSTypeImpl private constructor(
         }
     }
 
+    // TODO: fix calls to getKSTypeCached and use ksTypeArguments when available.
     override val arguments: List<KSTypeArgument> by lazy {
-        kotlinType.arguments.map { KSTypeArgumentDescriptorImpl.getCached(it) }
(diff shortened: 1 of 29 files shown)


ting-yuan:
I'm still adding tests but would like to share the idea as early as possible. Please let me know if you spot anything wrong.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
```diff
@@ -60,7 +60,7 @@ class KSTypeImpl private constructor(
     }
 
     override val arguments: List<KSTypeArgument> by lazy {
-        kotlinType.arguments.map { KSTypeArgumentDescriptorImpl.getCached(it) }
+        kotlinType.arguments.map { KSTypeArgumentDescriptorImpl.getCached(it, Origin.SYNTHETIC) }
```
neetopia, line 63: I'm not sure if SYNTHETIC is the best way to describe origin here. IMO there is no answer for origin in KSType.

neetopia (approved):

PR #420, opened 2021-05-08
Support KSDeclaration.getDocString

ting-yuan:
A doc string is a string in the comment right in front of an declaration
and is enclosed by /** and */. Leading whitespaces and * are removed for
each line.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
@@ -50,6 +50,7 @@ import org.jetbrains.kotlin.types.StarProjectionImpl
 import org.jetbrains.kotlin.types.TypeProjectionImpl
 import org.jetbrains.kotlin.types.replace
 import org.jetbrains.kotlin.load.java.lazy.ModuleClassResolver
+import org.jetbrains.kotlin.psi.psiUtil.siblings
 import org.jetbrains.kotlin.resolve.annotations.hasJvmStaticAnnotation
 import org.jetbrains.kotlin.resolve.scopes.DescriptorKindFilter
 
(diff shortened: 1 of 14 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -226,6 +227,21 @@ fun PsiElement.toLocation(): Location {
...
+    if (!t1.startsWith("/**") || !t1.endsWith("*/"))
+        return null
+    val lineSep = t1.findAnyOf(listOf("\r\n", "\n", "\r"))?.second ?: ""
+    return t1.trim('/').trim('*').lines().joinToString(lineSep) {
+        it.trimStart().trimStart('*')
+    }
+}
+
+fun PsiElement.getDocString(): String? =
+    this.firstChild.siblings().firstOrNull { it is PsiComment }?.let {
```
neetopia, line 241: Should we take care of the case when normal comments are used with document comments?

ting-yuan: What do you mean by normal comments? `//`, `/* */` or ...?

neetopia: both

ting-yuan: No, we should keep it simple and not include random comments.

neetopia: Consider this code:
```
// comment 1
/**
* document 
*/ 
class Foo
```
what will happen with your implementation?

ting-yuan: comment 1 will be ignored.

PR #421, opened 2021-05-08
Bugfix/issue 367

Virelion:


natario1:
Thanks for spotting this! I think it would be safer / more idiomatic if instead of copying the configurations, you move all the options construction inside the `tasks.register { }` block for the KSP task. This should make the resolution lazy. I'm not a maintainer though.

Virelion:
Good catch! Deffering configuration resolution seems like cleaner fix.
Additionaly, I allowed myself to do a bit of refactoring to lower the amount of lines in applyToCompilation.

ting-yuan:
Thanks for the fix! Could you remove those `mavenLocal` before we merge it?

Virelion:
Sure thing, I removed mavenLocal form this PR.

ting-yuan (approved):
Thanks!

PR #422, opened 2021-05-10
Extract create args to SymbolProcessorEnvironment

evant:
This allows future arguments to be added without forcing changes to downstream
processors. Naming is modeled after apt's ProcessingEnvironment.

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorProvider.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorProvider.kt
@@ -3,19 +3,9 @@ package com.google.devtools.ksp.processing
 /**
  * [SymbolProcessorProvider] is the interface used by plugins to integrate into Kotlin Symbol Processing.
  */
-fun interface SymbolProcessorProvider {
+interface SymbolProcessorProvider {
     /**
-    * Called by Kotlin Symbol Processing to create the processor.
-    *
-    * @param options passed from command line, Gradle, etc.
-    * @param kotlinVersion language version of compilation environment.
-    * @param codeGenerator creates managed files.
-    * @param logger for logging to build output.
-    */
-    fun create(
-        options: Map<String, String>,
-        kotlinVersion: KotlinVersion,
-        codeGenerator: CodeGenerator,
-        logger: KSPLogger
-    ): SymbolProcessor
+     * Called by Kotlin Symbol Processing to create the processor.
+     */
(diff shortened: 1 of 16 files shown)


ZacSweers (commented):
I like the idea, what would you think of just a shorter "KSPEnvironment" or "KSEnvironment"?

evant:
I was follow `SymbolProcessor` and `SymbolProcessorProvider` but don't have strong opinions here.

api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorProvider.kt
```diff
@@ -3,19 +3,30 @@ package com.google.devtools.ksp.processing
...
-    */
+     * Called by Kotlin Symbol Processing to create the processor.
+     *
+     * @param options passed from command line, Gradle, etc.
+     * @param kotlinVersion language version of compilation environment.
+     * @param codeGenerator creates managed files.
+     * @param logger for logging to build output.
+     */
+    @Suppress("DeprecatedCallableAddReplaceWith")
+    @Deprecated(message = "Use create that takes an environment instead")
```
ting-yuan, line 16: Please feel free to remove this and keep the fun interface; Backward compatibility don't need to be considered yet.

edrd-f:
> I like the idea, what would you think of just a shorter "KSPEnvironment" or "KSEnvironment"?

I like the way it is now because the `KS*` prefix is already used for Kotlin symbols (e.g. `KSClassDeclaration`), so this provides a clear separation of KSP APIs.

ZacSweers:
It's not just for symbols currently, see KSPLogger

api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorProvider.kt
evant: Was wondering about that. Does this need to be a `fun interface`? It isn't used anywhere as such.

edrd-f:
Indeed. Well, another option is to rename `KSPLogger` to `SymbolProcessorLogger`.

api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorProvider.kt
ting-yuan: It's fine to remove the fun interface. Can you remove the deprecated `create` and do a rebase?

ting-yuan (approved):

ting-yuan:
Thanks for the PR!

PR #425, opened 2021-05-12
Remove kotlinOptions.jvmTarget specification

mars885:
Kotlin 1.5 now targets JVM 1.8 [by defaut](https://kotlinlang.org/docs/whatsnew15.html#new-default-jvm-target-1-8), so there's no need to specify it again explicitly.


ting-yuan (approved):
Thanks!

PR #436, opened 2021-05-19
support value parameter for getSymbolsWithAnnotation

### LINKED ISSUE #351: `value-parameter` annotation are not retrieved
Given:
```kotlin
annotation class Builder

class MyClass(@Builder param: String)
```

A call to `getSymbolsWithAnnotation`, even with `inDept = true` is not picking it up.

If we declare it as a `val` instead, then it works:
```kotlin
class MyClass(@Builder val param: String)
```

neetopia:
fixes #351


ting-yuan (approved):
LGTM

btw, maybe this visitor should extend KSTopDownVisitor and only override visitAnnotated.

PR #441, opened 2021-05-21
Drop PSI cache in each round

ting-yuan:


integration-tests/src/test/resources/psi-cache/test-processor/src/main/kotlin/TestProcessor.kt
```diff
@@ -0,0 +1,44 @@
...
+) : SymbolProcessor {
+    var rounds = 0
+    override fun process(resolver: Resolver): List<KSAnnotated> {
+        rounds++
+
+        val syms = resolver.getSymbolsWithAnnotation("com.example.Anno").toList()
+
+        if (rounds == 2) {
+            syms.forEach {
+                if (!it.validate()) {
```
neetopia, line 20: If we want to reproduce the caching bug, you should be validating at the first round as well.

ting-yuan: Good catch!

neetopia (approved):

neetopia (approved):

Issue #418, opened 2021-05-07
getDeclarationsFromPackage misses elements from Java source files


ting-yuan:
Not sure whether this is an issue in the test framework, though. Things in the following test file in testData are not found:
```
// FILE: JavaSrc.java
package foo.bar;
...
```

neetopia:
should be fixed by #442

PR #442, opened 2021-05-21
fix an issue where getDeclarationsFromPackage not including java symbols

neetopia:
Previous implementation is using module descriptor which is specified as `Kotlin module` therefore not including Java symbols.


neetopia:
fixes #418

ting-yuan:
Do `getFunctionDeclarationsByName` and `getPropertyDeclarationByName` suffer from the same issue?

neetopia:
No, since Java does not have top level function/property.

ting-yuan (approved):

PR #446, opened 2021-05-21
fix getDeclarationsFromPackage to exclude symbols found in module descriptor

neetopia:
A follow up on #442 

Actually module descriptor is able to find symbols from Java source iff the said class is sitting in the file with name and path following Java convention, i.e. for a class `main.C`, it has to be in a file of path `/main/C.java`, therefore ignoring other classes in the same file, or any files not following path convention. With that, a distinct call is needed to filter redundant classes found.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
@@ -134,7 +134,7 @@ abstract class AbstractKotlinSymbolProcessingExtension(val options: KspOptions, 
         }
 
         // dirtyFiles cannot be reused because they are created in the old container.
-        val resolver = ResolverImpl(module, ksFiles.filterNot { it.filePath in cleanFilenames }, newFiles, deferredSymbols, bindingTrace, project, componentProvider, incrementalContext)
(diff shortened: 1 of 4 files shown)


ting-yuan (approved):

neetopia:
This update is still having one corner case where the path could be wrong while still matching package name, e.g. `/main/test/main/test/C.java` for package `main.test`. I can't find a good way to tell whether the first `main/test/` comes the path of the file or path of the project though, since the path we obtain from PSI is absolute path.

ting-yuan:
> I can't find a good way to tell whether the first `main/test/` comes the path of the file or path of the project though, since the path we obtain from PSI is absolute path.

Can you try `options.javaSourceRoots` and `options.javaOutputDir` in `AbstractKotlinSymbolProcessingExtension`?

neetopia:
> > I can't find a good way to tell whether the first `main/test/` comes the path of the file or path of the project though, since the path we obtain from PSI is absolute path.
> 
> Can you try `options.javaSourceRoots` and `options.javaOutputDir` in `AbstractKotlinSymbolProcessingExtension`?

I've considered that approach, but it is having the same issue, when you have multiple roots nesting in others, e.g. root1: `/FOLDER1/a/b/c`, root2: `/a/b/c`. Wasn't sure if this is valid though.

ting-yuan:
The path is valid if it can be found it any of the roots. Clashing is another issue that is also a problem in compilation.

PR #449, opened 2021-05-24
Add API to check if a property has a backing field

yigit:
Unfortunately, the compiler API does not cover deserialized descriptors fully so I've implemented a custom binary class visitor to check that (provided by  @ting-yuan )

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
@@ -35,12 +35,9 @@ import com.google.devtools.ksp.symbol.impl.synthetic.KSPropertyGetterSyntheticIm
 import com.google.devtools.ksp.symbol.impl.synthetic.KSPropertySetterSyntheticImpl
 import com.google.devtools.ksp.symbol.impl.synthetic.KSValueParameterSyntheticImpl
 import com.intellij.psi.impl.source.PsiClassImpl
-import org.jetbrains.kotlin.builtins.getFunctionalClassKind
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.load.java.JavaDescriptorVisibilities
 import org.jetbrains.kotlin.load.java.descriptors.JavaClassConstructorDescriptor
-import org.jetbrains.kotlin.load.java.isFromJava
-import org.jetbrains.kotlin.load.java.structure.JavaMember
 import org.jetbrains.kotlin.load.java.structure.impl.JavaConstructorImpl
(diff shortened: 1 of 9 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -422,3 +425,56 @@ internal fun KSAnnotated.getInstanceForCurrentRound(): KSAnnotated? {
...
+
+/**
+ * Custom check for backing fields of descriptors that support properties coming from .class files.
+ * The compiler API always returns true for them even when they don't have backing fields.
+ */
+internal fun PropertyDescriptor.hasBackingFieldWithBinaryClassSupport(): Boolean {
+    return this.hasBackingField(BindingContext.EMPTY) ||
+        (this is DeserializedPropertyDescriptor && this.hasBackingFieldInBinaryClass()) ||
+        this.declaresDefaultValue
+}
```
ting-yuan, line 437: Use of `hasBackingField` looks hacky: 
```
    kind == CallableMemberDescriptor.Kind.FAKE_OVERRIDE -> overriddenDescriptors.any { it.hasBackingField(bindingContext) }
    extensionReceiverParameter != null -> false
    source is KotlinSourceElement && bindingContext != null -> bindingContext.get(BindingContext.BACKING_FIELD_REQUIRED, this) ?: false
    compileTimeInitializer != null -> true
//    backingField != null -> true // TODO: in case of DeserializedPropetyDescriptor this function returns incorrect result for FO field
    getter != null -> false
    else -> true
```
The first case isn't applicable to KSP and is always skipped.
The third case always returns false for elements with Kotlin source code, because of `BindingContext.EMPTY`.
The getter check isn't right.

Can we expand it into something like this?
``` 
return when {
    extensionReceiverParameter != null -> false // common check 1
    compileTimeInitializer != null -> true // common check 2
    this is DeserializedPropertyDescriptor -> this.hasBackingFieldInBinaryClass() // kotlin class
    this.source is KotlinSourceElement -> this.declaresDefaultValue // kotlin source
    else -> true // Java source or class
}
```

yigit: sure, just to make it clear, you prefer I remove the call to the compiler provided has backing field and instead just rely on our own implementation? (since the compiler provided one is hacky/buggy anyways)

ting-yuan: Yes. (and you're right, it is more that `hasBackingField` is hacky/buggy rather than the way we use it)

ting-yuan (approved):

Issue #438, opened 2021-05-20
blockOtherCompilerPlugins is broken


ting-yuan:
Probably broken by some recent changes. The expected dependency now exists in configuration `:workload:kotlinCompilerPluginClasspath<source-set>` instead of `:workload:kotlinCompilerPluginClasspath`

PR #451, opened 2021-05-25
Update implementation of blockOtherCompilerPlugins

ting-yuan:
Previously, kotlin-gradle-plugin names compiler plugin configuration
after PLUGIN_CLASSPATH_CONFIGURATION_NAME, now it is postfixed by target
and source set names.


ting-yuan:
Fixes #438

neetopia (approved):

PR #454, opened 2021-05-26
Move test processors out of main source set

ting-yuan:
Also rewrote tests that use implementation classes except RecordJava*Processor, which rely on an internal dump
function.


compiler-plugin/testData/api/referenceElement.kt
```diff
@@ -22,16 +22,27 @@
 // KSClassifierReferenceImpl: Qualifier of C is A
 // KSClassifierReferenceImpl: Qualifier of Int is null
 // KSClassifierReferenceImpl: Qualifier of String is null
-// KSClassifierReferenceDescriptorImpl: Qualifier of B is A
-// KSClassifierReferenceDescriptorImpl: Qualifier of C<Int> is A<String>
```
neetopia, line 26: is the removal of these 2 lines expected?

ting-yuan: Yes. Previously, they were created from source file. Now they are moved to lib module and renamed.

neetopia (approved):

PR #456, opened 2021-05-26
fix findOverridee for kotlin/java properties.

### LINKED ISSUE #248: java.lang.IllegalStateException: unexpected psi: class org.jetbrains.kotlin.psi.KtProperty
this is very similar to #164 and started happening after the 2021-01-11 update.

For the given input:
```
class JavaImpl implements MyInterface {
                public int getX() {
                    return 1;
                }
                public int getY() {
                    return 1;
                }
                public void setY(int value) {
                }
            }
```

```
interface MyInterface {
                val x:Int
                var y:Int
            }
```
calling `findOverridee` on `setY` of `JavaImpl` throws the following exception:

```
Caused by: java.lang.IllegalStateException: unexpected psi: class org.jetbrains.kotlin.psi.KtProperty
	at com.google.devtools.ksp.symbol.impl.UtilsKt.toKSFunctionDeclaration(utils.kt:267)
	at com.google.devtools.ksp.symbol.impl.java.KSFunctionDeclarationJavaImpl.findOverridee(KSFunctionDeclarationJavaImpl.kt:52)
	at androidx.room.compiler.processing.ksp.ResolverExtKt.overrides(ResolverExt.kt:82)
	at androidx.room.compiler.processing.ksp.ResolverExtKt.overrides(ResolverExt.kt:73)
```

neetopia:
fixes #248

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSFunctionDeclarationDescriptorImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSFunctionDeclarationDescriptorImpl.kt
@@ -28,7 +28,7 @@ import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.symbol.impl.KSObjectCache
 import com.google.devtools.ksp.symbol.impl.findClosestOverridee
 import com.google.devtools.ksp.symbol.impl.toFunctionKSModifiers
-import com.google.devtools.ksp.symbol.impl.toKSFunctionDeclaration
+import com.google.devtools.ksp.symbol.impl.toKSDeclaration
 import com.google.devtools.ksp.symbol.impl.toKSModifiers
 import org.jetbrains.kotlin.descriptors.ClassConstructorDescriptor
 import org.jetbrains.kotlin.load.java.isFromJava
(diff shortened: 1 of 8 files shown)


api/src/main/kotlin/com/google/devtools/ksp/symbol/KSPropertyDeclaration.kt
```diff
@@ -81,7 +81,7 @@ interface KSPropertyDeclaration : KSDeclaration {
      * to the current containing declaration is selected. If they are in the same level, the
      * property of the first specified interface (in source) will be returned.
      *
-     * @return [KSPropertyDeclaration] for the overridden property, if overriding, otherwise null.
+     * @return [KSPropertyDeclaration] for the overridden declaration, if overriding, otherwise null.
```
ting-yuan, line 84: Seems unnecessary; A property can only override another property.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSFunctionDeclarationDescriptorImpl.kt
```diff
@@ -42,9 +42,9 @@ class KSFunctionDeclarationDescriptorImpl private constructor(val descriptor: Fu
         fun getCached(descriptor: FunctionDescriptor) = cache.getOrPut(descriptor) { KSFunctionDeclarationDescriptorImpl(descriptor) }
     }
 
-    override fun findOverridee(): KSFunctionDeclaration? {
+    override fun findOverridee(): KSDeclaration? {
         val descriptor = ResolverImpl.instance.resolveFunctionDeclaration(this)
```
ting-yuan, line 46: nit: not introduced by this commit, but this line is unnecessary

ting-yuan (approved):

PR #465, opened 2021-06-01
add a workaround for unsupported type substitution for property accessors

### LINKED ISSUE #462: Unsupported operation exception when java method overrides a property
When a java method overrides a kotlin property declaration, calling `javaMethod.asMemberOf(javaType)` throws unsupported operation exception.

e.g.

```
interface MyKotlinInterface {
    val x:Int
    var y:Int
}
```

```
class JavaImpl implements MyKotlinInterface { // remove implements and it will work fine
    public int getX() {
        return 1;
    }
    public int getY() {
        return 1;
    }
    public void setY(int value) {
    }
}
```

This test throws the exception:

```
resolver.getClassDeclarationByName("JavaImpl").let { javaImplDecl ->
            checkNotNull(javaImplDecl)
            val getX = javaImplDecl.getDeclaredFunctions().first { it.simpleName.asString() == "getX" }
            results.add(getX.asMemberOf(javaImplDecl.asStarProjectedType()).toSignature())
        }
```
Full repro: https://github.com/yigit/ksp/commit/f703801610b962b2f03fe4ee95fe870c69049d55

This might be related to #248  but not sure. Hit this while updating Room to beta01.

stacktrace:

```
error: [ksp] java.lang.UnsupportedOperationException
	at org.jetbrains.kotlin.descriptors.impl.PropertyAccessorDescriptorImpl.substitute(PropertyAccessorDescriptorImpl.java:122)
	at org.jetbrains.kotlin.descriptors.impl.PropertyAccessorDescriptorImpl.substitute(PropertyAccessorDescriptorImpl.java:31)
	at org.jetbrains.kotlin.resolve.calls.inference.DescriptorRelatedInferenceUtilsKt.substitute(DescriptorRelatedInferenceUtils.kt:42)
	at com.google.devtools.ksp.processing.impl.ResolverImpl.computeAsMemberOf(ResolverImpl.kt:757)
	at com.google.devtools.ksp.processing.impl.ResolverImpl.asMemberOf$compiler_plugin(ResolverImpl.kt:735)
	at com.google.devtools.ksp.symbol.impl.java.KSFunctionDeclarationJavaImpl.asMemberOf(KSFunctionDeclarationJavaImpl.kt:121)
```

neetopia:
fixes #462 
Current fix relies on the underlying naming convention for `JavaPropertyAccessor` from Kotlin compiler, which might change and break logic. Ultimate change should happen in upstream to fill out the `TODO` for `PropertyAccessorDescriptorImpl`

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -754,6 +754,11 @@ class ResolverImpl(
                 )
             }
             val typeSubstitutor = containing.kotlinType.createTypeSubstitutor()
+            if (declaration is PropertyAccessorDescriptor) {
+                val substitutedProperty = (declaration.correspondingProperty).substitute(typeSubstitutor)
+                // TODO: Fix in upstream for property accessors: https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/descriptors/impl/PropertyAccessorDescriptorImpl.java#L122
+                return KSFunctionImpl((substitutedProperty as PropertyDescriptor).accessors.single { it.name == declaration.name })
+            }
             val substituted = declaration.substitute(typeSubstitutor)
             return KSFunctionImpl(substituted)
         }
--- a/compiler-plugin/src/test/kotlin/com/google/devtools/ksp/processor/AsMemberOfProcessor.kt
+++ b/compiler-plugin/src/test/kotlin/com/google/devtools/ksp/processor/AsMemberOfProcessor.kt
@@ -85,6 +85,13 @@ class AsMemberOfProcessor : AbstractTestProcessor() {
                 }
             }
         }
+        val javaImpl = resolver.getClassDeclarationByName("JavaImpl")!!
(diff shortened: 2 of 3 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -754,6 +754,13 @@ class ResolverImpl(
                 )
             }
             val typeSubstitutor = containing.kotlinType.createTypeSubstitutor()
+            if (declaration is PropertyAccessorDescriptor) {
+                val substitutedProperty = (declaration.correspondingProperty).substitute(typeSubstitutor)
+                // TODO: Avoid relying on naming convention.
+                // Fix in upstream for property accessors: https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/descriptors/impl/PropertyAccessorDescriptorImpl.java#L122
+                val prefix = function.simpleName.asString().substring(0,3)
+                    return KSFunctionImpl((substitutedProperty as PropertyDescriptor).accessors.single { it.name.asString().substring(1,4) == prefix })
```
ting-yuan, line 762: Would `it.name == declaration.name` make sense? So that you don't rely on the naming convention.

ting-yuan (approved):

PR #479, opened 2021-06-12
default symbol validation should not check function body

### LINKED ISSUE #368: Can't resolve property type when type is implicit in assignment
**KSP version**: 1.4.31-1.0.0-alpha06

Using `KSValidateVisitor` (the one used by default by `KSNode.validate`), the following class can't be validated because the type of the variable `x` inside `baz()` can't be resolved.
Making `x` immutable, doesn't fix the issue.

```kotlin
@DummyAnnotation
class Foo {
  val bar: Int = 0

  fun baz(): Int {
    var x = bar
    return x
  }
}
```

Instead, specifying the type of `x` explicitly, fixes the issue:

```kotlin
@DummyAnnotation
class Foo {
  val bar: Int = 0

  fun baz(): Int {
    var x: Int = bar
    return x
  }
}
```

neetopia:
While this does not fundamentally fix #368 , default validate still should not consider  function body for performance.

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
@@ -99,7 +99,9 @@ fun KSDeclaration.isLocal(): Boolean {
  * Perform a validation on a given symbol to check if all interested types in symbols enclosed scope are valid, i.e. resolvable.
  * @param predicate: A lambda for filtering interested symbols for performance purpose. Default checks all.
  */
-fun KSNode.validate(predicate: (KSNode?, KSNode) -> Boolean = { _, _-> true } ): Boolean {
+fun KSNode.validate(predicate: (KSNode?, KSNode) -> Boolean = { parent, current ->
+    true
+} ): Boolean {
     return this.accept(KSValidateVisitor(predicate), null)
 }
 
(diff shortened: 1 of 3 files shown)


gdude2002 (approved):

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -99,7 +99,9 @@ fun KSDeclaration.isLocal(): Boolean {
  * Perform a validation on a given symbol to check if all interested types in symbols enclosed scope are valid, i.e. resolvable.
  * @param predicate: A lambda for filtering interested symbols for performance purpose. Default checks all.
  */
-fun KSNode.validate(predicate: (KSNode?, KSNode) -> Boolean = { _, _-> true } ): Boolean {
+fun KSNode.validate(predicate: (KSNode?, KSNode) -> Boolean = { parent, current ->
+    !(parent is KSFunctionDeclaration && current is KSDeclaration)
+} ): Boolean {
```
ting-yuan, line 104: Can we simply make `visitFunctionDeclaration` not calling `visitDeclarationContainer`?

neetopia: I want to leave the door open for those who still want to check for function body.

ting-yuan (approved):

PR #484, opened 2021-06-18
Support K/JS and K/Metadata

ting-yuan:
Currently, only a single `ksp()` configuration is supported in KMP, because the targets are added in configuration. Configurations like `kspJsMain` doesn't seem to be possible unless we built them unconditionally from presets. Unsupported targets are skipped by the else case in `val kspTaskClass = when (kotlinCompileTask) { ... }` around KspSubplugin.kt:246.

Also changed the path of KSP's output so as to accommodate different targets.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -37,6 +37,7 @@ import org.gradle.api.tasks.compile.JavaCompile
 import org.gradle.language.jvm.tasks.ProcessResources
 import org.gradle.tooling.provider.model.ToolingModelBuilderRegistry
 import org.jetbrains.kotlin.cli.common.arguments.Argument
+import org.jetbrains.kotlin.cli.common.arguments.CommonCompilerArguments
 import org.jetbrains.kotlin.cli.common.arguments.K2JVMCompilerArguments
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmOptionsImpl
 import org.jetbrains.kotlin.gradle.dsl.KotlinSingleTargetExtension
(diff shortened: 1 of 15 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -381,12 +403,124 @@ abstract class KspTask : KotlinCompile(KotlinJvmOptionsImpl()) {
...
+        args: K2JSCompilerArguments,
+        sourceRoots: SourceRoots,
+        changedFiles: ChangedFiles
+    ) {
+        args.addChangedFiles(changedFiles)
+        super.callCompilerAsync(args, sourceRoots, changedFiles)
+    }
+}
+
+abstract class KspTaskMetadata : KotlinCompileCommon(KotlinMultiplatformCommonOptionsImpl()), KspTask {
```
neetopia, line 464: Nit: there are some common code between different platforms.

integration-tests/src/test/resources/kmp/test-processor/src/main/kotlin/TestProcessor.kt
```diff
@@ -0,0 +1,56 @@
...
+class ClassVisitor : KSTopDownVisitor<OutputStreamWriter, Unit>() {
+    override fun defaultHandler(node: KSNode, data: OutputStreamWriter) {
+    }
+
+    override fun visitClassDeclaration(
+        classDeclaration: KSClassDeclaration,
+        data: OutputStreamWriter
+    ) {
+        super.visitClassDeclaration(classDeclaration, data)
+        //val symbolName = classDeclaration.qualifiedName!!.asString().replace('.', '_')
```
neetopia, line 45: unused

integration-tests/src/test/resources/kmp/workload/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+version = "1.0-SNAPSHOT"
+
+kotlin {
+    jvm {
+        withJava()
+    }
+//    linuxX64()
+//    mingwX64()
+//    macosX64()
+//    ios()
```
neetopia, line 15: unused

neetopia (approved):
LGTM, with a few minor nits.

CI fails with 
```
Lint infrastructure error
...
Caused by: java.lang.OutOfMemoryError: Metaspace
```

gavra0 (commented):
LGTM

PR #488, opened 2021-06-24
support Gradle 7.0's annotation requirements

### LINKED ISSUE #487: KspTask.configure()'s `@Internal` annotation is incompatible with Gradle 7.x
```
A problem was found with the configuration of task ':app:kspKotlin' (type 'KspTaskJvm').
  - In plugin 'com.google.devtools.ksp' type 'com.google.devtools.ksp.gradle.KspTaskJvm' method 'configure()' should not be annotated with: @Internal.
    
    Reason: Input/Output annotations are ignored if they are placed on something else than a getter.
    
    Possible solutions:
      1. Remove the annotations.
      2. Rename the method.
    
    Please refer to https://docs.gradle.org/7.1/userguide/validation_problems.html#ignored_annotations_on_method for more details about this problem.
```

`configure` also shadows Gradle's `configure()` function name.

This wasn't caught by tests/CI because everything is executing against Gradle 6.7.1.

RBusarow:
fixes #487

I updated the project to Gradle 7.1 in order to get the tests to run against it.  Quite a few tests failed after the update was applied, before fixing the annotations.

It's also possible to set the Gradle version to a different version while building the `GradleRunner`.  I decided against that because I'm not sure there's value to building the project using 6.7.1.  Also, there's no centralized function for creating the runner, so that would mean a much larger diff and it'd still be error prone.


RBusarow:
@googlebot I signed it!

ting-yuan (approved):
Thanks for catching and fixing this!

drewhamilton:
Any chance of a quick patch release with this? My project is using Gradle 7's version catalogs feature, so it'd be quite an effort to roll back to Gradle 6; yet I'd also like to support Kotlin 1.5.20 as soon as possible.

ting-yuan:
Sure. We'll do another release early next week, just in case there are other issues that need another release.

ting-yuan:
The fix is in: https://github.com/google/ksp/releases/tag/1.5.20-1.0.0-beta04

Thank you for the fix again!

PR #499, opened 2021-07-01
Fix typo in filename of SymbolProcessorEnvironment.

danysantiago:
This is a super small file rename, no API change since the class name is actually fine, it's just the filename that has a typo.


neetopia (approved):

neetopia:
Thanks for PR!

Issue #503, opened 2021-07-07
`KSValueParameter` has no reference to the `KSDeclaration` where is present


gmazzo:
A few weeks ago, I've reported this issue https://github.com/google/ksp/issues/351 regarding the lack of `KSValueParameter` on `getSymbolsWithAnnotation`.

Today we just discovered a very strange behavior regarding incremental compilation, where the generated classes were missing after the second run.

After a deep research, we realized the root cause was this peace of logic here (that we have in our code):
```kotlin
 environment.codeGenerator.createNewFile(
            dependencies = Dependencies(
                aggregating = true,
                sources = source
                    .flatMap { it.sources }
                    .mapNotNull { (it as? KSDeclaration)?.containingFile }
                    .distinct()
                    .toTypedArray()
            ),
```
So, as you may note, `KSValueParameter` can never be cast to `KSDeclaration` (which makes sense) but only `KSDeclaration` contains a `KSFile` instance to be used in the `Dependencies`.

We need a way in the API to be able to map `KSValueParameter` to a `KSFile` and right now it doesn't seems to have one right? 

My suggestion will be to add something like:
```kotlin
interface KSValueParameter : KSAnnotated {

     /**
      *  The reference to the function declaring this parameter.
      */
     val function: KSFunctionDeclaration
```

neetopia:
fixed in #573

PR #504, opened 2021-07-08
Added `containingFile` member to `KSValueParameter` interface

gmazzo:
This PR addresses https://github.com/google/ksp/issues/503 by adding `containingFile` to `KSValueParameter` interface signature.

To avoid repeating code, I took the opportunity to do a small refactor by introducing the `KSFileSymbol` (any better name?) interface and making `KSDeclaration` and `KSValueParameter` implement it.

I didn't add a test for this, but I'd be willing to. I'd to have an early validation of the approach first.

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSValueParameter.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSValueParameter.kt
@@ -33,6 +33,11 @@ interface KSValueParameter : KSAnnotated {
     val type: KSTypeReference
 
     /**
+     * The containing source file of this declaration, can be null if symbol does not come from a source file, i.e. from a class file.
+     */
+    val containingFile: KSFile?
+
+    /**
      * True if it is a vararg.
      */
     val isVararg: Boolean
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSValueParameterDescriptorImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSValueParameterDescriptorImpl.kt
@@ -41,6 +41,8 @@ class KSValueParameterDescriptorImpl private constructor(val descriptor: ValuePa
         descriptor.annotations.asSequence().map { KSAnnotationDescriptorImpl.getCached(it) }
     }
 
+    override val containingFile: KSFile? = null
+
     override val isCrossInline: Boolean = descriptor.isCrossinline
 
     override val isNoInline: Boolean = descriptor.isNoinline
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSValueParameterJavaImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSValueParameterJavaImpl.kt
@@ -23,8 +23,10 @@ import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.symbol.impl.KSObjectCache
 import com.google.devtools.ksp.symbol.impl.kotlin.KSNameImpl
 import com.google.devtools.ksp.symbol.impl.toLocation
+import com.intellij.psi.PsiJavaFile
 
 class KSValueParameterJavaImpl private constructor(val psi: PsiParameter) : KSValueParameter {
+
     companion object : KSObjectCache<PsiParameter, KSValueParameterJavaImpl>() {
         fun getCached(psi: PsiParameter) = cache.getOrPut(psi) { KSValueParameterJavaImpl(psi) }
     }
@@ -59,6 +61,10 @@ class KSValueParameterJavaImpl private constructor(val psi: PsiParameter) : KSVa
 
     override val type: KSTypeReference by lazy {
         KSTypeReferenceJavaImpl.getCached(psi.type)
+    }
+
+    override val containingFile: KSFile? by lazy {
+        KSFileJavaImpl.getCached(psi.containingFile as PsiJavaFile)
     }
 
     override val hasDefault: Boolean = false
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSValueParameterImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSValueParameterImpl.kt
@@ -30,6 +30,7 @@ import org.jetbrains.kotlin.psi.KtParameter
 import org.jetbrains.kotlin.psi.KtProperty
 
 class KSValueParameterImpl private constructor(val ktParameter: KtParameter) : KSValueParameter {
+
     companion object : KSObjectCache<KtParameter, KSValueParameterImpl>() {
         fun getCached(ktParameter: KtParameter) = cache.getOrPut(ktParameter) { KSValueParameterImpl(ktParameter) }
     }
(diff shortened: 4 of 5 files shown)


neetopia:
Thanks for your PR. It is a good idea overall. Parent information is available when you are visiting from the top level in a visitor.

Speaking of the PR, I think it is missing implementation for synthetic declarations where while these declarations itself does not have concrete symbol in the code, there parent does, and therefore they belong to a file in a case. What looks better is to have `parent` implemented for KSNode, in which case not only you can get the `containingFile` by looking up parent iteratively, you also get access to the parent if you want.

gmazzo:
> Thanks for your PR. It is a good idea overall. Parent information is available when you are visiting from the top level in a visitor.
> 
> Speaking of the PR, I think it is missing implementation for synthetic declarations where while these declarations itself does not have concrete symbol in the code, there parent does, and therefore they belong to a file in a case. What looks better is to have `parent` implemented for KSNode, in which case not only you can get the `containingFile` by looking up parent iteratively, you also get access to the parent if you want.

I agree with the `parent` approach as well. I did a quick review and it will pretty much impact all the KS definition classes.

Can't you make the two approaches coexist? I mean, we already have `containingFile` on `KSDeclaration`. With this proposal of `parent`, that method will be a shortcut, but still redundant.

gmazzo:
I just found another "issue" with the `parent` approach:

Do you on many of the implementation classes a `cache` like this:
```kotlin
class KSTypeArgumentKtImpl private constructor(val ktTypeArgument: KtTypeProjection) : KSTypeArgumentImpl() {
    companion object : KSObjectCache<KtTypeProjection, KSTypeArgumentKtImpl>() {
        fun getCached(ktTypeArgument: KtTypeProjection) = cache.getOrPut(ktTypeArgument) { KSTypeArgumentKtImpl(ktTypeArgument) }
    }
```

I'm not sure what have you in mind, but if I have to add also a reference to its parent, I'd modify the constructor like this:
```
class KSTypeArgumentKtImpl private constructor(override val parent: KSNode?, val ktTypeArgument: KtTypeProjection) : KSTypeArgumentImpl() {
```
I may be wrong, but that change will make the `KSObjectCache` pretty much useless, right?

neetopia:
For the `cache` issue, you need to make it a `Pair` for dual constructor parameter, see `KSTypeParameterImpl` for example.

For `parent`, yes it will impact all KSNode classes, but I feel this is the right way to do so. `parent` is more general and flexible than `containingFile`, and we hope to avoid redundant APIs. We can make `containingFile` an extensionFunction for KSNode which will look up through its parent list after implementing `parent` on `KSNode`.

gmazzo:
> For the `cache` issue, you need to make it a `Pair` for dual constructor parameter, see `KSTypeParameterImpl` for example.

Will it make sense then to cache `Pair< parent KSNode? , KSTypeParameterImpl>`? That's pretty much a unique combination to me unless that for some reason you have this logic in multiple places. But I think is doable.


> For `parent`, yes it will impact all KSNode classes, but I feel this is the right way to do so. `parent` is more general and flexible than `containingFile`, and we hope to avoid redundant APIs. We can make `containingFile` an extensionFunction for KSNode which will look up through its parent list after implementing `parent` on `KSNode`.

So, you want to convert `containingFile` to an extension function then. What should happen with the `containingFile` function already in `KSDeclaration` interface? Will you remove it?

I may missed something, but if you just move `containingFile` to `KSNode` with a default implementation `parent?.containingFile` you will still have the hierarchy approach you suggested.
Why an extension value will better? I actually see a performance price of this (due traversing) and will be losing the possibility to:
1) cache the value, with lazy
2) take advantage of saving computation time, by getting the data from an inner layer (like `KtPureElement.containingKtFile` in `KSDeclaration`)

neetopia:
At API level we want to stay minimalist so that we want to avoid APIs that duplicates on functions. It should be generally inexpensive to travel through the parent list to get containing file as most of the time it is not too deep. Plus that you can still write your own cache in your processor if this becomes a pain point.

gmazzo:
An what about this comment?
> Will it make sense then to cache Pair< parent KSNode? , KSTypeParameterImpl>?

I see all over the code the cache for inner layer symbols like this one that caches `KtTypeProjection`:
```kotlin
class KSTypeArgumentKtImpl private constructor(val ktTypeArgument: KtTypeProjection) : KSTypeArgumentImpl() {
    companion object : KSObjectCache<KtTypeProjection, KSTypeArgumentKtImpl>() {
        fun getCached(ktTypeArgument: KtTypeProjection) = cache.getOrPut(ktTypeArgument) { KSTypeArgumentKtImpl(ktTypeArgument) }
    }
```
I'm quite sure why this was introduced. Assuming for instance this was introduced because the same instance of `KtTypeProjection` is seen at different nodes of the tree you wanted to honor that and have the same instance of it's mapper `KSTypeArgumentKtImpl`.
If now we cache by a `Pair< KSNode? , KtTypeProjection>` it's not just the symbol but also where is found. Then the cache is pretty much pointless right? And it may break something else.

neetopia:
`KtTypeProjection` is bounded to a certain location in source file so this won't be an issue. I.e. `KtTypeProjection` is corresponding to a specific symbol in the source code, so if it is a different node in AST it should be a different instance for this.

gmazzo:
I've reduced my PR to the minimum changes to address https://github.com/google/ksp/issues/503. It would be great if you can merge it so I can unblock my team (the lack of this API is preventing us to fully migrate our processors to KSP).

Supporting this on `KSValueParameterSyntheticImpl` will require to scale `containingFile` to `KSPropertySetter` interface, and it's my understanding you don't want do to that.

Regarding the `KSNode.parent` change you are requesting. As we discussed already, the refactor is HUGE and I'm still not clear about the performance impact of it. To be honest, I don't want to be blamed for introducing such a thing.
`KtDeclaration` has a `parentDeclaration` field on it, and it does not have a `companion object : KSObjectCache`. I'm suspecting this is for the same reason I'm stating in [my previous comment](https://github.com/google/ksp/pull/504#issuecomment-880493872): introducing `parent` to `KSNode` will turn the cache strategy useless. 
If you are fine with it, I may allocate some time to address the refactor, but later. First I need to unblock my team.

neetopia:
It is understandable that refactoring can be huge, I can proceed with the refactoring on our side if that works. Just wonder if you are ok with an extension function approach for `containingFile` (yes it might have performance impact for not caching result, it should be possible to add a cache on processor side still.)

gmazzo:
> It is understandable that refactoring can be huge, I can proceed with the refactoring on our side if that works. Just wonder if you are ok with an extension function approach for `containingFile` (yes it might have performance impact for not caching result, it should be possible to add a cache on processor side still.)

I'm fine with it, yes. Actually, I thought a little improvement to take advantage of the "known file" when available. The problem is the whole parent refactor I need to do to reach there 😅 .

This was my draft idea for the extension function:
```kotlin
internal interface KnowsOwnerFile {
    val containingFile: KSFile?
}

val KSNode.containingFile: KSFile?
    get() = when (this) {
        is KnowsOwnerFile -> containingFile
        is KSFile -> this
        else -> parent?.containingFile
    }
```
So with this `internal KnowsOwnerFile` interface, you can still keep the performance gain by implementing it whenever you know the owner file from the inner layer, without polluting the public API.

Sadly for some reason, `tailrec val` is not supported (yet?). You may also consider using a `tailrec fun` but I kind of prefer the extension value for this.

neetopia:
After some internal discussion, we decided to go with the route that adding `parent` to `KSNode` and provide `containingFile` as an extension function, as a result, current `containingFile` API on `KSDeclaration` will be removed. The concern for non-cached `containingFile` shouldn't be a big concern since in most cases it is only 2-3 layers of look up, and can be easily implemented with a loop. I can try to deliver the aforementioned change in a few weeks, will that work for you?

gmazzo:
Sure! Meanwhile, I've used reflection to workaround the issue. I'll close this PR then

PR #508, opened 2021-07-13
Add support for KSAnnotated annotation methods 

AlexBeggs:
- mirror get annotation methods from Java
- Added support for the following extension functions 
   - KSAnnotated#isAnnotationPresent
   - KSAnnotated#getAnnotationsByType

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
@@ -17,6 +17,7 @@
 package com.google.devtools.ksp.symbol.impl
 
 import com.google.devtools.ksp.ExceptionMessage
+import com.google.devtools.ksp.KspExperimental
 import com.google.devtools.ksp.MemoizedSequence
 import com.google.devtools.ksp.processing.impl.ResolverImpl
 import com.google.devtools.ksp.symbol.*
@@ -58,6 +59,11 @@ import org.jetbrains.kotlin.types.StarProjectionImpl
 import org.jetbrains.kotlin.types.TypeProjectionImpl
 import org.jetbrains.kotlin.types.replace
 import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+import java.lang.reflect.InvocationHandler
+import java.lang.reflect.Method
+import java.lang.reflect.Proxy
+import java.util.concurrent.ConcurrentHashMap
+import kotlin.reflect.KClass
 
 private val jvmModifierMap = mapOf(
     JvmModifier.PUBLIC to Modifier.PUBLIC,
@@ -176,7 +182,8 @@ fun MemberDescriptor.toKSModifiers(): Set<Modifier> {
         DescriptorVisibilities.PUBLIC -> modifiers.add(Modifier.PUBLIC)
         DescriptorVisibilities.PROTECTED,
         JavaDescriptorVisibilities.PROTECTED_AND_PACKAGE,
-        JavaDescriptorVisibilities.PROTECTED_STATIC_VISIBILITY -> modifiers.add(Modifier.PROTECTED)
+        JavaDescriptorVisibilities.PROTECTED_STATIC_VISIBILITY,
+        -> modifiers.add(Modifier.PROTECTED)
         DescriptorVisibilities.PRIVATE -> modifiers.add(Modifier.PRIVATE)
         DescriptorVisibilities.INTERNAL -> modifiers.add(Modifier.INTERNAL)
         // Since there is no modifier for package-private, use No modifier to tell if a symbol from binary is package private.
@@ -228,7 +235,9 @@ fun PsiElement.findParentAnnotated(): KSAnnotated? {
         is PsiJavaFile -> null
         is PsiMethod -> KSFunctionDeclarationJavaImpl.getCached(parent)
         is KtProperty -> KSPropertyDeclarationImpl.getCached(parent)
-        is KtPropertyAccessor -> if (parent.isGetter) { KSPropertyGetterImpl.getCached(parent) } else {
+        is KtPropertyAccessor -> if (parent.isGetter) {
+            KSPropertyGetterImpl.getCached(parent)
+        } else {
             KSPropertySetterImpl.getCached(parent)
         }
         is KtTypeAlias -> KSTypeAliasImpl.getCached(parent)
@@ -490,7 +499,7 @@ internal fun PropertyDescriptor.hasBackingFieldWithBinaryClassSupport(): Boolean
  */
 internal object BinaryFieldsCache : KSObjectCache<ClassId, Set<Name>>() {
     fun getCached(
-        kotlinJvmBinaryClass: KotlinJvmBinaryClass
+        kotlinJvmBinaryClass: KotlinJvmBinaryClass,
     ) = cache.getOrPut(kotlinJvmBinaryClass.classId) {
         val visitor = PropNamesVisitor()
         kotlinJvmBinaryClass.visitMembers(visitor, null)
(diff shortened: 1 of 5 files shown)


AlexBeggs:
@googlebot I signed it!

AlexBeggs:
@googlebot I fixed it

AlexBeggs:
@googlebot I fixed it.

AlexBeggs:
@googlebot I signed it!

AlexBeggs:
@googlebot I fixed it.

compiler-plugin/src/test/kotlin/com/google/devtools/ksp/processor/AnnotatedUtilProcessor.kt
```diff
@@ -0,0 +1,129 @@
...
+        val booleanValue: Boolean = false,
+        val byteValue: Byte = 2,
+        val charValue: Char = 'b',
+        val doubleValue: Double = 3.0,
+        val floatValue: Float = 4.0f,
+        val intValue: Int = 5,
+        val longValue: Long = 6L,
+        val stringValue: String = "emptystring",
+        // fails on getting the arguments from the KSAnnotation when no value is set for the kClassValue in
+        // the declaration. Throws an NPE with using a default value
```
popovanton0, line 98: I also faced this problem https://github.com/google/ksp/issues/512

compiler-plugin/src/test/kotlin/com/google/devtools/ksp/processor/AnnotatedUtilProcessor.kt
```diff
@@ -0,0 +1,129 @@
...
+                .let { data.add(it.toString()) }
+            annotated.getAnnotationsByType(Throws::class).let { data.add(it.toString()) }
+        }
+
+        override fun defaultHandler(node: KSNode, data: MutableCollection<String>) {
+        }
+    }
+
+    @Suppress("LongParameterList")
+    annotation class ParametersTestAnnotation(
```
popovanton0, line 88: Does your code support instances of `kotlin.Annotation` subclasses as a type of annotation class values? Kotlin documentation states:
> Allowed parameter types:
> - [...]
> - Other annotations
> - [...]

[Source](https://kotlinlang.org/docs/annotations.html#constructors)

AlexBeggs: No it doesn't, I didn't take that into consideration. I can add that functionality.

AlexBeggs: This is a port from my repo https://github.com/AlexBeggs/ksp-annotation-parser, I am currently implementing tests for this repo, but I ran into some issues over the weekend. Finally got it worked out last night.

popovanton0: I really need this functionality. It would be very nice if it would be added to ksp :)

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -484,3 +488,101 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
         is KtParameter -> declaration.defaultValue != null
         else -> false
     }
+
+fun <T : Annotation> KSAnnotated.getAnnotation(annotationKClass: KClass<T>): T? {
+    return this.annotations.firstOrNull { annotationKClass.qualifiedName == it.annotationType.resolve().declaration.qualifiedName?.asString() }
```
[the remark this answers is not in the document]
AlexBeggs: Not sure I am following, the signature is a single Annotation or null this complies with the Java version of this method. Is this not the signature we want to use?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -484,3 +487,140 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
...
+    val cache = ConcurrentHashMap<Pair<Class<*>, Any>, Any>(arguments.size)
+    return InvocationHandler { proxy, method, _ ->
+        if (method.name == "toString" && arguments.none { it.name?.asString() == "toString" }) {
+            "${clazz.canonicalName}" +
+                arguments.map { argument: KSValueArgument ->
+                    // handles default values for enums otherwise returns null
+                    val methodName = argument.name?.asString()
+                    val value = proxy.javaClass.methods.find { m -> m.name == methodName }?.invoke(proxy)
+                    "$methodName=$value"
+                }.toList()
```
AlexBeggs, line 531: This handles the `toString` method and obtaining the default values by reusing the implemented proxy methods

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -484,3 +487,140 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
...
+        } else {
+            val argument = try {
+                arguments.first { it.name?.asString() == method.name }
+            } catch (e: NullPointerException) {
+                throw IllegalArgumentException("This is a bug using the default KClass for an annotation", e)
+            }
```
AlexBeggs, line 537: There is a known issue https://github.com/google/ksp/issues/512

compiler-plugin/src/test/kotlin/com/google/devtools/ksp/processor/AnnotatedUtilProcessor.kt
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+
+        override fun defaultHandler(node: KSNode, data: MutableCollection<String>) {
+        }
+    }
+}
+
+@Suppress("LongParameterList")
+annotation class ParametersTestAnnotation(
```
AlexBeggs, line 84: had to define the annotations twice (in test file and in processor) so the processor could use it in the getAnnotation and getAnnotationByType function calls

compiler-plugin/testData/api/annotatedUtil.kt
```diff
@@ -0,0 +1,160 @@
...
+    NONE, VALUE1, VALUE2
+}
+
+annotation class InnerAnnotation(val value: String = "default")
+
+annotation class OuterAnnotation(
+    val innerAnnotation : InnerAnnotation = InnerAnnotation()
+)
+
+/////////////////////////////////////////////////////////
```
AlexBeggs, line 110: Note: I wasn't able to figure out how to properly setup Java annotation testing

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
neetopia: this issue was just fixed in latest release.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -484,3 +488,140 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
         is KtParameter -> declaration.defaultValue != null
         else -> false
     }
+
+fun <T : Annotation> KSAnnotated.getAnnotation(annotationKClass: KClass<T>): T? {
+    return this.annotations.firstOrNull {
```
neetopia, line 493: My concern here is on the [repeating annotations](https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html), in this case, you might need to think about whether returning the first instance of a repeating annotation is desired.

Meanwhile there is no such issue with `KSAnnotated.getAnnotationsByType(annotationKClass: KClass<T>)`

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -484,3 +488,140 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
...
+}
+
+fun <T : Annotation> KSAnnotated.getAnnotationsByType(annotationKClass: KClass<T>): Sequence<T> {
+    return this.annotations.filter {
+        it.annotationType.resolve().declaration.qualifiedName?.asString() == annotationKClass.qualifiedName
+    }.map { it.toAnnotation(annotationKClass) }
+}
+
+fun <T : Annotation> KSAnnotated.isAnnotationPresent(annotationKClass: KClass<T>): Boolean =
+    getAnnotation(annotationKClass)?.let { true } ?: false
```
neetopia, line 512: this can be `getAnnotation(annotationKClass) != null`

AlexBeggs: good call

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
AlexBeggs: @neetopia I was mirroring this API https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/AnnotatedElement.html#getAnnotation-java.lang.Class- for consistency

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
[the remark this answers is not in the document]
AlexBeggs: It doesn't state it in the Javadoc, but there might be a AnnotationFormatError that gets thrown when more than one is defined and you call getAnnotation (going to validate that tonight)

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
AlexBeggs: will rebase later

AlexBeggs: I rebased it and still getting the issue. I created an example of the issue I am running into  https://github.com/google/ksp/issues/538
Must be a different issue

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
AlexBeggs: Based on the [repeating annotations](https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html) documentation that you sent it also states "The behavior of the methods that return a single annotation, such as AnnotatedElement.getAnnotation(Class<T>), are unchanged in that they only return a single annotation if one annotation of the requested type is present. If more than one annotation of the requested type is present, you can obtain them by first getting their container annotation". So this should be satisfied by this behavior and consistent.

I was also keeping the KSAnnotated (KSP Kotlin) and AnnotatedElement (Java) as much in sync as possible with these methods for easy adoption and based on the Slack conversation that we had.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -533,3 +539,140 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
...
+fun <T : Annotation> KSAnnotated.getAnnotation(annotationKClass: KClass<T>): T? {
+    return this.annotations.firstOrNull {
+        annotationKClass.qualifiedName == it.annotationType.resolve().declaration.qualifiedName?.asString()
+    }?.toAnnotation(annotationKClass)
+}
+
+fun KSAnnotated.getAnnotations(): Sequence<out Annotation> {
+    return this.annotations.map {
+        val kClass = Class.forName(it.annotationType.resolve().declaration.qualifiedName?.asString()).kotlin
+        it.toAnnotation(kClass as KClass<Annotation>)
```
ting-yuan, line 552: What if some of the annotations aren't available at compile time? I.e., when `Class.forName` fails.

IMHO, a closer analogy to `KSAnnotated` in Java API might be [AnnotatedConstruct](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/AnnotatedConstruct.html), rather than `AnnotatedElement`. If we really want all annotations without specifying `KClass` or `Class` (being able to write that in processor ensures their availability at processor's runtime / user code's compile time), we need to define the behavior. I recommend removing this in this PR and add it in a later PR.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -533,3 +539,140 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
...
+    }
+}
+
+fun <T : Annotation> KSAnnotated.getAnnotationsByType(annotationKClass: KClass<T>): Sequence<T> {
+    return this.annotations.filter {
+        it.annotationType.resolve().declaration.qualifiedName?.asString() == annotationKClass.qualifiedName
```
ting-yuan, line 558: An "early return" optimization will be very appreciated. Something like:
`it.simpleName.asString() == annotationKClass.simpleName && ...`

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
AlexBeggs: I see, I didn't realize he was suggesting removal altogether, and not mirroring the Java API. That's fine I can understand that.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
AlexBeggs: Maybe I misunderstand, but wouldn't the Annotation classes have to be available if we were compiling the codebase, otherwise wouldn't it fail compilation in general?

AlexBeggs: Or is it possible to have a different compile classpath vs the processors classpath? If so then that makes sense to remove this.

ting-yuan: Yes. For example, the processor won't be able to instantiate `A`, because it isn't even compiled yet. I think that is why methods in `AnnotatedConstruct` require a `Class<A>`; that makes sure `A` is instantiable in processor.

```
// Same file
annotation class A

@A
class C
```

ting-yuan:
Another thing that I forgot to mention: Can we annotate those functions with `@KspExperimental`? Apparently the implementation is JVM only, but I haven't thought thoroughly about how it should behave on other platforms. Any idea?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
AlexBeggs: done

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -533,3 +543,133 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
...
+                        }
+                        method.returnType.isAnnotation -> {
+                            val value = { (result as KSAnnotation).asAnnotation(method.returnType) }
+                            cache.getOrPut(Pair(method.returnType, result), value)
+                        }
+                        method.returnType.name == "java.lang.Class" -> {
+                            val value = { (result as KSType).asClass() }
+                            cache.getOrPut(Pair(method.returnType, result), value)
+                        }
+                        method.returnType.name == "byte" -> {
```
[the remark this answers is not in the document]
AlexBeggs: the values that are returned for the Byte values are in the form an Integer and need to be narrowcast explicitly otherwise a ClassCastException occurs. The other values don't require casting so this isn't an issue and can be returned without casting.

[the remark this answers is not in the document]
ting-yuan: Thanks for the investigation. Does [short](https://kotlinlang.org/docs/basic-types.html) need to be handled as well?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -533,3 +543,133 @@ private val PropertyDescriptor.declaresDefaultValue: Boolean
...
+    return Proxy.newProxyInstance(
+        annotationClass.classLoader,
+        arrayOf(annotationClass),
+        createInvocationHandler(annotationClass)
+    ) as T
+}
+
+@Suppress("TooGenericExceptionCaught")
+private fun KSAnnotation.createInvocationHandler(clazz: Class<*>): InvocationHandler {
+    val cache = ConcurrentHashMap<Pair<Class<*>, Any>, Any>(arguments.size)
```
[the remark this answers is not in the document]
AlexBeggs: I was going to refactor this to use the `result` only as the key. However that is problematic in the cases with Byte or something that might cast to another result type. As you could have the same value for a Byte and an Integer return value but this would get overridden by the last cached value. So you might end up returning an Integer for a Byte type because the values were the same, which would cause a ClassCastException to occur. So keeping it as a `Pair`.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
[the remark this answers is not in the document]
AlexBeggs: good catch, missed that type. 

To add more clarification, to the casting issues. It seems that this is an issue really only for the Annotations written in Java. Kotlin provides the correct class, however Java provides an Integer.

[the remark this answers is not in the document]
AlexBeggs: Add an update and addressed the missing Short value

ting-yuan (approved):

ting-yuan:
Thanks for the PR!

Issue #496, opened 2021-06-29
Execution failed for task ':foo:kspKotlin'. > Extension of type 'KspExtension' does not exist. 


ursusursus:
After upgrading to beta-04 project won't compile. Note the only ksp use case is `moshi-ksp`. There I've been told it seems a upstream issue.

Let me know if it is obvious to you from the trace, I've not tried a minimal example yet

```
Execution failed for task ':user:contract:kspKotlin'.
> Extension of type 'KspExtension' does not exist. Currently registered extension types: [ExtraPropertiesExtension]

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':user:contract:kspKotlin'.
	at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:38)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:200)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:195)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$3.execute(DefaultBuildOperationRunner.java:75)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$3.execute(DefaultBuildOperationRunner.java:68)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:153)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:68)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.call(DefaultBuildOperationRunner.java:62)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.lambda$call$2(DefaultBuildOperationExecutor.java:76)
[...]

PR #522, opened 2021-07-19
Register options to processors when registering KSP tasks

ting-yuan:
The KspTask.apOptions is solely used to let Gradle invalidate the build
whenever there is a change in options to processors. Previously it was
implmented by a default function. This commit reimplements it in a more
Gradle idiomatic way to avoid some potential issues with Gradle.


neetopia (approved):

gavra0 (approved):

ting-yuan:
Fixes #496

PR #523, opened 2021-07-20
support checked exceptions declared in class files

### LINKED ISSUE #507: getJvmCheckedException does not work for already compiled files
Looks like the method body in ResolverImpl returns emptySequence for element that in classpath.

This is critical for Room because we override kotlin methods in Java, hence we would need to get thrown types to produce the right code when the code under question is obtained from classpath (happens when a database refers to a Dao coming from a dependency)

neetopia:
fixes #507

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -33,6 +33,7 @@ import com.google.devtools.ksp.symbol.impl.synthetic.*
 import com.intellij.openapi.project.Project
 import com.intellij.psi.*
 import com.intellij.psi.impl.source.PsiClassReferenceType
+import java.io.File
 import org.jetbrains.kotlin.builtins.jvm.JavaToKotlinClassMap
 import org.jetbrains.kotlin.codegen.ClassBuilderMode
 import org.jetbrains.kotlin.codegen.OwnerKind
@@ -42,6 +43,7 @@ import org.jetbrains.kotlin.container.get
 import org.jetbrains.kotlin.container.tryGetService
 import org.jetbrains.kotlin.descriptors.*
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
+import org.jetbrains.kotlin.descriptors.resolveClassByFqName
 import org.jetbrains.kotlin.incremental.components.NoLookupLocation
 import org.jetbrains.kotlin.load.java.components.TypeUsage
 import org.jetbrains.kotlin.load.java.descriptors.JavaForKotlinOverridePropertyDescriptor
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -663,6 +674,32 @@ class ResolverImpl(
...
+                        access: Int,
+                        name: String?,
+                        descriptor: String?,
+                        signature: String?,
+                        exceptions: Array<out String>?
+                    ): MethodVisitor {
+                        if (name == function.simpleName.asString() && jvmDesc == descriptor) {
+                            exceptions?.toList()?.let { exceptionNames.addAll(it) }
+                        }
+                        return object: MethodVisitor(API_VERSION) {
```
ting-yuan, line 697: nit: should this simply return `super.visitMethod(...)`?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -663,6 +674,32 @@ class ResolverImpl(
...
                 extractThrowsAnnotation(function)
             }
+            Origin.KOTLIN_LIB, Origin.JAVA_LIB -> {
+                val descriptor = (function as KSFunctionDeclarationDescriptorImpl).descriptor
+                val jvmDesc = this.mapToJvmSignature(function)
+                val virtualFileContent = if(function.origin == Origin.KOTLIN_LIB) {
+                    (descriptor.getContainingKotlinJvmBinaryClass() as VirtualFileKotlinClass).file.contentsToByteArray()
+                } else {
+                    (((descriptor.source as JavaSourceElement).javaElement as BinaryJavaMethod).containingClass as BinaryJavaClass).virtualFile.contentsToByteArray()
+                }
```
ting-yuan, line 684: could those `as` ever fail? If so, returning empty or an error sentinel seems better than a hard exception?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -663,6 +674,35 @@ class ResolverImpl(
             Origin.KOTLIN -> {
                 extractThrowsAnnotation(function)
             }
+            Origin.KOTLIN_LIB, Origin.JAVA_LIB -> {
+                val descriptor = (function as KSFunctionDeclarationDescriptorImpl).descriptor
+                val jvmDesc = this.mapToJvmSignature(function)
+                val virtualFileContent = if(function.origin == Origin.KOTLIN_LIB) {
+                    (descriptor.getContainingKotlinJvmBinaryClass() as VirtualFileKotlinClass).file.contentsToByteArray()
```
ting-yuan, line 681: Here as well

ting-yuan (approved):

PR #524, opened 2021-07-20
Fixed Maven repositories

gmazzo:
This PR updates the repositories list to be able to resolve snapshot versions of Kotlin and Idea dependencies.

I recently cloned this repo to address an issue and noted that Kotlin's Bintray repository was down and the JetBrains's Snapshots one missing.


neetopia (approved):
Thanks for PR, let me sync with upstream to see if this is a temporary outage or permanent repo change before merging.

PR #525, opened 2021-07-21
Incremental: support finer-grained classpath changes

ting-yuan:
This PR uses KAPT's `KaptTask.findClasspathChanges` to find classpath changes and invalidate source files accordingly.

Note that lookups are tracked at symbol level (e.g., class names, function names, etc) in the compiler. Also, `KaptTask.findClasspathChanges` finds changed classes rather than symbols. Therefore `DualLookupTracker` is introduced to emulate source-to-declarations-of-a-class lookups.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
@@ -33,7 +33,6 @@ import org.jetbrains.kotlin.container.get
 import org.jetbrains.kotlin.descriptors.CallableMemberDescriptor
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.incremental.*
-import org.jetbrains.kotlin.incremental.components.LookupTracker
 import org.jetbrains.kotlin.incremental.components.Position
 import org.jetbrains.kotlin.incremental.components.ScopeKind
 import org.jetbrains.kotlin.incremental.storage.BasicMap
@@ -193,10 +192,17 @@ class IncrementalContext(
     private val modified = options.knownModified.map{ it.relativeTo(baseDir) }.toSet()
     private val removed = options.knownRemoved.map { it.relativeTo(baseDir) }.toSet()
 
-    private val lookupTracker: LookupTracker = componentProvider.get()
-    private val lookupCacheDir = options.cachesDir
+    private val dualLookupTracker: DualLookupTracker = componentProvider.get()
     private val PATH_CONVERTER = RelativeFileToPathConverter(baseDir)
-    private val lookupCache = LookupStorage(lookupCacheDir, PATH_CONVERTER)
+
+    private val symbolLookupTracker = dualLookupTracker.symbolTracker
+    private val symbolLookupCacheDir = File(options.cachesDir, "symbolLookups")
+    private val symbolLookupCache = LookupStorage(symbolLookupCacheDir, PATH_CONVERTER)
+
+    // TODO: rewrite LookupStorage to share file-to-id, etc.
+    private val classLookupTracker = dualLookupTracker.classTracker
+    private val classLookupCacheDir = File(options.cachesDir, "classLookups")
+    private val classLookupCache = LookupStorage(classLookupCacheDir, PATH_CONVERTER)
 
     private val sourceToOutputsMap = FileToFilesMap(File(options.cachesDir, "sourceToOutputs"))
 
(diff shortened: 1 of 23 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -288,6 +303,17 @@ class IncrementalContext(
         return visited
     }
 
+    private fun logDirtyFilesByCP(dirtyFiles: Collection<File>) {
+        if (!options.incrementalLog)
+            return
+
+        val logFile = File(logsDir, "kspDirtySetByCP.log")
```
gavra0, line 310: Is this file annotated as task output? If so, it may be an issue if cache is relocatable i.e output from one machine is retrieve on another machine.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingPlugin.kt
```diff
@@ -202,6 +200,14 @@ enum class KspCliOption(
     "log dirty files",
     false,
     false
+    ),
+
+    CHANGED_CLASSES_OPTION(
+    "changedClasses",
+    "<changedClasses>",
+    "dirty classes in classpath",
```
gavra0, line 208: nit: Add if it is internal or canonical name (using / or . as separators).

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -348,11 +366,25 @@ interface KspTask : Task {
     @get:InputFiles
     abstract val processorClasspath: ConfigurableFileCollection
 
-    fun configureCompilation(kotlinCompilation: KotlinCompilationData<*>, kotlinCompile: AbstractKotlinCompile<*>)
+    /**
+     * Output directory that contains caches necessary to support incremental annotation processing.
+     */
+    @get:Internal
```
gavra0, line 372: This should probably be @LocalState?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -361,7 +393,95 @@ abstract class KspTaskJvm : KotlinCompile(KotlinJvmOptionsImpl()), KspTask {
                 kotlinCompile.compilerArgumentsContributor
             }
         )
-        this.useClasspathSnapshot.value(false).disallowChanges()
+
+        isIntermoduleIncremental = project.findProperty("ksp.incremental.intermodule")?.toString().toBoolean()
+        if (isIncremental && isIntermoduleIncremental) {
+            val classStructureIfIncremental = project.configurations.create("_classStructure${name}")
```
gavra0, line 399: You can change this to be a detached configuration i.e use `project.configurations.detachedConfiguration`.

If you want to create a named one (as you are doing right now), you need to create it at the time the task is registered, otherwise it may happen too late because of lazy task configuration.

gavra0 (commented):
Nice reuse of the kapt classpath snapshot work :)

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
ting-yuan: Logs are not annotated at all; They are generated into `build/`.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingPlugin.kt
ting-yuan: Done. ce0b3b1bceef5fcf7f2b4b4f47364336f91b8713

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
ting-yuan: Done and rebased into 4030401

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
ting-yuan: Did you mean this? 8b9dc928be0fe610adae3a5cddad2f168ff19285

gavra0: Exactly 👍

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
gavra0: Is it unique per-task? Also I feel that moving them to the incremental dir (which is annotated with @LocalState) is the way to go. In such way:
- it is not restored from cache when there is a cache hit
- in a non-incremental run you already have logic to clean up the entire dir, so things are consistent and correct
wdyt?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
```diff
@@ -196,20 +193,26 @@ class IncrementalContext(
     private val removed = options.knownRemoved.map { it.relativeTo(baseDir) }.toSet()
 
     private val lookupTracker: LookupTracker = componentProvider.get()
-    private val lookupCacheDir = options.cachesDir
+    // Disable incremental processing if somehow DualLookupTracker failed to be registered.
+    // This may happen when a platform hasn't support incremental compilation yet. E.g, Common / Metadata.
+    private val isIncremental = options.incremental && lookupTracker is DualLookupTracker
```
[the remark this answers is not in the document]
ting-yuan: Unfortunately, in the case of K2MetadataCompiler, it is always LookupTracker.DO_NOTHING. In that case even the basic source-only-deps doesn't work.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/Incremental.kt
ting-yuan: Done: c397a30

neetopia (approved):

PR #528, opened 2021-07-23
support annotation default value for class files compiled from Kotlin source.

neetopia:


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSAnnotationDescriptorImpl.kt
```diff
@@ -122,11 +148,108 @@ fun ClassConstructorDescriptor.getAbsentDefaultArguments(excludeNames: List<Stri
...
     return when (psi) {
         null -> {
-            // TODO: This will only work for symbols from Java class.
-            ResolverImpl.instance.javaActualAnnotationArgumentExtractor.extractDefaultValue(this, this.type)?.toValue()
+            val defaultFromJava = ResolverImpl.instance.javaActualAnnotationArgumentExtractor.extractDefaultValue(this, this.type)?.toValue()
+            if (defaultFromJava != null) {
+                defaultFromJava
+            } else {
+                val file =
+                    (this.containingDeclaration.getContainingKotlinJvmBinaryClass() as VirtualFileKotlinClass).file.contentsToByteArray()
```
ting-yuan, line 227: Can we use safe cast?

ting-yuan (approved):

PR #530, opened 2021-07-27
Enable ktlint in workflow and fix issues

AlexBeggs:
This PR enables ktlint in the workflow and fixes all of the current violations. Merging will have future PRs validated against ktlint.

Currently ktlint is does not run in the workflow allowing PRs to be merged without complying. Either ktlint should be removed or this should be merged. ktlint is enabled in the `./gradlew check` call.


neetopia (approved):
Overall looks good, there is one thing needs to be address before I can merge.

```
rename api/src/main/kotlin/com/google/devtools/ksp/visitor/{KSAllSupersVisitor.kt => KSDefaultVisitor.kt} (99%)
``` 

this appears to be a result of not working on latest HEAD, can you rebase locally with origin to see if this goes away?

AlexBeggs:
> Overall looks good, there is one thing needs to be address before I can merge.
> 
> ```
> rename api/src/main/kotlin/com/google/devtools/ksp/visitor/{KSAllSupersVisitor.kt => KSDefaultVisitor.kt} (99%)
> ```
> 
> this appears to be a result of not working on latest HEAD, can you rebase locally with origin to see if this goes away?

I rebased, it looks like the latest HEAD also has the class name different to the filename, which is why the linter is complaining and why I made the rename of the file. 

https://github.com/google/ksp/blob/main/api/src/main/kotlin/com/google/devtools/ksp/visitor/KSAllSupersVisitor.kt#L26

.github/workflows/main.yml
```diff
@@ -61,6 +61,9 @@ jobs:
       with:
         path: ~/.gradle/wrapper
         key: ${{ runner.os }}-gradle-wrapper-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties') }}
+    # Run ktlint
+    - name: lint
+      run: ./gradlew ktlint
     # Run ksp generated tests
```
AlexBeggs, line 67: @neetopia I added a lint check to the workflow. I am not that familiar with .github workflows so I am unsure if this is correct or not. I also suspect it might run in all 3 environments (any tips)?
Wanted to add this check in (even if it was wrong) to prevent regressions, I believe there might already be regressions since I did a rebase.

AlexBeggs:
@neetopia this now passes with the ktlint as part of the workflow. However, this probably won't be valid for too long after someone commits something.

AlexBeggs:
@neetopia after merging this it should enforce the ktlint ruleset, and this PR currently passes it for now.

.github/workflows/main.yml
AlexBeggs: doesn't look to be an issue

ting-yuan (approved):

ting-yuan:
Thanks for the PR!

PR #534, opened 2021-07-28
Gradle: replace destinationDir with destinationDirectory

ting-yuan:
The former is deprecated and will be removed in Gradle 8.0.


neetopia (approved):

Mohit-Hasan:
What is the generally path in flutter project to replace "destinationDir" this things?

PR #539, opened 2021-08-02
Adds generated resources folder to Android's variant `processJavaResourcesProvider` task

gmazzo:
Replaces `registerPostJavacGeneratedBytecode` by `processJavaResources.from` to allow generated resources by KSP processor to be present on the final classpath of an Android module.

We just discovered this issue today as we are using `ServiceLoader` to find implementation generated by KSP and it was returning none even that `META-INF/services/xxx` was present on `build/generated/ksp/${variantName}/resources`.

I confirmed the bug by introducing this workaround in our code:
```kotlin
        android.applicationVariant.all {
            processJavaResourcesProvider.configure {
                from("$buildDir/generated/ksp/${this@all.name}/resources")
            }
        }
 ```


gmazzo:
Let me add a test case in `AndroidIt.kt`

gmazzo:
Let me close this for now, I think the problem was we weren't using the latest version of KSP

PR #541, opened 2021-08-03
Support Kotlin Native

ting-yuan:
Also publish `symbol-processing-cmdline` to be used with `kotlin-compiler` for command line and K/N.

### DIFF
--- a/symbol-processing/build.gradle.kts
+++ b/symbol-processing/build.gradle.kts
@@ -40,7 +40,7 @@ tasks {
 
 publishing {
     publications {
-        val publication = create<MavenPublication>("shadow") {
+        create<MavenPublication>("shadow") {
             artifactId = "symbol-processing"
             artifact(tasks["sourcesJar"])
             artifact(tasks["shadowJar"])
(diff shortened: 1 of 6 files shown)


symbol-processing/build.gradle.kts
```diff
@@ -68,5 +68,15 @@ publishing {
                 }
             }
         }
+
+        create<MavenPublication>("cmdline") {
```
neetopia, line 72: should this name be `native` instead?

neetopia (approved):

symbol-processing/build.gradle.kts
ting-yuan: No, because it's not native specific; It's more like *unrelocated*. 

FWIW, I was considering renaming:
`symbol-processing` -> `symbol-processing-embeddable`
`symbol-processing-cmdline` -> `symbol-processing`

Because they shouldn't be used directly by users. Well, maybe with the exception of `kotlin-compile-testing`.

Naming is too hard. So finally I settled down with `symbol-processing-cmdline`. Any thoughts?

neetopia: I see, I was having that impression due to the description. Feel free to merge.

PR #543, opened 2021-08-07
Support Gradle build cache

ting-yuan:
and fix a few annotations on properties of KspTask.


neetopia:
integration test for build cache is failing due to build result is `SUCCESS` rather than `FROM_CACHE`

neetopia (approved):

Issue #37, opened 2020-06-10
IDE not indexing generated source


davidjwiner:
Looks like that kapt has an IDE plugin to achieve this: KaptProjectResolverExtension.kt

We either need to copy it over, or simply tell users to mark generated source root themselves in the preview.

ting-yuan:
Deferring again to 2021Q1.

seongug-jung:
```kotlin
buildTypes {
    getByName("debug") {
        sourceSets {
            getByName("main") {
                java.srcDir(File("build/generated/ksp/debug/kotlin"))
            }
        }
    }
    getByName("release") {
        sourceSets {
            getByName("main") {
                java.srcDir(File("build/generated/ksp/release/kotlin"))
            }
        }
    }
}
```
for now.

ZeroBrain:
```kotlin
buildTypes {
    getByName("debug") {
        isMinifyEnabled = false
        proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
        buildConfigField("boolean", "DEBUG_BUILD", "true")
    }

    getByName("release") {
        isMinifyEnabled = false
        proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
        buildConfigField("boolean", "DEBUG_BUILD", "false")
    }
}
```
I am using above script in build.gradle.kts and working as well for now.

keyboardr:
As a slight modification to the above, here's an approach that's a little more compact and flexible:
```kotlin
    applicationVariants.all {
        val variantName = name
        sourceSets {
            getByName("main") {
                java.srcDir(File("build/generated/ksp/$variantName/kotlin"))
            }
        }
    }
```

gmazzo:
We use this logic in our tooling to property support all extension types:
```kotlin
    fun Project.setupKSPOutputsAsGeneratedFolders() = afterEvaluate {
        plugins.withId("com.google.devtools.ksp") {
            the<BaseExtension>().variants.configureEach variant@{
                registerJavaGeneratingTask(
                    tasks.named("ksp${this@variant.name.capitalize(Locale.ROOT)}Kotlin"),
                    layout.buildDirectory.dir("generated/ksp/debug/kotlin").get().asFile
                )
            }
        }
    }

    val BaseExtension.variants: DomainObjectSet<out BaseVariant>
        get() = when (this) {
            is AppExtension -> applicationVariants
[...]

PR #545, opened 2021-08-07
Update Quickstart instructions on IDE mapping of generated files

edrd-f:
It's good for newcomers to know of this workaround while https://github.com/google/ksp/issues/37 is open.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -166,6 +166,21 @@ build/generated/ksp/main/java/
 ```
 
 and perhaps also resource directory if your IDE supports them:
+
 ```
-build/generated/ksp/main/resources
+build/generated/ksp/main/resources/
+```
+
+It may also be necessary to configure these directories in your KSP consumer module:
+
+###### build.gradle.kts
+```kotlin
+kotlin {
+    sourceSets.main {
+        kotlin.srcDir("build/generated/ksp/main/kotlin")
+    }
+    sourceSets.test {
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -166,6 +166,21 @@ build/generated/ksp/main/java/
 ```
 
 and perhaps also resource directory if your IDE supports them:
+
+```
+build/generated/ksp/main/resources/
 ```
-build/generated/ksp/main/resources
+
+It may also be necessary to configure these directories in your KSP consumer module:
```
ting-yuan, line 174: May I know in what case that marking those dirs in IDE doesn't work? E.g., some IDE may not support marking generated sources?

edrd-f: I have this problem with IntelliJ in all my projects using KSP. Marking the directory with generated sources makes code highlighting / symbol resolution work when opening the generated files, however application code referencing generated files still present an error:

bark-ide-issue-1

Adding the directories to the source sets solves it.

I thought this was a common issue, since a few people commented in #37.

ting-yuan: I see. Let's merge this and allow me to investigate on why the IDE marking doesn't work later. Thanks for the PR!

ting-yuan (approved):

docs/quickstart.md
edrd-f: Cool, thank you!

PR #548, opened 2021-08-09
check name instead of psi for PSI synthetic elements

### LINKED ISSUE #240: java.lang.IllegalStateException: Cannot find descriptor for values
When calling `getDeclaredMethods` on an enum class declared in java, calling `getJvmName` on the `values` method throws IllegalStateException.

```java
                public enum JavaEnum {
                    VAL1,
                    VAL2;
                }
```
```
val javaEnum = resolver.getClassDeclarationByName("JavaEnum")
val valuesMethod = javaEnum.getDeclaredMethods().filter {it.name == "values" }
resolver.getJvmName(valuesMethod) // throws
```

ksp version: 20210107

neetopia:
fixes #240

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -395,11 +395,17 @@ class ResolverImpl(
                         }
                 } else null
                 property ?: moduleClassResolver
-                    .resolveContainingClass(psi)
-                    ?.findEnclosedDescriptor(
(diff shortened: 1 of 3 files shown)


neetopia:
for some reason CI is failing, but the failed test passed on my machine with latest main rebase.

ting-yuan (approved):

ting-yuan:
> for some reason CI is failing, but the failed test passed on my machine with latest main rebase.

Looks like it only differs in the order?

neetopia:
> > for some reason CI is failing, but the failed test passed on my machine with latest main rebase.
> 
> Looks like it only differs in the order?

it is. But I can't reproduce that ordering locally, with main rebase. Could possibly be due to JDK, tried with unsetting JAVA_HOME but still getting same result (passing with my change).

PR #553, opened 2021-08-11
add signing plugin to publishing task

neetopia:


ting-yuan (commented):
Can we skip singing when publishing to maven local? Also, only when the key is supplied. For example:
https://github.com/tschuchortdev/kotlin-compile-testing/blob/master/build.gradle#L137

Also, it would be safer / less-error-prone to use the keyring file, instead of storing the key in a gradle property.

neetopia:
> Can we skip singing when publishing to maven local? Also, only when the key is supplied. For example:
> https://github.com/tschuchortdev/kotlin-compile-testing/blob/master/build.gradle#L137
> 
> Also, it would be safer / less-error-prone to use the keyring file, instead of storing the key in a gradle property.

I can add skip mavenLocal. But `signingKey` should be fine since it is just the ID of the key not the actual content.

ting-yuan (approved):

PR #561, opened 2021-08-13
Enable configuration cache in integration tests

ting-yuan:
...for most of the tests. It is still disabled in 2 test:
1. AndroidIT's assembleRelease is still failing and under investigation.
2. K/N compilation doesn't support configuration cache yet.

Real change is in ea681b3, which avoids referencing project in execution phase.

Also upgraded AGP to 7.0.0.


gavra0:
Looks good! Just to double check, are there enough tests that execute build at least twice in order to make sure that state is serialized and then deserialized correctly? I guess all incremental tests are doing that, but just double checking.

gavra0 (approved):

ting-yuan:
Yes, incremental tests are doing that. And I repeated two tests in the last commit. Thanks!

Issue #566, opened 2021-08-15
InvalidUserDataException with native (androidNative?) targets


natario1:
Hi, as discussed in Slack - getting the following exception when using a single androidNative target:

```
Failed building KotlinMPPGradleModel
org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreationException: Could not create task ':sample:kspKotlinAndroidNativeX64'.
Caused by: org.gradle.api.InvalidUserDataException: Cannot change dependencies of dependency configuration ':sample:kotlinCompilerPluginClasspathAndroidNativeX64Main' after it has been resolved.
	at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.preventIllegalMutation(DefaultConfiguration.java:1260)
	at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.validateMutation(DefaultConfiguration.java:1229)
	at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.lambda$validateMutationType$1(DefaultConfiguration.java:308)
	at org.gradle.internal.ImmutableActionSet$SingletonSet.execute(ImmutableActionSet.java:225)
	at org.gradle.api.internal.DefaultDomainObjectSet.assertMutableCollectionContents(DefaultDomainObjectSet.java:71)
	at org.gradle.api.internal.DefaultDomainObjectCollection.add(DefaultDomainObjectCollection.java:252)
	at org.gradle.api.internal.DelegatingDomainObjectSet.add(DelegatingDomainObjectSet.java:110)
	at org.gradle.api.internal.artifacts.DefaultDependencySet.add(DefaultDependencySet.java:68)
	at org.gradle.api.internal.artifacts.DefaultDependencySet.add(DefaultDependencySet.java:36)
	at com.google.devtools.ksp.gradle.KspGradleSubplugin.applyToCompilation$lambda-16$lambda-15(KspSubplugin.kt:343)
```

I set up a project that shows this issue: https://github.com/natario1/KSPBugs , but it basically boils down to:

```kotlin
plugins {
    id("kotlin-multiplatform")
    id("com.google.devtools.ksp")

}

kotlin {
    androidNativeX64()
}

dependencies {
    ksp(project(":processor"))
}
```

Note that other targets like jvm / js work properly. Did not try with native targets other than androidNative*.

ting-yuan:
Per discussion on Slack, this is reproducible in AS. I verified that #568 works. Thanks for reporting the issue and the test case!

PR #568, opened 2021-08-17
Native: add api jar at project evaluation

ting-yuan:
The compiler plugin classpath configuration cannot be changed at
configuration time.


ting-yuan:
Note that I haven't been able to reproduce the error in #566 yet, which motivates this fix.

gavra0 (approved):

PR #581, opened 2021-08-20
handle java array type resolution

neetopia:


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -509,15 +510,27 @@ class ResolverImpl(
         incrementalContext.recordLookup(psi)
         val javaType = JavaTypeImpl.create(psi)
 
-        var parent: PsiElement? = (psi as? PsiClassReferenceType)?.resolve()
+        var parent: PsiElement? = if (psi is PsiArrayType) {
+            (psi.deepComponentType as? PsiClassReferenceType)?.resolve()
+        } else {
+            (psi as? PsiClassReferenceType)?.resolve()
```
ting-yuan, line 516: It looks a bit weird to me, that the innermost scope starts from the resolved type. What about `val x: C<T>` in 
```
class C {
  class D<T> {
    val x: C<T>
  }
}
```
?

Instead of starting from `psi.resolve()`, should the scope construction start from `psi.findParentPsiDeclaration()`? Then we shouldn't need to handle array type and type parameters of the resolved type.

neetopia: `PsiType` does not support a `parent` operation, therefore a resolve is needed to make it to be a type parameter or class.

ting-yuan: Another example is
```
class C<T> {
}

class D<T> {
  val x: C<T> = ...
}
```
and this patch will get wrong `T`.

neetopia: that's why it is needed to visit the type parameters in the next line. added a test case covering this.

ting-yuan: There are still holes. For example, `T2` in `Foo<T1, T2>` will go wrong.
```
class C<T1, T2> {
 class D<T2> {
    val x: Foo<T1, T2> = ...
  }
}
```

neetopia: not sure where does Foo and Bar come from, I modified the test to be 
```
class Foo<T1,T2> {}
class Bar<T1, T2> {}

class Base<T,P> {
    void genericT(List<T> t){};
    void singleT(T t){};
    void varargT(T... t){};
    void arrayT(T[] t){};

    class Inner<P> {
        void genericT(List<T> t){};
        void singleT(T t){};
        void varargT(T... t){};
        void arrayT(T[] t){};
        Foo<T, P> foo;
        Bar<P, T> bar;
    }
}
```
and the newly added `foo` and `bar` got following output
```
Foo<Base.T?, Base.Inner.P?>?
Bar<Base.Inner.P?, Base.T?>?
```

ting-yuan (approved):

PR #597, opened 2021-08-30
Disable isIntermoduleIncremental when isIncremental is false

ting-yuan:
Also label isKspIncremental and isIntermoduleIncremental with @Input

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -286,6 +286,7 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                     }
                 )
             }
+            kspTask.isKspIncremental = isIncremental
         }
 
         fun configureAsAbstractCompile(kspTask: AbstractCompile) {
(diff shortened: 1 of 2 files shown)


ting-yuan:
Running with
1. `ksp.incremental=true` and `ksp.incremental.intermodule=true` and then
2. `ksp.incremental=false` and `ksp.incremental.intermodule=true`
confused kapt's class snapshot checks; The later combination should not bother checking classpath snapshots at all.

@ZacSweers this fixes https://github.com/ZacSweers/MoshiX/runs/3406356413

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -477,7 +478,7 @@ abstract class KspTaskJvm : KotlinCompile(KotlinJvmOptionsImpl()), KspTask {
         )
 
         isIntermoduleIncremental = project.findProperty("ksp.incremental.intermodule")?.toString()?.toBoolean() ?: true
-        if (isIncremental && isIntermoduleIncremental) {
+        if (isKspIncremental && isIntermoduleIncremental) {
```
neetopia, line 481: it looks to me that `isIntermoduleIncremental` is always set to be overridden by `isKspIncremental`, can we merge the check into where we assign the value of `isIntermoduleIncremental`?

neetopia (approved):

neetopia (approved):

PR #604, opened 2021-09-02
handle annotation argument of type kotlin array reflection object

### LINKED ISSUE #172: Composite KClass cast to KSType will be omitted wrapper type
I have some code.
```kotlin
class FooProcessor : SymbolProcessor {

    private lateinit var logger: KSPLogger

    override fun init(options: Map<String, String>, kotlinVersion: KotlinVersion, codeGenerator: CodeGenerator, logger: KSPLogger) {
        this.logger = logger
    }

    override fun process(resolver: Resolver) {
        resolver
            .getSymbolsWithAnnotation(Foo::class.qualifiedName!!)
            .filterIsInstance<KSClassDeclaration>()
            .forEach {classDeclaration ->
                val emailsProperty = classDeclaration.getAllProperties().single { it.simpleName.asString() == "emails" }
                val type = emailsProperty.annotations[0].arguments[0].value as KSType

                logger.warn("emails property type = $type")
               // emailsProperty has declared annotation @FooProperty and @FooProperty type is Array<String>::class
               // actual result: emails property type = String
              // expected result: emails property type = Array
            }
    }

    override fun finish() {}

}

annotation class Foo

annotation class FooProperty(val type: KClass<*>)

@Foo
class Employee(
    @FooProperty(type = Array<String>::class)
    val emails: String?,
)
```

In example. 
```kotlin
@Foo
class Employee(
    @FooProperty(type = Array<String>::class)
    val emails: String?,
)
```
email property has be declared Annotation ``@FooProperty(type = Array<String>::class)``. When ``Array<String>::class)`` cast to KSType. 

I got type value ``String`` and not ``Array``

my expected result is ``Array`` type

neetopia:
fixes #172

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSAnnotationDescriptorImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSAnnotationDescriptorImpl.kt
@@ -113,7 +113,8 @@ private fun <T> ConstantValue<T>.toValue(parent: KSNode): Any? = when (this) {
             it.simpleName.asString() == value.second.asString()
     }?.let { (it as KSClassDeclaration).asStarProjectedType() }
     is KClassValue -> when (val classValue = value) {
-        is KClassValue.Value.NormalClass -> classValue.classId.findKSType()
+        is KClassValue.Value.NormalClass -> if (classValue.arrayDimensions > 0)
+            ResolverImpl.instance.builtIns.arrayType else classValue.classId.findKSType()
         is KClassValue.Value.LocalClass -> getKSTypeCached(classValue.type)
     }
     is ErrorValue, is NullValue -> null
--- a/compiler-plugin/testData/api/annotationValue.kt
+++ b/compiler-plugin/testData/api/annotationValue.kt
@@ -23,6 +23,7 @@
 // Foo
 // File
 // Local
+// Array
 // @Foo
 // @Suppress
 // G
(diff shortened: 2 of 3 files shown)


neetopia:
and this can only happen for `kotlin.Array` since only `kotlin.Array<T>` is a valid input with type argument for `::class`

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSAnnotationDescriptorImpl.kt
```diff
@@ -113,7 +113,8 @@ private fun <T> ConstantValue<T>.toValue(parent: KSNode): Any? = when (this) {
             it.simpleName.asString() == value.second.asString()
     }?.let { (it as KSClassDeclaration).asStarProjectedType() }
     is KClassValue -> when (val classValue = value) {
-        is KClassValue.Value.NormalClass -> classValue.classId.findKSType()
+        is KClassValue.Value.NormalClass -> if (classValue.arrayDimensions > 0)
+            ResolverImpl.instance.builtIns.arrayType else classValue.classId.findKSType()
```
ting-yuan, line 117: `builtIns.arrayType` is `Array<*>`. We probably need `arrayType.replace(listOf(ResolverImpl.instance.getTypeArgument(classValue.classId.findKSType(), Variance.INVARIANT)))`.

neetopia: It is not possible to do `Array<T>.class` in Java, and it is only possible in Kotlin with `Array` as the type, therefore I decided to put just `Array` here for consistence.

ting-yuan (approved):

Issue #603, opened 2021-09-02
Validate annotation arguments too in KSP's default validator.


danysantiago:
Similar to auto-common's superficial validator it would be nice if KSP would also validate annotation arguments, specifically the arguments of types. Right now, only the annotation type itself is validated, here: https://github.com/google/ksp/blob/main/api/src/main/kotlin/com/google/devtools/ksp/visitor/KSValidateVisitor.kt#L44

I expect a `KSClassDeclaration.validate()` call to return `false` for a source like this:
```
@Entity(
    foreignKeys = [ 
        @ForeignKey(
            entity = NotYetGeneratedType.class,
            parentColumns = "id",
            childColumns = "id"
    ]
)
data class MyEntity(@PrimaryKey id: Long)
```

Issue #538, opened 2021-07-30
KSAnnotated#arguments causes an NPE


AlexBeggs:
The KSAnnotation.arguments throws an NPE when one of the arguments is a KClass.
If a default is defined but doesn't define one in the declaration. 

See branch code example

The call that throws an NPE

https://github.com/AlexBeggs/ksp/commit/bd4b66b35cae566d44b7449296b2b22126978940#diff-99a569b02956aecc148c7b198d7b3d5b083f664c795bd6e35deb141c067b041bR31

The definition of the annotation with the KClass

https://github.com/AlexBeggs/ksp/commit/bd4b66b35cae566d44b7449296b2b22126978940#diff-9fd74f2bedbcd4b782c13de9fe99cc8b1bda160f1461aab4713c315b9c81e1b2R33

`
error: [ksp] java.lang.NullPointerException
	at org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluatorVisitor.visitClassLiteralExpression(ConstantExpressionEvaluator.kt:937)
	at org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluatorVisitor.visitClassLiteralExpression(ConstantExpressionEvaluator.kt:373)
	at org.jetbrains.kotlin.psi.KtClassLiteralExpression.accept(KtClassLiteralExpression.kt:23)
	at org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluatorVisitor.evaluate(ConstantExpressionEvaluator.kt:389)
	at org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluator.evaluateExpression(ConstantExpressionEvaluator.kt:288)
	at org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluator.evaluateExpression$default(ConstantExpressionEvaluator.kt:281)
	at org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluator.evaluateToConstantValue(ConstantExpressionEvaluator.kt:300)
	at com.google.devtools.ksp.processing.impl.ResolverImpl.evaluateConstant(ResolverImpl.kt:370)
	at com.google.devtools.ksp.symbol.impl.binary.KSAnnotationDescriptorImplKt.getDefaultValue(KSAnnotationDescriptorImpl.kt:258)
[...]

PR #605, opened 2021-09-03
Add annotation argument validation to validate API

danysantiago:
Also updated getting default value to be a `KSErrorType` when the `ValueDescriptor` has an error type, this is because without an expected type it seems KSP can't resolve the constant default value.

Fixes: https://github.com/google/ksp/issues/603


ting-yuan:
The CI failure is due to OOM, which fixed in 33db4b98a1f6a4b1b47097520251e33125bd015b. Could you rebase and upload again?

danysantiago:
Rebased, thanks.

neetopia:
Overall looks good, will merge once #538 is fixed

danysantiago:
Now that https://github.com/google/ksp/pull/508 is merged can you please take a look at this change? Thanks!

neetopia:
the dependent issue here is on #538 , which is not fixed by #508 . I have an idea what went wrong for #538 but currently have no bandwidth to fix it..

PR #606, opened 2021-09-03
Allowing getting sources from compile tasks

ting-yuan:
instead of default source set and its dependencies.

There are 3 ways (that I know of) to register extra sources to
compilation tasks:

  1. Add to default source sets.
  2. Add to dependencies of default source sets.
  3. Add to compile task directly, e.g., via AbstractCompile.source

KSP picked up from the first 2 only. Now it also does third.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -297,7 +297,13 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                 classOutputDir,
                 resourceOutputDir
             )
-            kotlinCompilation.allKotlinSourceSets.forEach { sourceSet -> kspTask.source(sourceSet.kotlin) }
+            kspTask.source(kotlinCompileTask.source)
+            if (kotlinCompileTask is AbstractKotlinCompile<*>) {
+                val sourceRoots = kotlinCompileTask.getSourceRoots()
+                if (sourceRoots is SourceRoots.ForJvm) {
+                    kspTask.source(sourceRoots.javaSourceRoots)
+                }
+            }
 
             // Don't support binary generation for K/N yet.
             // FIXME: figure out how to add user generated libraries.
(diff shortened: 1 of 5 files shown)


ting-yuan:
This is consistent with kaptGenerateStubsTask by the way.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -297,7 +297,13 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                 classOutputDir,
                 resourceOutputDir
             )
-            kotlinCompilation.allKotlinSourceSets.forEach { sourceSet -> kspTask.source(sourceSet.kotlin) }
+            kspTask.source(kotlinCompileTask.source)
```
gavra0, line 300: This may be tricky as kotlin compile task will have java/kotlin sources that are generated by ksp processors. In `kaptGenerateStubsTask` this is avoided by filtering out sources that are in the task output directories.

What is the motivation for this change, typically sourceSets should contain all information, and plugins should avoid adding sources directly to the task?

ting-yuan: I was thinking about minimizing surprises for users moving from kapt. This particular case / bug report is about sources added via `BaseVariant.registerJavaGeneratingTask` by other plugins.

Having said that, this behavior isn't quite right. We should probably not picking up generated sources at all.

neetopia (approved):

ting-yuan:
An option is added to switch between the behaviors. Default is the old behavior.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
gavra0: With the toggle, this LGTM. We can also follow up with users that do enable it, to understand their use-cases a bit better.

gavra0 (approved):

PR #614, opened 2021-09-08
Add support for DeeplinkDispatch

StefMa:


neetopia (approved):
Thanks for supporting KSP!

PR #615, opened 2021-09-08
Add Lich SavedState to supported libraries.

yamasa:
https://github.com/line/lich/tree/master/savedstate#lich-savedstate
Lich SavedState is a library that provides type-safe access to saved instance state.
Now, it officially supports KSP to generate arguments classes for the saved instance state.


yamasa:
@googlebot I signed it!

neetopia (approved):
Thanks for supporting KSP!

Issue #625, opened 2021-09-14
Classpath property is sensitive to jar timestamps, breaking remote cacheability


BenSchwab:
Follow up to:
https://github.com/google/ksp/issues/490

In #543 `@get:CompileClasspath` was removed in favor of `@get:PathSensitive(PathSensitivity.RELATIVE)`. 

If there is a locally compiled jar on the classpath, the timestamp will be different then the jar in the remote gradle cache, which will break the remote cacheability of the task.

At the time of filing the original issue, the file collection had no input markers, and I incorrectly suggested Relative path sensitivity. I believe that the correct marker is indeed classpath, as according to the comment it also factors in path differences:

```
/**
 * <p>Marks a property as specifying a Java compile classpath for a task. Attaching this annotation to a property means that changes that do not affect the API of the classes in classpath will be ignored. The following kinds of changes to the classpath will be ignored:</p>
 *
 * <ul>
 *     <li>Changes to the path of jar or top level directories.</li>
 *     <li>Changes to timestamps and the order of entries in Jars.</li>
 *     <li>Changes to resources and Jar manifests, including adding or removing resources.</li>
 *     <li>Changes to private class elements, such as private fields, methods and inner classes.</li>
 *     <li>Changes to code, such as method bodies, static initializers and field initializers (except for constants).</li>
 *     <li>Changes to debug information, for example when a change to a comment affects the line numbers in class debug information.</li>
 *     <li>Changes to directories, including directory entries in Jars.</li>
 * </ul>
```

ting-yuan:
> At the time of filing the original issue, the file collection had no input markers,

May I know what you mean by *the file collection had no input markers*?

`CompileClasspath` was used before that commit:
https://github.com/google/ksp/pull/543/files

According to Gradle [documentation](https://docs.gradle.org/current/userguide/more_about_tasks.html), the 3 markers seem to have the following relations in terms of "cache-ability": `CompileClasspath` implies `Classpath` implies `PathSensitive(PathSensitivity.RELATIVE)`.

I understand that `@CompileClasspath` can cause wrong optimizations / caches and it should be `Classpath` or `PathSensitive(PathSensitivity.RELATIVE)` for correctness. What I'm not sure is, if `CompileClasspath` doesn't result in cache hit, why would `Classpath` or `PathSensitive(PathSensitivity.RELATIVE)` do?

ting-yuan:
A tentative fix is here, but I need to figure out why `CompileClasspath` caused cache miss before.
[...]

PR #626, opened 2021-09-15
gradle-plugin: Update overly conservative input annotations

ting-yuan:
Classpath implies PathSensitive(PathSensitivity.RELATIVE)


neetopia (approved):

gavra0:
This looks good, but can you also add a test that:
- uses a custom location for the [build cache](https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_configure)
- creates a project, runs full build
- copies that project to another location and runs full build

In the second build, the ksp task should not run and it should be `from cache`.

ting-yuan:
@gavra0 I added a test case, which creates 2 projects that share a single build cache dir.

Just in case, I also removed `@get:InputFiles` per #625. There should be no difference besides [compatibility](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Classpath.html) to early Gradle versions (< 3.2). Since kapt also doesn't use `@InputFiles`, it should be safe?

PR #631, opened 2021-09-17
add prebuilt jar for fir high level api, add module for ksp-fir-impl

neetopia:


ksp-fir-impl/build.gradle.kts
```diff
@@ -0,0 +1,67 @@
+description = "Kotlin Symbol Processor FIR implementation"
```
ting-yuan, line 1: Can we use "Analysis API", as they are used by upstream?

ksp-fir-impl/build.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+
+dependencies {
+    implementation(kotlin("stdlib", kotlinBaseVersion))
+    implementation("org.jetbrains.kotlin:high-level-api-for-ide:1.6.255")
+    implementation("org.jetbrains.kotlin:kotlin-compiler:$kotlinBaseVersion")
+
+    implementation(project(":api"))
+
+    // workaround: IntelliJ doesn't resolve packed classes from included builds.
+    if (kotlinProjectPath != null) {
```
ting-yuan, line 38: Do you want composite build? I haven't used that mode for more than one year, and I feel that it is likely broken now.

settings.gradle.kts
```diff
@@ -19,6 +19,7 @@ include("gradle-plugin")
 include("compiler-plugin")
 include("symbol-processing")
 include("integration-tests")
+include("ksp-fir-impl")
```
ting-yuan, line 22: Can we use something like "kotlin-analysis-api"

ting-yuan (approved):
LGTM, just a few nit-picks

PR #638, opened 2021-09-25
fix getSymbolsWithAnnotation for @setparam and @param use site target

### LINKED ISSUE #636: Getting annotated property setter parameter from Resolver return nothing
For the following source:
```
class Foo {
  @setparam:MyAnnotation
  var property: String = ""
}
```
Calling `resolver.getSymbolsWithAnnotation("MyAnnotation")` returns nothing, where as I would expect it to return a `KSValueParameter` corresponding to the property setter parameter.

neetopia:
fixes #636 

also handled the annotation use site target for constructor parameter. note that with such handling logic, 2 symbols will be returned (one for property declaration and one for value parameter) for a property declared in primary constructor, unless `@param` is specified.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -307,6 +307,7 @@ class ResolverImpl(
             }
 
             override fun visitPropertySetter(setter: KSPropertySetter, data: Unit) {
+                setter.parameter.accept(this, data)
                 visitAnnotated(setter, data)
             }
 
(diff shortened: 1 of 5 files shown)


ting-yuan:
> note that with such handling logic, 2 symbols will be returned (one for property declaration and one for value parameter) for a property declared in primary constructor, unless `@param` is specified.

What's the behavior of non-use-site-target val param in explicit constructors? Same?

neetopia:
> > note that with such handling logic, 2 symbols will be returned (one for property declaration and one for value parameter) for a property declared in primary constructor, unless `@param` is specified.
> 
> What's the behavior of non-use-site-target val param in explicit constructors? Same?

you mean secondary constructors with `val`s in parameter list? That is not allowed.

ting-yuan:
For example, does it return 2 symbols as well for `class C(@Deprecated("msg") val x: Int)`?

neetopia:
I see what you mean, in that case, yes. I'll add a test case for that.

ting-yuan (approved):

PR #640, opened 2021-09-27
Add gRPC Dekorator to supported libraries

mottljan:
[gRPC Dekorator](https://github.com/mottljan/grpc-dekorator) is a library that you can use to decorate your RPCs to add them some special functionality which is not possible using gRPC's ClientInterceptors. It generates decorator classes using KSP.

I would also like to appreciate and thank you for your amazing job on KSP!


mottljan:
@googlebot I fixed it.

PR #648, opened 2021-10-04
ignore all dist folders

neetopia:


ting-yuan:
Can we enumerate all of them instead of using **?

ting-yuan (approved):

PR #650, opened 2021-10-05
Re-include the `ksp` dependency for all targets

ting-yuan:
So that the behavior aligns with KSP 1.0.0. Note that the apply-to-all
`ksp` configuration is being deprecated in the following way:

  1. A seperate `kspTest` configuration is available for test source
     sets now. In the future `ksp` won't apply to test source sets.

  2. `ksp` will not be available at all for multiplatform projects.

The `ksp.allow_all_target_configuration` gradle property is also
introduced, so that users can early-access the new behavior by setting
it to false.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
@@ -16,7 +16,7 @@ class KspConfigurations(private val project: Project) {
     }
 
     private val allowAllTargetConfiguration =
-        project.findProperty("ksp.allow_all_target_configuration")?.let {
+        project.findProperty("ksp.allow.all.target.configuration")?.let {
             it.toString().toBoolean()
         } ?: true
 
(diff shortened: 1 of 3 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
```diff
@@ -15,6 +15,14 @@ class KspConfigurations(private val project: Project) {
         private const val PREFIX = "ksp"
     }
 
+    private val allowAllTargetConfiguration =
+        project.findProperty("ksp.allow_all_target_configuration")?.let {
```
gavra0, line 19: nit: Maybe `ksp.allow.all.target.configuration` or `ksp.allowAllTargetConfiguration` to match Gradle/Kotlin/AGP naming.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
```diff
@@ -66,13 +74,17 @@ class KspConfigurations(private val project: Project) {
             is KotlinMultiplatformExtension -> {
                 kotlin.targets.configureEach(::decorateKotlinTarget)
 
-                // Adding multiplatform configuration removed support for the root ksp configuration.
-                // Try to make this breaking change less breaking by adding a clear error.
-                project.configurations.create("ksp").dependencies.all {
-                    throw InvalidUserCodeException(
-                        "The 'ksp' configuration cannot be used in Kotlin Multiplatform projects. " +
+                project.afterEvaluate {
```
gavra0, line 77: Avoid usage of `afterEvaluate` as the lifecycle is tricky, to say the least. Instead, you can have `configuartion.dependencies.whenObjectAdded { reportIssueOnce(...) }`.

PR #667, opened 2021-10-12
Import contents of playground.zip into examples/

ting-yuan:


examples/playground/.gitignore
```diff
@@ -0,0 +1,5 @@
+.gradle/
```
neetopia, line 1: is the `.gitignore` under playground necessary?

neetopia (approved):

PR #684, opened 2021-10-15
Fix ClassNotFoundException for initializing annotation arguments in a…

bennyhuo:
Fix ClassNotFoundException for initializing annotation arguments in an annotation. The classloader may not be the same with the 'this' class.

In my case: 

```kotlin
@Target(AnnotationTarget.ANNOTATION_CLASS, AnnotationTarget.CLASS)
@Retention(AnnotationRetention.BINARY)
annotation class Builder(
    ...
    /**
     * For Activities Only
     */
    val pendingTransition: PendingTransition = PendingTransition(),
    ...
)
```

I use `val builder = ksClassDeclaration.getAnnotationByType(Builder::class).first()` to get the instance of Builder, but get '[ksp] java.lang.IllegalArgumentException: interface PendingTransition is not visible from class loader'.

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
@@ -402,7 +402,7 @@ private fun KSAnnotation.asAnnotation(
     annotationInterface: Class<*>,
 ): Any {
     return Proxy.newProxyInstance(
-        this.javaClass.classLoader, arrayOf(annotationInterface),
+        annotationInterface.classLoader, arrayOf(annotationInterface),
         this.createInvocationHandler(annotationInterface)
     ) as Proxy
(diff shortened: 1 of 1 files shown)


bennyhuo:
I have also checked the test cases but it seems not easy to load an annotation class  in the cases.

ting-yuan:
The second commit 71fc14bc4e6841281936ecc2cea65e7a594855ca reverts the change in the first commit 3552e4c00d584b3f9fda85eb3d9160887c33432c. The net change of this PR looks good to me. I'm assuming that the first commit was needed for a test case (for the second commit) that was abandoned. If I guess it right, I'll rename this PR and squash the first commit into the second before merging.

Or is the ClassNotFoundException a real issue and you are still working it?

PR #685, opened 2021-10-15
Fix enum lookup in annotation proxies

ZacSweers:
Before, this would take the `toString()` of the KSType, but that's wrong because then you have something like `AnnotationRetention.valueOf("kotlin.annotation.AnnotationRetention.RUNTIME")`, which obviously won't work. This unpacks the short name from the `KSType` directly so that it actually looks up just `RUNTIME` in the given example.

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/utils.kt
@@ -451,7 +451,12 @@ private fun List<*>.toArray(method: Method, valueProvider: (Any) -> Any): Array<
 
 @Suppress("UNCHECKED_CAST")
 private fun <T> Any.asEnum(returnType: Class<T>): T =
-    returnType.getDeclaredMethod("valueOf", String::class.java).invoke(null, this.toString()) as T
+  returnType.getDeclaredMethod("valueOf", String::class.java)
+    .invoke(
+      null,
+      // Change from upstream KSP - pull out the simple name to safely look it up
+      (this as KSType).declaration.simpleName.getShortName()
+    ) as T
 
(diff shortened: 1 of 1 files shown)


ZacSweers:
Weirdly I can't repro this issue, but it does repro if you use the current KSP 1.5.31 impl in this PR https://github.com/square/moshi/pull/1393

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -451,7 +451,15 @@ private fun List<*>.toArray(method: Method, valueProvider: (Any) -> Any): Array<
 
 @Suppress("UNCHECKED_CAST")
 private fun <T> Any.asEnum(returnType: Class<T>): T =
-    returnType.getDeclaredMethod("valueOf", String::class.java).invoke(null, this.toString()) as T
+    returnType.getDeclaredMethod("valueOf", String::class.java)
+        .invoke(
+            null,
+            if (this is KSType) {
```
ting-yuan, line 457: (Related discussion; Not for this PR)

Without the type checking it would result in the following error, where the type of `this` is `TestEnum`, rather than a `KSType`.
https://github.com/google/ksp/runs/3911397835#step:9:729

@neetopia Does the need of type checking here indicate a discrepancy in implementations of annotation value? In other words, shouldn't it always be `KSType`?

ting-yuan: For example, [this](https://github.com/google/ksp/commit/25fb02ab28a074641c5b6ef873f59181173dfd7a) was one of the difference among the implementations of .class and sources.

It could also be the reason why it's only reproducible with KSP 1.5.31-1.0.0; The `toString()`s of class instance and `KSType` are different.

PR #686, opened 2021-10-16
Add CodeGenerator.associateWithClasses

ting-yuan:
So that changes in unreachable classes can invalidate corresponding
outputs and trigger re-processing correctly.

The implementation creates an internal, virtual NoSourceFile for each
class to propagate dirtyness.


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/CodeGeneratorImpl.kt
```diff
@@ -96,6 +98,19 @@ class CodeGeneratorImpl(
...
 
+    override fun associateWithClasses(
+        classes: List<KSClassDeclaration>,
+        packageName: String,
+        fileName: String,
+        extensionName: String
+    ) {
+        val path = pathOf(packageName, fileName, extensionName)
+        val files = classes.map {
+            it.containingFile ?: NoSourceFile(projectBase, it.qualifiedName?.asString().toString())
```
neetopia, line 109: should the `NoSourceFile` be cached based on qualified name?

neetopia (approved):
LGTM, with one minor comment

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/CodeGeneratorImpl.kt
ting-yuan: It's only used to calculate the path name and will be GC'd.

PR #693, opened 2021-10-21
Mute a warning from ScriptingGradleSubplugin

ting-yuan:
which tries to get `sourceSetName` before this task is configured.


ting-yuan:
This is the problematic line:
https://github.com/JetBrains/kotlin/blob/01dd15cc3ed59f7080bd1e6fcd8ff8cc81d4c588/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/scripting/internal/ScriptingGradleSubplugin.kt#L81

PR #705, opened 2021-10-29
update documentation for getSymbolsAnnotatedWith behavior for multiple round and deferred symbols

neetopia:


neetopia:
cc @elihart

api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessor.kt
```diff
@@ -33,7 +33,7 @@ interface SymbolProcessor {
      * Called by Kotlin Symbol Processing to run the processing task.
      *
      * @param resolver provides [SymbolProcessor] with access to compiler details such as Symbols.
-     * @return A list of deferred symbols that the processor can't process.
+     * @return A list of deferred symbols that the processor can't process. Only symbols that can't be processed at this round should be returned, and symbols from libraries will be forfeited by KSP.
```
ting-yuan, line 36: replace
>, and symbols from libraries will be forfeited by KSP.

with
>. Symbols in compiled code (libraries) are always valid and are ignored if returned in the deferral list.

ting-yuan (commented):
Can you also comment and document that `getNewFiles` won't return a file even if a symbol in it is deferred?

btw, this needs to be updated, too:
https://github.com/google/ksp/blob/main/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt#L36

neetopia:
> Can you also comment and document that `getNewFiles` won't return a file even if a symbol in it is deferred?
> 
> btw, this needs to be updated, too: https://github.com/google/ksp/blob/main/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt#L36

updated `getAllFiles()`, however I feel `getNewFiles()` is good enough for now, since it explicitly says `new files generated from last last round`

ting-yuan (approved):
Yes that's what I mean. Thanks!

PR #706, opened 2021-11-01
chore: update lyricist support status

adrielcafe:
Update Lyricist library status from experimental to officially supported.


neetopia (approved):
Thanks for supporting KSP!

PR #712, opened 2021-11-05
Introduce APIs for jvm modifiers

ting-yuan:
Specifically,
```
ResolverImpl.isJvmTransient(property: KSPropertyDeclaration)
ResolverImpl.isJvmVolatile(property: KSPropertyDeclaration)
ResolverImpl.isJvmStrictfp(function: KSFunctionDeclaration)
ResolverImpl.isJvmSynchronized(function: KSFunctionDeclaration)
```

Those modifiers / access flags are part of Kotlin compilers implementation on the JVM
platform. They don't exist directly in the Kotlin language. Therefore,
those APIs are introduced so that processors don't need to guess by
themselves.


ting-yuan:
+@ZacSweers, this is a tentative solution.

ting-yuan:
Maybe we can use an unified `ResolverImpl.effectiveJavaModifiers`. Will update later.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1089,6 +1095,82 @@ class ResolverImpl(
...
+            Origin.JAVA -> {
+                addVisibilityModifiers()
+                if (declaration is KSClassDeclaration && declaration.classKind == ClassKind.INTERFACE)
+                    modifiers.add(Modifier.ABSTRACT)
+            }
+            Origin.KOTLIN -> {
+                addVisibilityModifiers()
+                if (!declaration.isOpen())
+                    modifiers.add(Modifier.FINAL)
+                if (declaration.hasAnnotation("kotlin.jvm.JvmStatic"))
```
neetopia, line 1123: minor: maybe extract consts.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1089,6 +1095,82 @@ class ResolverImpl(
...
+                if (declaration.hasAnnotation("kotlin.jvm.Transient"))
+                    modifiers.add(Modifier.JAVA_TRANSIENT)
+                if (declaration.hasAnnotation("kotlin.jvm.Volatile"))
+                    modifiers.add(Modifier.JAVA_VOLATILE)
+                when (declaration) {
+                    is KSClassDeclaration -> {
+                        if (declaration.isCompanionObject)
+                            modifiers.add(Modifier.JAVA_STATIC)
+                        if (declaration.classKind == ClassKind.INTERFACE)
+                            modifiers.add(Modifier.ABSTRACT)
```
neetopia, line 1142: Nit: feeling kind of tricky for this one. You can actually write `abstract interface` but it does not have any effects.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -734,3 +782,42 @@ internal val KSDeclarationContainer.declarationsInSourceOrder: Sequence<KSDeclar
...
     }
+
+internal val KSPropertyDeclaration.jvmAccessFlag: Int
+    get() = when (origin) {
+        Origin.KOTLIN_LIB -> {
+            val descriptor = (this as KSPropertyDeclarationDescriptorImpl).descriptor
+            val kotlinBinaryJavaClass = descriptor.getContainingKotlinJvmBinaryClass()
+            kotlinBinaryJavaClass?.let {
+                BinaryClassInfoCache.getCached(it).fieldAccFlags.get(this.simpleName.asString()) ?: 0
+            } ?: 0
```
neetopia, line 793: not sure if return 0 is best here, I feel if the containing class can't be obtained, it rather implies an error?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
ting-yuan: This follows JVM [spec](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.1). Java interfaces have `abstract` **effectively**.

> If the ACC_INTERFACE flag is set, the ACC_ABSTRACT flag must also be set, and the ACC_FINAL, ACC_SUPER, and ACC_ENUM flags set must not be set.

neetopia: That's what I mean, interface is `abstract` by itself, adding a modifier seems a little bit unnecessary to me. Either way works for me.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
ting-yuan: `getContainingKotlinJvmBinaryClass()` has null checks everywhere and I'm not so confident with it. For example, I got a null for some `<init>` of `Companion`. Also there can be some builtin functions, e.g., `Boolean.equals` that doesn't have sources.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
ting-yuan: Done.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
ting-yuan: Added a few comments in code.

neetopia (approved):

PR #718, opened 2021-11-11
1.0.1 release

neetopia:
unit test is not possible due to the bytecode visitor version conflict, integration tested locally passed.


ting-yuan (approved):
Tests all failed because of the asm version issue :(

Can you test it locally with a custom build of Kotlin 1.5.31 without upgrading to 9.0 of the asm library before merging?

PR #726, opened 2021-11-16
update CI and auto-merge to 1.0.2-release branch

neetopia:


.github/workflows/main.yml
```diff
@@ -4,9 +4,9 @@ name: CI
 
 on:
   push:
-    branches: [ main, 1.0.1-release, 1.0.0-release ]
+    branches: [ main, 1.0.2-release ]
```
ting-yuan, line 7: let's keep `1.0.1-release` so that, if we need to release it with next Kotlin release, it is tested.

ting-yuan (approved):

PR #732, opened 2021-11-19
Introduce Resolver.getJavaWildcard

ting-yuan:
which computes the corresponding Java wildcard, from the given
reference.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -44,6 +44,7 @@ import org.jetbrains.kotlin.builtins.jvm.JavaToKotlinClassMap
 import org.jetbrains.kotlin.codegen.ClassBuilderMode
 import org.jetbrains.kotlin.codegen.OwnerKind
 import org.jetbrains.kotlin.codegen.state.KotlinTypeMapper
+import org.jetbrains.kotlin.codegen.state.updateArgumentModeFromAnnotations
 import org.jetbrains.kotlin.config.LanguageVersionSettingsImpl
 import org.jetbrains.kotlin.container.ComponentProvider
 import org.jetbrains.kotlin.container.get
@@ -74,8 +75,11 @@ import org.jetbrains.kotlin.load.java.structure.impl.JavaTypeParameterImpl
 import org.jetbrains.kotlin.load.java.structure.impl.classFiles.BinaryJavaClass
 import org.jetbrains.kotlin.load.java.structure.impl.classFiles.BinaryJavaMethod
 import org.jetbrains.kotlin.load.java.structure.impl.classFiles.BinaryJavaMethodBase
+import org.jetbrains.kotlin.load.kotlin.TypeMappingMode
 import org.jetbrains.kotlin.load.kotlin.VirtualFileKotlinClass
 import org.jetbrains.kotlin.load.kotlin.getContainingKotlinJvmBinaryClass
+import org.jetbrains.kotlin.load.kotlin.getOptimalModeForReturnType
+import org.jetbrains.kotlin.load.kotlin.getOptimalModeForValueParameter
 import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.FqNameUnsafe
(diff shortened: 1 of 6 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1176,6 +1182,107 @@ class ResolverImpl(
...
+        // is KSValueParameter -> RefPosition.PARAMETER_TYPE
+        // is KSTypeParameter -> RefPosition.PARAMETER_TYPE
+        else -> RefPosition.PARAMETER_TYPE
+    }
+
+    private fun KSTypeReference.isReturnTypeOfAnnotationMethod(): Boolean {
+        var candidate = this.parent
+        while (this !is KSClassDeclaration && candidate != null)
+            candidate = candidate.parent
+        return candidate.safeAs<KSClassDeclaration>()?.classKind == ClassKind.ANNOTATION_CLASS
```
neetopia, line 1220: it looks like this is about a type reference inside an annotation class declaration, should it be named annotation method in that sense? I thought annotation method should means a use-site annotation rather than declaration site.

ting-yuan:
[This](https://github.com/JetBrains/kotlin/blob/master/compiler/backend/src/org/jetbrains/kotlin/codegen/state/KotlinTypeMapper.kt#L1495) is an example of how `getVarianceForWildcard` is used when Kotlin compiler maps Kotlin types to Java types for codegen. Note that
* type mapping happens recursively.
* updateArgumentModeFromAnnotations takes care of `@JvmWildcard` and `@JvmSuppressWildcard`.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1176,6 +1182,107 @@ class ResolverImpl(
...
+        is KSPropertyDeclaration -> when (ref) {
+            parent.type -> RefPosition.RETURN_TYPE
+            else -> RefPosition.PARAMETER_TYPE
+        }
+        is KSClassDeclaration -> RefPosition.SUPER_TYPE
+        // is KSTypeArgument -> RefPosition.PARAMETER_TYPE
+        // is KSAnnotation -> RefPosition.PARAMETER_TYPE
+        // is KSTypeAlias -> RefPosition.PARAMETER_TYPE
+        // is KSValueParameter -> RefPosition.PARAMETER_TYPE
+        // is KSTypeParameter -> RefPosition.PARAMETER_TYPE
```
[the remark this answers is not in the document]
ting-yuan: Those are left intentionally to match the above TODO. They are commented out for code size and performance.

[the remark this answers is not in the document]
ting-yuan: In other words, they are documents.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
ting-yuan: Yes, this is checking whether it is a return type / property type in a annotation class declaration. Do you mean change the name or it should check use-site instead?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1176,6 +1182,107 @@ class ResolverImpl(
...
+
+    // TODO: Strict mode for catching unhandled cases.
+    private fun findRefPosition(ref: KSTypeReference): RefPosition = when (val parent = ref.parent) {
+        is KSCallableReference -> when (ref) {
+            parent.returnType -> RefPosition.RETURN_TYPE
+            else -> RefPosition.PARAMETER_TYPE
+        }
+        is KSFunctionDeclaration -> when (ref) {
+            parent.returnType -> RefPosition.RETURN_TYPE
+            else -> RefPosition.PARAMETER_TYPE
```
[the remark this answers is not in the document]
ting-yuan: Yes. Only RETURN_TYPE and SUPER_TYPE are treated differently; They are the only cases where declaration site annotations should be ignored (and `@JvmSuppressWildcard(true)`). I can change PARAMETER_TYPE to DEFAULT if that's more clear.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1176,6 +1182,107 @@ class ResolverImpl(
...
+        // is KSAnnotation -> RefPosition.PARAMETER_TYPE
+        // is KSTypeAlias -> RefPosition.PARAMETER_TYPE
+        // is KSValueParameter -> RefPosition.PARAMETER_TYPE
+        // is KSTypeParameter -> RefPosition.PARAMETER_TYPE
+        else -> RefPosition.PARAMETER_TYPE
+    }
+
+    private fun KSTypeReference.isReturnTypeOfAnnotationMethod(): Boolean {
+        var candidate = this.parent
+        while (this !is KSClassDeclaration && candidate != null)
```
neetopia, line 1218: this loop condition seems to be always making`candidate` to be `null` when `this` is not `KSClassDeclaration`, is that intended?

ting-yuan: No, this is a bug. Thanks for catching that!

neetopia (approved):

PR #734, opened 2021-11-19
Update Kotshi in the readme

ansman:
As of 2.7.0 Kotshi has official KSP support


neetopia (approved):
Thanks for supporting KSP!

PR #742, opened 2021-11-25
update to 1.6.20-dev-7317

neetopia:
upstream kotlin compiler moved `buildCompilerArgs` and `buildCommonArgs` to `private` for native compile task, need to figure out a way to pass plugin classpath and options to compile task.

New gradle plugin behavior introduced with this upstream merge, KSP needs to adjust accordingly as well, new behavior is to follow the new Kotlin/Metadata task behavior where build tasks are named after source set names.


neetopia:
failed macos CI with `Cannot resolve external dependency org.jetbrains.kotlin:kotlin-klib-commonizer-embeddable:1.6.20-dev-7317 because no repositories are defined.`, passed locally though.

ting-yuan:
Have you tried moving back to macOS-10.15, downgrade/upgrade Gradle version or run CI with a trivial change?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
```
ting-yuan, line 3: seems an unnecessary change; might have something to do with the CI failure

ting-yuan (approved):

PR #747, opened 2021-12-01
Analysis API implementation

neetopia:
* All KSSymbols skeleton.
* use same bootstrap version for compiler and analysis api.
* implemented some no resolution involved KSSymbol functions, many are still missing, e.g. modifiers conversion (this should be able to be reused from current frontend implementation).
* implemented simple resolution logic as analysis api resolution concept proofing. 
* Note that analysis API unifies symbols implementation, so there is no need to have 3 implementations for java symbols, kotlin symbols, and descriptor symbols anymore, however, for certain functions, we still need to type check the wrapped element inside a KtSymbol and performance operations accordingly.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
@@ -1,0 +1,38 @@
+package com.google.devtools.ksp.impl.symbol.kotlin
+
+import com.google.devtools.ksp.symbol.KSAnnotation
+import com.google.devtools.ksp.symbol.KSNode
+import com.google.devtools.ksp.symbol.KSReferenceElement
+import com.google.devtools.ksp.symbol.KSType
+import com.google.devtools.ksp.symbol.KSTypeReference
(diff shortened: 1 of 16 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.google.devtools.ksp.symbol.Modifier
+import com.google.devtools.ksp.symbol.Origin
+import org.jetbrains.kotlin.analysis.api.symbols.markers.KtTypeAndAnnotations
+
+class KSTypeReferenceImpl(private val ktTypeAndAnnotations: KtTypeAndAnnotations) : KSTypeReference {
+    override val element: KSReferenceElement?
+        get() = TODO("Not yet implemented")
+
+    override fun resolve(): KSType {
+        return KSTypeImpl(ktTypeAndAnnotations.type)
```
ting-yuan, line 19: This reminds me of the lazy impl in the LL API. We need to keep an eye on the underlying implementation and performance.

ting-yuan (approved):
LGTM. I left a few comments regarding with performance as reminders. No need to address them at the moment. Let's focus on correctness for now and revisit performance in a later iteration.

PR #750, opened 2021-12-03
add JS/IR to KMP integration test configuration

neetopia:


neetopia:
for some reason `JS(IR)` config is working on my machine, make it a draft instead.

neetopia:
close as no longer needed.

Issue #744, opened 2021-11-28
kspJs failing with "Collection has more than one element." when processing code in commonMain


janvladimirmostert:
I'm testing out KSP on a multiplatform setup, but it's failing when I add `kspJs`

kspJvm works fine, as soon as I add kspJs, it fails with  "Collection has more than one element."

```
dependencies {
	add("kspJvm", "io.jvaas:jvaas-observe")
	add("kspJs", "io.jvaas:jvaas-observe")
}
``` 

> 
> Execution failed for task ':applications:od-server:compileProductionExecutableKotlinJs'.
> > Failed to calculate the value of task ':applications:od-server:compileProductionExecutableKotlinJs' property 'entryModule$kotlin_gradle_plugin'.
>    > Collection has more than one element.

I've added detailed code-snippets on StackOverflow:
https://stackoverflow.com/questions/70139266/ksp-on-kotlin-multiplatform-fails-on-the-kspjs-with-collection-has-more-than-on

and also a code sample that can reproduce the error:
https://github.com/janvladimirmostert/observable-demo

ting-yuan:
The fix will be available in 1.0.2.

ting-yuan:
Btw, It looks to me that this only happens with `js(IR)`. As a workaround, could you please try to use `js()`/`js(LEGACY)` instead of `js(IR)`?

janvladimirmostert:
looking forward to try out 1.0.2, thanks @ting-yuan !

js legacy doesn't generate the TypeScript headers which means I can't use the generated JavaScript in external TypeScript projects. Since this is for demo-purposes only, it's not mission critical that it works, there are other ways to do what I want to do for my Kotlin talk.

mdietrichstein:
Thanks for fixing the issue @ting-yuan!

Just wanted to chime in and say that the project I'm working on also requires IR due to Typescript, but the feature is not a high priority one right now.

PR #756, opened 2021-12-07
Supported js(BOTH) and disallow binary generation for non-JVM targets

ting-yuan:
When `js(BOTH)` is used, we map `jsLegacy` and `jsIr` to `js` because both of the compiler created targets share the same source set. For simplicity and consistency, let's not distinguish the two targets when creating and specifying KSP configurations.

The mechanism of generating binary (or js) directly from processors is undefined yet.

### DIFF
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/KMPImplementedIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/KMPImplementedIT.kt
@@ -57,9 +57,16 @@ class KMPImplementedIT {
         )
 
         verify(
-            "workload/build/libs/workload-js-1.0-SNAPSHOT.jar",
+            "workload/build/libs/workload-jslegacy-1.0-SNAPSHOT.jar",
             listOf(
-                "playground-workload.js"
+                "playground-workload-js-legacy.js"
+            )
+        )
+
+        verify(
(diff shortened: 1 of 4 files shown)


ting-yuan:
Fixes #744

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/KMPImplementedIT.kt
```diff
@@ -57,9 +57,16 @@ class KMPImplementedIT {
...
             listOf(
-                "playground-workload.js"
+                "playground-workload-js-legacy.js"
+            )
+        )
+
+        verify(
+            "workload/build/libs/workload-jsir-1.0-SNAPSHOT.klib",
+            listOf(
+                "default/ir/types.knt"
```
neetopia, line 69: I found lots of files in the generated `klib` file, is this a special file we are interested in? I somehow reached to same test but I just passed an empty list for content to check.

neetopia (approved):

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/KMPImplementedIT.kt
ting-yuan: I chose this somewhat looks stable file only to make sure the output isn't empty. Future compiler bumps could break it but it'll be caught.

PR #761, opened 2021-12-09
Correct compiler args for JSIR tasks

ting-yuan:
Also updated test cases for all platforms.

TODO: K/N


ting-yuan:
Alright, K/N is just test case's fault.

neetopia (approved):

glureau:
fun `isIncrementalCompilationEnabled$kotlin_gradle_plugin`(): Boolean = false

Should I understand that incremental compilation is not usable with JS/IR ?

ting-yuan:
Being implemented as a compiler plugin, KSP Gradle plugin invokes the compiler in its own Gradle task. Therefore this only affects KSP's tasks where symbol processing happens, but not the normal compilation tasks.

Other targets also have incremental disable in corresponding ksp tasks (e.g.: [jvm](https://github.com/google/ksp/blob/main/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L539)). This hacky override is needed because JS/IR relies on it in addition to the general `incremental` property in Kotlin Gradle Plugin.

Also worth mentioning is that disabling incremental compilation has nothing to do with incremental processing. KSP needed to implement incremental processing independently because the dirty sets of incremental processing and incremental compilation are different.

PR #771, opened 2021-12-14
extract utility functions into a separate module

neetopia:


symbol-processing-cmdline/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+    `maven-publish`
+    signing
+}
+
+val packedJars by configurations.creating
+
+dependencies {
+    packedJars(project(":compiler-plugin")) { isTransitive = false }
+    packedJars(project(":util")) { isTransitive = false }
+    packedJars(project(":api")) { isTransitive = false }
```
ting-yuan, line 19: Although it is more convenient for command line, I'm not sure if `:api` should be included here. When using K/N in Gradle, the classes of the API jar will appear twice in the classpath: one here and one from Gradle deps. Can you remove `:api` so as to keep the current behavior?

ting-yuan (commented):
Can you check if there's any change before and after this PR in the published .pom files? Source and doc jars are also worth comparing.

api/build.gradle.kts
```diff
@@ -16,6 +17,9 @@ plugins {
     id("org.jetbrains.dokka") version ("1.4.32")
 }
 
+dependencies {
+    implementation(kotlin("stdlib", kotlinBaseVersion))
+}
```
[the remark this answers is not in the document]
neetopia: yes, we just never find it before because we don't pack `api` jar into `symbol-processing-cmdline`

ting-yuan (approved):

common-util/build.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+dependencies {
+    implementation(kotlin("stdlib", kotlinBaseVersion))
+    implementation("org.jetbrains.kotlin:kotlin-compiler:$kotlinBaseVersion")
+    implementation(project(":api"))
+}
+
+tasks {
+    val sourcesJar by creating(Jar::class) {
```
ting-yuan, line 31: Not needed / used

symbol-processing-cmdline/build.gradle.kts
```diff
@@ -29,21 +32,16 @@ tasks {
...
         dependsOn(project(":compiler-plugin").tasks["dokkaJavadocJar"])
-    }
-
-    val sourcesJar by creating(Jar::class) {
-        archiveClassifier.set("sources")
-        from(project(":compiler-plugin").sourceSets.main.get().allSource)
-        from(project(":common-util").sourceSets.main.get().allSource)
+        dependsOn(project(":compiler-plugin").tasks["sourcesJar"])
+        dependsOn(project(":common-util").tasks["dokkaJavadocJar"])
+        dependsOn(project(":common-util").tasks["sourcesJar"])
```
ting-yuan, line 37: unused: `dependsOn(project(":common-util").tasks["sourcesJar"])`

symbol-processing/build.gradle.kts
```diff
@@ -30,21 +33,18 @@ tasks {
...
         dependsOn(project(":compiler-plugin").tasks["dokkaJavadocJar"])
-    }
-
-    val sourcesJar by creating(Jar::class) {
-        archiveClassifier.set("sources")
-        from(project(":compiler-plugin").sourceSets.main.get().allSource)
-        from(project(":common-util").sourceSets.main.get().allSource)
+        dependsOn(project(":compiler-plugin").tasks["sourcesJar"])
+        dependsOn(project(":common-util").tasks["dokkaJavadocJar"])
+        dependsOn(project(":common-util").tasks["sourcesJar"])
```
ting-yuan, line 38: unused

symbol-processing-cmdline/build.gradle.kts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("shadow") {
+            artifactId = "symbol-processing-cmdline"
+            artifact(tasks["shadowJar"])
+            pom {
+                name.set("com.google.devtools.ksp:symbol-processing-cmdline")
+                description.set("Symbol processing for K/N and command line")
```
ting-yuan, line 48: These 2 lines are not enough. ShadowJar missed the pom file. You need to either create the dependencies manually lke [this](https://github.com/google/ksp/blob/main/symbol-processing/build.gradle.kts#L47), or figure out how to make ShadowJar copy over the POM file.

ting-yuan: For reference, this is the diff before and after the change in the POM:
```
-  <dependencies>
-    <dependency>
-      <groupId>org.jetbrains.kotlin</groupId>
-      <artifactId>kotlin-stdlib</artifactId>
-      <version>1.7.0-dev-1132</version>
-      <scope>runtime</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.jetbrains.kotlin</groupId>
-      <artifactId>kotlin-compiler</artifactId>
-      <version>1.7.0-dev-1132</version>
-      <scope>runtime</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.google.devtools.ksp</groupId>
-      <artifactId>symbol-processing-api</artifactId>
-      <version>2.0.0</version>
-      <scope>runtime</scope>
-    </dependency>
-  </dependencies>
```

common-util/build.gradle.kts
```diff
@@ -1,5 +1,7 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
+evaluationDependsOn(":api")
+
```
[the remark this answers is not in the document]
neetopia: there is a dependency on api module, therefore putting it here, not breaking anything if removed

symbol-processing-cmdline/build.gradle.kts
neetopia: it was copied from old `symbol-processing-cmdline` publication task, the pom was missing before because in K/N and cmd use case, pom is not needed, I followed the behavior, I can add it back if necessary.

ting-yuan: Thanks for adding it. So the interesting thing is that `symbol-processing-cmdline` WAS produced directly by the maven publish plugin, which produces POM automatically. It's ShadowJar, which we are using now, doesn't copy the POM. Also, although cmd line doesn't need it, K/N in Gradle does require it.

ting-yuan (approved):

Issue #775, opened 2021-12-16
`KSFunctionDeclaration.isAbstract` returns `true` for static methods inside java interface.


Jeffset:
`KSFunctionDeclaration.isAbstract` returns `true` for static methods originating from java interface.
This doesn't seem like intended behavior.
The trouble seems to be with `KSFunctionDeclarationJavaImpl` implementation.

ting-yuan:
Thanks for the fix!

PR #776, opened 2021-12-16
#775: Take `static` into account when computing `isAbstract` for java.

Jeffset:
Before this CL the method originating from Java interface was presumed abstract
if not marked `default`. This is not entirely correct, as static methods are also
not abstract. This CL fixes the issue.


neetopia:
Can you add a test case for this?

Jeffset:
> Can you add a test case for this?

Yes, done, I think.

neetopia (approved):

Jeffset:
@neetopia Hi, just wondered, is there any particular schedule to merge this?

ting-yuan:
Thanks for the patch!

PR #790, opened 2022-01-08
add KSP benchmark

neetopia:
* add an exhaustive processor that visits all files and invokes APIs
* apply a patch to Tachiyomi for applying exhaustive processor.


benchmark/runner/runner.sh
```diff
@@ -0,0 +1,12 @@
+cd ../exhaustive-processor
+./gradlew :processor:publishToMavenLocal
+cd ..
+mkdir tmp
+cd tmp
+git clone https://github.com/inorichi/tachiyomi.git
```
ting-yuan, line 6: Can you add a TODO comment that instead of installing to maven local, install to tmp/repo? so that it doesn't leave a state across runs.

benchmark/runner/runner.sh
```diff
@@ -0,0 +1,12 @@
...
+cd tachiyomi
+git checkout 938339690eecdfe309d83264b6a89aff3c767687
```
ting-yuan, line 8: any reason not to use a tag / release?

benchmark/runner/runner.sh
```diff
@@ -0,0 +1,12 @@
...
+git apply ../../runner/tachi.patch
```
ting-yuan, line 9: If commands below are commented out because you're going to run through command line or a customized launcher, this patch isn't needed. In that case the exhaustive processor doesn't need to be published / installed, either.

benchmark/exhaustive-processor/processor/src/main/kotlin/ExhaustiveProcessor.kt
```diff
@@ -0,0 +1,186 @@
...
+import com.google.devtools.ksp.processing.CodeGenerator
+import com.google.devtools.ksp.processing.Resolver
+import com.google.devtools.ksp.processing.SymbolProcessor
+import com.google.devtools.ksp.processing.SymbolProcessorEnvironment
+import com.google.devtools.ksp.processing.SymbolProcessorProvider
+import com.google.devtools.ksp.symbol.*
+import com.google.devtools.ksp.visitor.KSTopDownVisitor
+import java.io.File
+
+class ExhaustiveProcessor(
```
ting-yuan, line 11: Have you checked whether this ExhaustiveProcessor uses all the newly added APIs? Except those added in Resolver or utils.

ting-yuan (approved):

benchmark/build.sh
```diff
@@ -0,0 +1,20 @@
...
+cd tachiyomi
+git checkout 938339690eecdfe309d83264b6a89aff3c767687
+git apply $ROOT/benchmark/runner/tachi.patch
+./gradlew :app:copyDepsDevDebug
+mkdir $ROOT/benchmark/runner/lib && cp app/build/output/devDebug/lib/* $ROOT/benchmark/runner/lib/
+cd $ROOT
+./gradlew -PkspVersion=2.0.255 clean publishAllPublicationsToTestRepository
+cp -a build/repos/test/. $ROOT/benchmark/runner
+cd $ROOT/benchmark/runner
+./runner.sh -cp $(echo lib/*.jar | tr ' ' ':') $ROOT/benchmark/tachiyomi/app/**/*.kt
```
madsager, line 20: Is the `**/` here a zsh thing? If so, you can replace it with something like:

`find $ROOT/benchmark/tachiyomi/app/ -name "*.kt" | xargs ./runner.sh -cp $(echo lib/*.jar | tr ' ' ':') `

benchmark/runner/runner.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
```
madsager, line 2: Add `set -e` so that the script stops after the first thing that fails.

benchmark/runner/runner.sh
```diff
@@ -0,0 +1,28 @@
...
+KSP_PLUGIN_ID=com.google.devtools.ksp.symbol-processing
+KSP_PLUGIN_OPT=plugin:$KSP_PLUGIN_ID
+
+KSP_PLUGIN_JAR=./com/google/devtools/ksp/symbol-processing-cmdline/2.0.255/symbol-processing-cmdline-2.0.255.jar
+KSP_API_JAR=./com/google/devtools/ksp/symbol-processing-api/2.0.255/symbol-processing-api-2.0.255.jar
+
+AP=processor-1.0-SNAPSHOT.jar
+
+mkdir out
```
madsager, line 11: Maybe `mkdir -p out` so that you do not get errors if the directory exists?

benchmark/build.sh
madsager: Maybe remove the run of the benchmark from the builder script. That way, if the build succeeds, but the processor fails, golem will correctly report it as a benchmark run failure and not as a failure to build KSP.

benchmark/runner/tachi.patch
```diff
@@ -0,0 +1,55 @@
+diff --git a/app/build.gradle.kts b/app/build.gradle.kts
```
madsager, line 1: Maybe you should create a separate directory for each benchmark containing the code needed to generate the dependency jars? This patch file is in the runner dir which seems a little arbitrary. :)

benchmark/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/zsh
+
+ROOT=$PWD/..
+rm -rf runner/out runner/processor-1.0-SNAPSHOT.jar runner/com tachiyomi runner/lib
+./gradlew :jar
+cd $ROOT/benchmark/exhaustive-processor
+./gradlew build
+cp processor/build/libs/processor-1.0-SNAPSHOT.jar ../runner/
+cd $ROOT/benchmark
```
madsager, line 9: Should we create a temporary directory for this instead of checking out tachiyomi in our current checkout?

benchmark/bench-tachi/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
madsager, line 2: Add copyright notice, please. :)

benchmark/build.sh
```diff
@@ -0,0 +1,24 @@
...
+SCRIPT_DIR=$(dirname "$(realpath $0)")
+BENCHMARK=$1
+ROOT=$SCRIPT_DIR/..
+cd $SCRIPT_DIR
+rm -rf runner/out runner/processor-1.0-SNAPSHOT.jar
+./gradlew :jar
+cd $SCRIPT_DIR/exhaustive-processor
+./gradlew build
+cp processor/build/libs/processor-1.0-SNAPSHOT.jar ../runner/
+#cd $SCRIPT_DIR
```
madsager, line 13: Remove these parts that were moved to the benchmark build script?

benchmark/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
```
madsager, line 2: Add copyright notice (after the #! line) please. :)

benchmark/runner/runner.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
```
madsager, line 2: Copyright notice. :)

madsager (approved):
Thanks Jiaxiang. Let's get this landed so I can play with the go/lem setup for it over the next week. :)

PR #802, opened 2022-01-22
FIR: adapt to latest Analysis API and use upstream providers

jsjeon:
While working on integrating FIR UAST into Android Lint, I referred to KSP's use case (thanks!), and noticed that it's using a bit old version of Analysis API, and that it had unnecessary provider implementations. These commits bumped up the version of Analysis API; adapted to API changes; made uses of existing providers from upstream; and removed stale provider implementations.


neetopia:
🤦 Actually I have the same change on my repo: https://github.com/neetopia/ksp/tree/fir-bump

But your PR looks better since it also adds the upstreaming service providers, I will merge this one after taking a look.

neetopia (approved):
LGTM, thanks!

PR #804, opened 2022-01-24
FIR: more uses of upstream providers

jsjeon:
On top of https://github.com/google/ksp/pull/802, I keep adding provider registrations into Android Lint, and found two more places where upstream provider implementation or provider factory function can be used.


neetopia (approved):
Thanks!

PR #809, opened 2022-01-26
FIR: more uses of upstream providers and remove unnecessary ones

jsjeon:
This is another round of finding/using upstream providers (similar to https://github.com/google/ksp/pull/804).

Also, there are many unrelated providers registered. I guess those are needed for LSP (for editing), but not for KSP (and Android Lint). At least, we can have a commit history and revert if it turns out some of them are needed.


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/FirSealedClassInheritorsProcessorFactoryImpl.kt
```diff
@@ -1,19 +1,11 @@
...
 import org.jetbrains.kotlin.analysis.low.level.api.fir.api.services.FirSealedClassInheritorsProcessorFactory
-import org.jetbrains.kotlin.fir.declarations.FirRegularClass
 import org.jetbrains.kotlin.fir.declarations.SealedClassInheritorsProvider
-import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.fir.declarations.SealedClassInheritorsProviderImpl
 
 class FirSealedClassInheritorsProcessorFactoryImpl: FirSealedClassInheritorsProcessorFactory() {
     override fun createSealedClassInheritorsProvider(): SealedClassInheritorsProvider {
-        return SealedClassInheritorProviderImpl()
+        return SealedClassInheritorsProviderImpl
```
jsjeon, line 9: As mentioned in the internal doc, this one may require upfront FIR facade run. Otherwise, I wonder the attribute about sealed class inheritors are not ready. Well, it'd be just same as `emptyList()` below, for now.

neetopia: It does require a fir analysis, guess we don't really need to rely on this provider to get the sealed inheritors, instead run analyse phase and get the attribute manually should also work?

jsjeon: That upstream implementation actually returns the attribute (or empty list if not available), and that attribute is set if fir analysis run up to SEALED_... phase. So I think we need to run FIR analysis, not the whole phases, but up to that phase.

jsjeon: But, as per the discussion offline, we can fallback to the original implementation: empty list.

neetopia (approved):

Issue #816, opened 2022-01-31
`parent` and `parentDeclaration` of the generated `value` and `valueOf` static function from Java enums are null


kuanyingchou:
When processing Java enums from source, I can find the generated `values` and `valueOf` static functions with KSP. However, their `parent` and `parentDeclaration` are null, not the KSClassDeclaration that represents the enum type. Tested with 1.6.10-1.0.2.

ting-yuan:
Fixed in #817

PR #817, opened 2022-02-01
Find parents for LightMethod

ting-yuan:
Unfortunately, LightMethod doesn't implement parent and need a special
treatment.


ting-yuan:
Fixes #816.

The lengths of outputs of each implementation (kotlin lib, java lib, kotlin src, java src) are different because of inconsistencies in their implementations of `KSClassDeclaration.declarations`, etc. They need to be addressed in separate changes.

neetopia (approved):
I do have difficulty understanding the output of the test, but since you mentioned it in the comment, LGTM for now.

ting-yuan:
The output is really confusing. For example, these coexist:
```
parent of Y: YUV
parent of YUV: Y
```
The first is member and declaration container, and the second is class and its super.

Sadly, there is no "parent of RGB: R" because of the implementation consistency 🤦 

Also, there's no `values` nor `valueOf` of `CMYK` (kotlin src) 😢

PR #818, opened 2022-02-01
Remove workaround for KT-50889

ting-yuan:
where descriptors cached in CommonDependenciesContainer was
initialized multiple times.


ting-yuan:
This isn't needed in 1.0.3-release. Let's do this in the compiler bump.

PR #822, opened 2022-02-01
Update to Kotlin 1.7.0-dev-1132

ting-yuan:


ting-yuan:
Note that the fix to KT-50889 isn't available in this version yet.

neetopia (approved):

ting-yuan:
btw, I hit a JRE [bug](https://youtrack.jetbrains.com/issue/JBR-3434) while running integration tests locally. Problem is gone after switching to another brand (I used adopt jdk-11). Switching to jdk-17 also solves the issue.

PR #823, opened 2022-02-02
Unconditionally disable FIR in KSP Gradle tasks

ting-yuan:
KSP for FIR is still under development. This change allows KSP to be
used with compiler tasks where FIR is enabled.


ting-yuan:
Note that K/N doesn't work with FIR yet. It doesn't need and has no knob for FIR.

BTW, some multiplatform test cases are problematical. They generate sources in platform specific tasks (and therefore platform specific) and reference from common sources.

neetopia (approved):

PR #824, opened 2022-02-02
Handle more errors when replacing type arguments

ting-yuan:
Cases are:
* #arguments doesn't match
* There are error types in arguments

### DIFF
--- a/compiler-plugin/src/test/kotlin/com/google/devtools/ksp/processor/ReplaceWithErrorTypeArgsProcessor.kt
+++ b/compiler-plugin/src/test/kotlin/com/google/devtools/ksp/processor/ReplaceWithErrorTypeArgsProcessor.kt
@@ -51,10 +51,11 @@ open class ReplaceWithErrorTypeArgsProcessor : AbstractTestProcessor() {
         val yargs = y.type.element!!.typeArguments
 
         for (decl in decls) {
-            results.add(decl.asStarProjectedType().replace(xargs).toString())
-            results.add(decl.asStarProjectedType().replace(yargs).toString())
-            results.add(decl.asType(xargs).toString())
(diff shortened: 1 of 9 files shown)


compiler-plugin/src/test/kotlin/com/google/devtools/ksp/processor/ReplaceWithErrorTypeArgsProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+            resolver.getClassDeclarationByName("KLE")!!,
+            resolver.getClassDeclarationByName("KLE.E")!!,
+        )
+        val x = resolver.getPropertyDeclarationByName(resolver.getKSNameFromString("x"), true)!!
+        val xargs = x.type.element!!.typeArguments
+        val y = resolver.getPropertyDeclarationByName(resolver.getKSNameFromString("y"), true)!!
+        val yargs = y.type.element!!.typeArguments
+
+        for (decl in decls) {
+            results.add(decl.asStarProjectedType().replace(xargs).toString())
```
neetopia, line 54: what's the difference between calling and not calling `asStarProjectedType()`?

neetopia: might help reading test case if the name of the declaration is put in the beginning of lines?

ting-yuan: 1. Because there are two different APIs :) `KSType.replace` and `KSClassDeclaration.asType`.
2. Done improving the readability of the outputs.

neetopia (approved):

PR #833, opened 2022-02-09
use index to track memoized sequence iterator status

neetopia:
I have to use indexed access for following reasons:
 * `Iterator` won't work due to concurrent modification exception, this is not an issue for old implementation because in old implementation when the buffer exhausts, it reads from the main sequence iterator which is not the cache iterator that has the modification.
 * `MutableListIterator` won't work because `add` function of this iterator will add the element into the element before the `next` element, which effectively won't change the result of `next` call, which means the element being added is not access by iterator at all.

With these factors, it might make sense to implement our own list as well.


ting-yuan (approved):
This is good enough for this PR. The benefit of implementing a customized list is mainly that individual items can be GC'ed. Looks like a time / space trade-off to me and let's evaluate it later.

nit: can you make cache an ArrayList explicitly? I.e.,
```
private cache = arrayListOf<T>()
```

Although Kotlin's [default implementation of List is ArrayList](https://kotlinlang.org/docs/collections-overview.html#list), let's make it explicit now that we are using index.

Issue #784, opened 2021-12-23
Null default value for an annotation argument of an inner annotation type


Udinic:
Example:
// Java
public @interface InterestingAnnotation {
  @interface LevelOfInterest {
    boolean forEveryone() default false;
  }

  LevelOfInterest level() default @LevelOfInterest();
}

// Kotlin
@InterestingAnnotation
class KotlinTestClassKsp {}

When trying to get the argument values of InterestingAnnotation, I'll get a null value for the "level()" argument.

I noticed that if the annotation is declared in a separate module (Origin of its declaration is JAVA_LIB) - this works.

I tested this with 1.0.2.
Looks related to #692, but maybe it's not covering this specific case.

neetopia:
fixed by #836

PR #836, opened 2022-02-11
handle enum entry evaluation for annotation default value

### LINKED ISSUE #778: Default annotation param value is null when the param is a local Enum
Consider the following:

// FILE: KspTester.kt
class KspTester {
  @LocalAnnotation
  fun niceMethod(): Int {
    return 8
  }
}

annotation class LocalAnnotation(val annoParam: LocalEnum = LocalEnum.VALUE_1)

enum class LocalEnum {
  VALUE_1,
  VALUE_2
}
// END OF FILE: KspTester.kt

When checking the annotation values for niceMethod(), we'll get "annoParam" with a value of "null".

If I'll move LocalAnnotation and LocalEnum to a different module - I'll get a value of VALUE_1, as expected. 
The problem persists if I move these to a different file within the same module.

Tested with 1.6.0-1.0.2. 
Related issues #692, #53

neetopia:
fixes #778 #784


neetopia:
This is not the exact replica of the compiler resolution logic for enum expression, which is very lengthy (~300 lines), but this should serve the purpose for the annotation default value.

neetopia:
added another commit for #784 , reusing this PR to avoid conflicts.

ting-yuan (approved):

Issue #792, opened 2022-01-12
KSP needs a variant of Resolver.overrides that receives the containing ClassDeclaration


yigit:
Input:

```
             interface Interface {
                fun f1()
            }

            open class Base {
                fun f1() {
                }
            }

            class Subject : Interface, Base() {
                fun f2() {}
            }
```

The code above compiles successfully as `Subject` overrides `Interface.f1` via `Base.f1`. Unfortunately, there is no API in KSP to be able to check if `Base.f1` overrides `Interface.f1` when it is a member of `Subject`.

yigit:
Not sure if it will affect the implementation but same problem can happen with properties as well.

```
            interface Interface {
                val prop: String
                fun f1()
            }

            open class Base {
                val prop:String = ""
                fun f1() {
                }
            }

            class Subject : Interface, Base() {
                fun f2() {}
            }
            ```

PR #841, opened 2022-02-15
Add Resolver.overrides(overrider: KSDeclaration, overridee: KSDeclaration, containingClass: KSClassDeclaration): Boolean

neetopia:
this function checks if a declaration is overriding another declaration in the context of contained in the containing class.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -96,6 +96,7 @@ import org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluat
 import org.jetbrains.kotlin.resolve.descriptorUtil.classId
 import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperClassifiers
 import org.jetbrains.kotlin.resolve.descriptorUtil.module
+import org.jetbrains.kotlin.resolve.descriptorUtil.propertyIfAccessor
 import org.jetbrains.kotlin.resolve.lazy.DeclarationScopeProvider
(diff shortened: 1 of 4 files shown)


neetopia:
fixes #792

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -463,6 +464,58 @@ class ResolverImpl(
...
+        incrementalContext.recordLookupForDeclaration(candidate)
+        val originalDescriptor = when (original) {
+            is KSPropertyDeclaration -> resolvePropertyDeclaration(original)
+            is KSFunctionDeclaration -> resolveFunctionDeclaration(original)?.propertyIfAccessor
+            else -> return false
+        }
+
+        val candidateDescriptors = when (candidate) {
+            is KSPropertyDeclaration -> resolvePropertyDeclaration(candidate)?.overriddenDescriptors
+            is KSFunctionDeclaration -> resolveFunctionDeclaration(candidate)?.overriddenDescriptors
```
ting-yuan, line 478: Should we check whether `resolveFunctionDeclaration(candidate)` is a fake override or so? So as to make sure that it is the original. I.e., in case there is a `f1()` defined in `BaseOverride`.

BTW, `FunctionDescriptor.original` sounds like what we need but I'm not sure.

neetopia: original didn't help here, my guess is this is for type substitution, or other purposes.  we might not really want to find the one which is ultimately the original declaration, this function is mainly for filtering out unnecessary checkings, so leaving the check loose is safer than making it more strict.

ting-yuan (approved):

PR #844, opened 2022-02-15
bump compiler version to 1.7.0-dev-1452

neetopia:


ting-yuan (commented):
Can you remove the workaround for KT-50889?

ting-yuan (approved):

PR #853, opened 2022-02-17
Add EasyAdapter to supported libraries

AmrDeveloper:
[EasyAdapter](https://github.com/AmrDeveloper/EasyAdapter) is a Library to generate your adapters only with Annotations on your model using KSP

Thanks for KSP ❤️


neetopia:
Thanks for supporting KSP!

PR #867, opened 2022-02-23
update room reference in docs.

yigit:


neetopia (approved):
Thanks for the effort to official support of KSP :)

PR #868, opened 2022-02-23
Track API changes in CI

ting-yuan:
using Metalava.


neetopia (approved):
with this change, is it the expected behavior that API change will be monitored per commit and we need to manually accumulate these changes to compile a list of API changes when doing a release?

ting-yuan:
Yes. Potentially breaking changes like new functions will also be caught and the list needs to be updated if they don't harm. This build target generates the baseline file from current source: `:api:updateApi`. However, for some reason, the list isn't consistent with the source that it is from 🤷 Specifically, the `@interface` annotation of `KspExperiment` needs to be replaced with `interface` manually.

The build target in CI is `:api:checkApi` which can also run locally to check source code against `api.base`

PR #869, opened 2022-02-23
add else to when

sangyongchoi:
Hello.
I know this will be updated in Kotlin 1.6.
```
We will soon prohibit non-exhaustive when statements to make the behavior consistent with when expressions. To ensure smooth migration, Kotlin 1.6.0 reports warnings about non-exhaustive when statements with an enum, sealed, or Boolean subject. These warnings will become errors in future releases.
```

It seems to be a violation of this, so I will correct it and PR.
thank you.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
@@ -235,41 +235,41 @@ internal fun ModuleClassResolver.resolveContainingClass(psiMethod: PsiMethod): C
 }
 
 internal fun getInstanceForCurrentRound(node: KSNode): KSNode? {
-    when (node.origin) {
+    return when (node.origin) {
         Origin.KOTLIN_LIB, Origin.JAVA_LIB -> return null
-    }
-    return when (node) {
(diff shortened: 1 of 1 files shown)


ting-yuan:
Thanks! Could you sign the CLA so that we can merge it?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -235,41 +235,41 @@ internal fun ModuleClassResolver.resolveContainingClass(psiMethod: PsiMethod): C
 }
 
 internal fun getInstanceForCurrentRound(node: KSNode): KSNode? {
-    when (node.origin) {
+    return when (node.origin) {
         Origin.KOTLIN_LIB, Origin.JAVA_LIB -> return null
```
ting-yuan, line 239: Could you remove the `return` in `return null`?

sangyongchoi: done.

thank you for telling me

ting-yuan (approved):
Thanks!

sangyongchoi:
> can you fix the ktlint error?

@neetopia done.

neetopia:
Thanks! I will merge once CI is finished.

PR #871, opened 2022-02-25
Use Analysis API standalone mode entry points

jsjeon:
As entry points of Analysis API standalone mode is [available](https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide-plugin-dependencies/org/jetbrains/kotlin/analysis-api-standalone-for-ide/), this PR is about switching to use upstream entry points.


neetopia:
mannualy merged with decab20f1698020abedae7a142263e3261cfdbb3

Issue #861, opened 2022-02-21
Add support for variant aware KSP arguments


runningcode:
I would like to provide variant specific KSP arguments. For example, I would like `kspDebug` and `kspRelease` to have different arguments. 

An example use case is `Room` or `DeepLinkDispatch`. Both are kotlin symbol processors which write output to a directory. The directoy must be task specific so that `kspDebug` and `kspRelease` do not write to the same directory. Two tasks which write to the same directory disable up-to-date checks and caching in Gradle because the build becomes non-deterministic.

For reference, Java annotation processors and KAPT both provide a way of passing task specific arguments to the annotation processors.

runningcode:
Good question. KAPT [pulls them out of the java annotation processor options here](https://github.com/JetBrains/kotlin/blob/68f4580492894ee18a60319547dc376983a764d3/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt#L392-L405)

The Gradle Java plugin has a way of passing [`compilerArgumentProviders` via the dsl](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.compile.CompileOptions.html#org.gradle.api.tasks.compile.CompileOptions:compilerArgumentProviders).

I think we need to think about how all these work holistically. The approach that KAPT had to pull the arguments from the javacompile options might not make sense for KSP since the arguments would not affect the outcome of javacompile tasks nor KotlinCompile tasks.

We should also think about this in concert with [this PR](https://github.com/google/ksp/pull/872).

neetopia:
should this be closed now that linked PR is merged?

ting-yuan:
Nope, the merged PR is related but not solving this.

runningcode:
This also probably needs to be resolved first before a variant aware API compatible with the latest versions of AGP can be created: https://issuetracker.google.com/issues/223367107

neetopia:
I might not have full context on this issue, what's the current status?

runningcode:
Forget my comment above. It is not needed since we will not be passing arguments from AGP's `CommandLineArgumentProvider` to KSP.

This issue to add support for variant aware KSP arguments has no status update.

TomBell-Trove:
Would love this.  I want to pass an arg to my processor which varies based on my Android build variant to alter its behaviour during production builds.

[...]

PR #872, opened 2022-02-25
Add CommandLineArgumentProvider

### LINKED ISSUE #847: Add support for passing `CommandLineArgumentProvider` to `KspExtension`
Please add support for passing `CommandLineArgumentProvider` to the `KspExtension`.

https://docs.gradle.org/current/javadoc/org/gradle/process/CommandLineArgumentProvider.html

This will allow the KSP task to have cache-relocateable inputs in order to avoid cache misses in [Room]() and [DeepLinkDispatch]()

[Room]: https://github.com/airbnb/DeepLinkDispatch#generated-deep-links-documentation
[DeepLinkDispatch]: https://issuetracker.google.com/issues/132245929

runningcode:
This allows one to configure the KspExtension with a CommandLineArgumentProvider.
Inputs and outputs can now properly declared on any additional input or output directories to the KSP compilation.

Fixes #847

Note: this also upgrades the Room dependency to 2.4.2 across the project since 2.3.0 does not support KSP.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -139,8 +139,8 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             }
             kspExtension.commandLineArgumentProviders.forEach {
                 val argument = it.asArguments().joinToString("")
-                if (!argument.matches(Regex("/S=/S"))) {
-                    throw IllegalArgumentException("KSP apoption does not match /S=/S: $argument")
+                if (!argument.matches(Regex("\\S+=\\S+"))) {
+                    throw IllegalArgumentException("KSP apoption does not match \\S+=\\S+: $argument")
                 }
                 options += SubpluginOption("apoption", argument)
             }
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -138,7 +138,11 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                 options += SubpluginOption("apoption", "${it.key}=${it.value}")
             }
             kspExtension.commandLineArgumentProviders.forEach {
-                options += SubpluginOption("apoption", it.asArguments().joinToString(""))
+                val argument = it.asArguments().joinToString("")
+                if (!argument.matches(Regex("/S=/S"))) {
+                    throw IllegalArgumentException("KSP apoption does not match /S=/S: $argument")
+                }
+                options += SubpluginOption("apoption", argument)
             }
             return options
         }
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -38,6 +38,7 @@ import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.compile.AbstractCompile
 import org.gradle.api.tasks.compile.JavaCompile
 import org.gradle.language.jvm.tasks.ProcessResources
+import org.gradle.process.CommandLineArgumentProvider
 import org.gradle.tooling.provider.model.ToolingModelBuilderRegistry
 import org.gradle.util.GradleVersion
 import org.gradle.work.Incremental
@@ -136,6 +137,9 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kspExtension.apOptions.forEach {
                 options += SubpluginOption("apoption", "${it.key}=${it.value}")
             }
+            kspExtension.commandLineArgumentProviders.forEach {
+                options += SubpluginOption("apoption", it.asArguments().joinToString(""))
+            }
             return options
         }
     }
@@ -228,6 +232,7 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                     )
                 }
             )
+            kspTask.commandLineArgumentProviders.addAll(kspExtension.commandLineArgumentProviders)
             kspTask.destination = kspOutputDir
             kspTask.blockOtherCompilerPlugins = kspExtension.blockOtherCompilerPlugins
             kspTask.apOptions.value(kspExtension.arguments).disallowChanges()
@@ -392,6 +397,9 @@ internal fun findJavaTaskForKotlinCompilation(compilation: KotlinCompilation<*>)
 interface KspTask : Task {
     @get:Internal
     val options: ListProperty<SubpluginOption>
+
+    @get:Nested
+    val commandLineArgumentProviders: ListProperty<CommandLineArgumentProvider>
 
     @get:OutputDirectory
     var destination: File
--- a/gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
+++ b/gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
@@ -18,6 +18,7 @@ package com.google.devtools.ksp.gradle
 
 import com.google.common.truth.Truth.assertThat
 import com.google.devtools.ksp.gradle.processor.TestSymbolProcessorProvider
+import com.google.devtools.ksp.gradle.testing.DependencyDeclaration.Companion.artifact
 import com.google.devtools.ksp.gradle.testing.DependencyDeclaration.Companion.module
 import com.google.devtools.ksp.gradle.testing.KspIntegrationTestRule
 import com.google.devtools.ksp.processing.CodeGenerator
@@ -26,6 +27,7 @@ import com.google.devtools.ksp.processing.KSPLogger
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.processing.SymbolProcessor
 import com.google.devtools.ksp.symbol.KSAnnotated
+import org.gradle.testkit.runner.TaskOutcome
 import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.TemporaryFolder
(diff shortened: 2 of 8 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -136,6 +137,9 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kspExtension.apOptions.forEach {
                 options += SubpluginOption("apoption", "${it.key}=${it.value}")
             }
+            kspExtension.commandLineArgumentProviders.forEach {
+                options += SubpluginOption("apoption", it.asArguments().joinToString(""))
```
ting-yuan, line 141: The compiler plugin side assumes each apoption to be of the form "k=v". You probably need one more loop instead of `joinToString("")`. E.g.,
```
kspExtension.commandLineArgumentProviders.forEach {
  it.asArguments.forEach {
    options += SubpluginOption("apoption", it)
  } 
}
```

On the other hand, it looks like that there's no guaranteed form of strings in `CommandlineArgumentProvider.asArguments()`. Can you insert a check here to make sure that all arguments follow the form "k=v"?

runningcode: Sounds good. The minimum check would be that `asArguments().joinToString("")` matches the regex `\S=\S`?

I will add that regex check.

runningcode: regex and test added

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -136,6 +137,13 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kspExtension.apOptions.forEach {
                 options += SubpluginOption("apoption", "${it.key}=${it.value}")
             }
+            kspExtension.commandLineArgumentProviders.forEach {
+                val argument = it.asArguments().joinToString("")
+                if (!argument.matches(Regex("/S=/S"))) {
+                    throw IllegalArgumentException("KSP apoption does not match /S=/S: $argument")
+                }
+                options += SubpluginOption("apoption", argument)
```
PhilGlass, line 145: This differs from [the kapt implementation](https://github.com/JetBrains/kotlin/blob/68f4580492894ee18a60319547dc376983a764d3/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/Kapt3KotlinGradleSubplugin.kt#L392-L405) in a couple of ways:

1️⃣ A `SubpluginOption`'s value - which contains an absolute path in the case of `room.schemaLocation` - [becomes a task input](https://github.com/JetBrains/kotlin/blob/8753ea592f5c682bb94e0d24875113290512324a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt#L1087-L1112). It might be worth adding an integration test that uses the local build cache from a different location and checking that there are cache hits?

2️⃣ A `CommandLineArgumentProvider` can have arbitrary lazy properties that are not necessarily safe to read at this point in the configuration phase. It'll appear to work for simple cases like `room.schemaLocation` pointing to a static directory, but I think it'll break when there are properties that can only be safely evaluated at execution time - see [this kapt bug](https://youtrack.jetbrains.com/issue/KT-39715).

For some reason `InternalSubpluginOption` doesn't support lazy values, so I think the workaround here would be to use `CompositeSubpluginOption("apoption", lazy { it.asArguments().joinToString("")) }, emptyList())` (you could also pass a list containing a single `InternalSubpluginOption` as the final argument, which probably makes more sense semantically but behaves equivalently in practice).

runningcode: Thanks for pointing this out. I will update to make this more in line with the KAPT approach.

runningcode: I think we need to take a step back and think about how this would work in concert with #861 before moving forward here.

ting-yuan: KSP didn't call `Task.registerSubpluginOptionsAsInputs`. It registers AP options as task inputs through [KspTask.apOptions](https://github.com/google/ksp/blob/510130e9d32623d0fe1f8498a38afc418f4ba6c3/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L412) instead. The options are evaluated at execution time (wrapped by a provider in configuration time, and `get()` at execution time). Therefore the issue is slightly different: they need to be registered manually.

@runningcode We plan to build 1.0.5 release on Apr. 6. Can we land this first and handle variants in another patch?

runningcode: Thanks for the update.
Can you tell me which of the concerns that Phil raised should be addressed in this PR and how they should be addressed?

ting-yuan: Calling `asArguments()` and creating `SubpluginOption` here is fine because it is evaluated at execution time.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -228,6 +236,7 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                     )
                 }
             )
+            kspTask.commandLineArgumentProviders.addAll(kspExtension.commandLineArgumentProviders)
```
ting-yuan, line 239: Because it's unsafe to call `asArgument()` at configuration time, I guess we need to wrap this with project.provider like `kspTask.options` in the lines above. (Or, is `CommandLineArgumentProvider` already considered lazy values? If so, LGTM)
```
kspTask.commandLineArgumentProviders.addAll(
    kspTask.project.provider {
        kspExtension.commandLineArgumentProviders
    }
)
```

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
ting-yuan: To be clear, the only concern I have is that this may need to be wrapped by a provider:
```
kspTask.commandLineArgumentProviders.addAll(kspExtension.commandLineArgumentProviders)
```

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -136,6 +137,13 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kspExtension.apOptions.forEach {
                 options += SubpluginOption("apoption", "${it.key}=${it.value}")
             }
+            kspExtension.commandLineArgumentProviders.forEach {
+                val argument = it.asArguments().joinToString("")
+                if (!argument.matches(Regex("/S=/S"))) {
```
ting-yuan, line 142: Pardon me if this is a stupid question, but I cannot find what `/S` mean. Kotlin's [Regex doc](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/) says that it is [java.util.regex.Pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) but I can't find it in the doc.

Can we replace the check with `if ('=' !in argument) throw ...`?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
PhilGlass: Ah, my mistake! I didn't check the call chain properly and assumed all `SubpluginOption`s went through that method.

I don't _think_ there's any need for further laziness, since calling `asArguments()` is already guarded by the provider that creates the list of options?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
PhilGlass: I think the intention here was to match anything except whitespace (`\S`), but the regex isn't quite right - it should be something more like `"\\S+=\\S+"`.

https://pl.kotl.in/pItEH8Yd8

ting-yuan:
@runningcode Could you fix the regex and lint error before we merge it?

runningcode:
regex and lint error fixed.

runningcode:
I ran the test locally and it passed. It passed on the macOS ci but not linux. Is this a flaky test? It might relate to the changes though.

gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
```diff
@@ -115,4 +117,117 @@ class GradleCompilationTest {
...
+                            println("${'$'}{option.key}=${'$'}{option.value}")
+                        }
+                    }
+                }
+                
+            """.trimIndent()
+        )
+        val result = testRule.runner().withArguments(":app:assembleDebug").build()
+        assertThat(result.output)
+            .contains("apoption=room.schemaLocation=/private${testRule.appModule.moduleRoot}/schemas")
```
ting-yuan, line 199: The location seems platform dependent. The error message is

```
    expected to contain:
        apoption=room.schemaLocation=/private/tmp/junit16660779885054734687/junit9088641825282887015/app/schemas
    but was:
        ...
        apoption=room.schemaLocation=/tmp/junit16660779885054734687/junit9088641825282887015/app/schemas
```

Can you use a regexp to match it? E.g.,

```
    val pattern1 = Regex.escape("apoption=room.schemaLocation=")
    val pattern2 = Regex.escape("${testRule.appModule.moduleRoot}/schemas")
    assertTrue(Regex("$pattern1\\S*$pattern2").containsMatchIn(result.output))
```

ting-yuan:
Let's merge this as is and allow me to fix the test in a separate commit. Thank you so much!

Issue #876, opened 2022-02-28
KSClassDeclaration.isOpen returns `true` for nested data classes


yigit:
For the given input:

```
interface NestedDataClassSubject {
    data class NestedDataClass(
      val field: String,
     )
  }
```

Calling `NestedDataClass.isOpen` returns `true` instead of `false`.

Looks like it is coming from [this line](https://github.com/google/ksp/blob/main/api/src/main/kotlin/com/google/devtools/ksp/utils.kt#L218) but it is very old code so I couldn't figure out why we check for the parent declaration's `classKind`.

I've not tested but reading the code, this probably affects any other code. 

Related room issue: https://issuetracker.google.com/issues/221280927

neetopia:
the major reason is that the interface members are effectively open for member functions and properties, but it looks like nested class inside interface is an outlier here.

yigit:
oh i just noticed that while trying to write the test in KSP.

I think this is intended for methods and properties in interfaces, you are right.

yigit:
sent a PR: https://github.com/google/ksp/pull/877

PR #877, opened 2022-02-28
Don't mark nested classes as open

yigit:
This CL fixes a bug in in KSP where we were marking nested
classes as public by mistake (the code seems to focus on
properties and methods of interfaces).

To keep a similar logic, I just excluded KSClassDeclarations
from that check.

I've also updated the test to be a bit more readable to only
list properties that return true.

Fixes: #876
Test: declarationUtil.kt


neetopia (approved):

neetopia:
LGTM, will merge once lint is fixed.

yigit:
so nice to see we finally have lint 👍 .
Fixed.

Issue #851, opened 2022-02-17
API for `KSType` for checking whether the underlying Java type is raw.


Jeffset:
Sometimes it's required to known this when overriding Java methods. There's no way (except for the slightly ridiculous `"raw " in type.toString()`) to check that now. Can we have a proper API method for that?

PR #889, opened 2022-03-07
#851: Add isJavaRawType function to Resolver.

Jeffset:
This is required for checking if the underlying type is Java's raw type.
Such information is sometimes required to correctly override Java methods
 while generating Java code.


Jeffset:
I see that CI checks failed due new API being added. I assume that adding abstract non-default method to a public interface is a breaking change. So what is the flow of adding new API? The #851's  target version is `1.0.5` which is not a major version bump. So should I try to introduce new API in a non-breaking way, like adding it to `Resolver` as utility method? Help much appreciated :)

ting-yuan:
Adding it to `Resolver` is preferable because `isRawType` is only applicable to JVM. Platform specific functions are put under `Resolver` so that KS* won't be too verbose / complicated.

You'll likely still get an API check error when adding it to `Resolver` but that should be fine; We care mostly about ABI compatibility and need an error if the baseline file is not up-to-date.

ting-yuan (approved):
Thanks!

PR #898, opened 2022-03-15
bump compiler and analysis API version to 1.7.0-dev-2437

neetopia:


kotlin-analysis-api/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ description = "Kotlin Symbol Processing implementation using Kotlin Analysis API
 
 val intellijVersion: String by project
 val junitVersion: String by project
-val analysisAPIVersion = "1.7.0-dev-1984"
+val analysisAPIVersion = "1.7.0-dev-2437"
```
ting-yuan, line 5: May I know why we need a different variable for this? Looks like it is the same as `kotlinBaseVersion`.

ting-yuan (approved):

kotlin-analysis-api/build.gradle.kts
ting-yuan: Also could you move it to `gradle.properties` if the variable needs to exist alongside with `kotlinBaseVersion`?

neetopia: the release frequency of analysis API and bootstrap compiler were different. I will remove this in a separate PR later.

ting-yuan: Sounds good. BTW, should the compiler and stdlib version stick with `kotlinBaseVersion` instead of using `analysisAPIVersion` below?

neetopia: maybe just make everything depending on `kotlinBaseVersion`?

PR #913, opened 2022-03-29
synthetic constructors should have null qualified name, to align with non synthetic constructors

neetopia:


jaschdoc:
@ting-yuan is this still relevant?

PR #916, opened 2022-03-30
Benchmark script fixes.

madsager:
To build the benchmark runner and ksp:

`./benchmark/build.sh`

To build the TachiyomiExhaustive benchmark libs:

`ANDROID_SDK_ROOT=<path to root> ./benchmark/TachiyomiExhaustive/build.sh`

To run the benchmark:

`./benchmark/runner/runner.sh TachiyomiExhaustive ..`

The second argument to runner.sh is the relative path from the script
directory to where the Tachiyomi source and libs is located. This
will be different on the benchmark runners, so needs to be passed
in.


benchmark/TachiyomiExhaustive/build.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
madsager, line 1: The only change in this file is the comment syntax in the copyright notice.

benchmark/build.sh
```diff
@@ -1,33 +1,28 @@
...
 SCRIPT_DIR=$(dirname "$(realpath $0)")
-BENCHMARK=$1
 ROOT=$SCRIPT_DIR/..
 cd $SCRIPT_DIR
 rm -rf runner/out runner/processor-1.0-SNAPSHOT.jar
 ./gradlew :jar
 cd $SCRIPT_DIR/exhaustive-processor
 ./gradlew build
 cp processor/build/libs/processor-1.0-SNAPSHOT.jar ../runner/
-cd $SCRIPT_DIR/$BENCHMARK
```
madsager, line 29: We should not build the benchmark data here, we should just build ksp and the runner. We will build and upload the benchmark data separately.

benchmark/runner/runner.sh
```diff
@@ -1,26 +1,26 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 set -e
 BENCHMARK_DIR=$1
+PREFIX=$2
```
madsager, line 19: This is a bit messy, but we need flexibility here to make this work on the runners.

benchmark/src/main/kotlin/com/google/devtools/ksp/BenchRunner.kt
```diff
@@ -20,16 +20,17 @@ import org.jetbrains.kotlin.cli.jvm.K2JVMCompiler
 import kotlin.system.measureNanoTime
 
 fun main(args: Array<String>) {
-    val runs = 10
-    val warmup = 10
-    val benchName = "Tachiyomi"
+    val runs = 3
+    val warmup = 3
+    val benchName = args[0]
```
madsager, line 25: To make the runner usable for other benchmarks as well. :)

benchmark/src/main/kotlin/com/google/devtools/ksp/BenchRunner.kt
```diff
@@ -20,16 +20,17 @@ import org.jetbrains.kotlin.cli.jvm.K2JVMCompiler
 import kotlin.system.measureNanoTime
 
 fun main(args: Array<String>) {
-    val runs = 10
-    val warmup = 10
-    val benchName = "Tachiyomi"
+    val runs = 3
```
madsager, line 23: This takes a long time to run, so reducing the number of runs to 3. We might want to parameterize the runner, but we can do that later.

neetopia (approved):

PR #917, opened 2022-03-30
Align superTypes among implementationsSupertypes

ting-yuan:
If there is no direct super type, return kotlin.Any. Otherwise, hide Any.

This is the behavior of descriptors / libraries.


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -102,7 +103,19 @@ class KSClassDeclarationImpl private constructor(val ktClassOrObject: KtClassOrO
...
 
     override val superTypes: Sequence<KSTypeReference> by lazy {
-        ktClassOrObject.superTypeListEntries.asSequence().map { KSTypeReferenceImpl.getCached(it.typeReference!!) }
+        val resolver = ResolverImpl.instance
+        ktClassOrObject.superTypeListEntries
+            .asSequence()
+            .map { KSTypeReferenceImpl.getCached(it.typeReference!!) }
+            .ifEmpty {
+                sequenceOf(
+                    KSTypeReferenceDescriptorImpl.getCached(
```
neetopia, line 112: should it be `KSTypeReferenceSyntheticImpl`?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/java/KSClassDeclarationJavaImpl.kt
```diff
@@ -138,7 +138,14 @@ class KSClassDeclarationJavaImpl private constructor(val psi: PsiClass) :
...
     override val superTypes: Sequence<KSTypeReference> by lazy {
-        psi.superTypes.asSequence().map { KSTypeReferenceJavaImpl.getCached(it, this) }.memoized()
+        val adjusted = if (!psi.isInterface && psi.superTypes.size > 1) {
+            psi.superTypes.filterNot {
+                it.className == "Object" && it.canonicalText == "java.lang.Object"
+            }
+        } else {
+            psi.superTypes.toList()
+        }
+        adjusted.asSequence().map { KSTypeReferenceJavaImpl.getCached(it, this) }.memoized()
```
neetopia, line 148: Will this add `Object` to super types if super type list empty?

ting-yuan: I doubt that the super list from psi can be empty; It seems to me that it is always available from psi, according to the test cases.

Issue #125, opened 2020-10-18
KSType assignability between java string and kotlin string returns false


yigit:
```
val javaString = resolver.getClassDeclarationByName("java.lang.String")!!.asStarProjectedType()
val kotlinString = resolver.getClassDeclarationByName("kotlin.String")!!.asStarProjectedType()
println(javaString.isAssignableFrom(kotlinString).toString())
println(kotlinString.isAssignableFrom(javaString).toString())
```
this prints `false` `false` instead of `true` `true`.

I'm not yet sure why [this test](https://github.com/google/ksp/blob/master/compiler-plugin/testData/api/javaTypes.kt) does not catch it though :/. Maybe `getClassDeclarationByName` has a different resolution path ?

ting-yuan:
This is intended. `java.lang.String` is [mapped](https://kotlinlang.org/docs/reference/java-interop.html#mapped-types) to `kotlin.String!` when written in Kotlin source. We didn't perform the mapping in `getClassDeclarationByName`, so that it's possible to get the original class.

neetopia:
Close this issue for now as we have a discussion thread going on in #126

PR #920, opened 2022-03-31
Fix subtype checking for Java types.

### LINKED ISSUE #890: `isAssignableFrom` behaves incorrectly with Java-specific types (from `Resolver.getJavaClassByName`)
Suppose we do the following:
```kotlin
val javaCollection = resolver.getJavaClassByName("kotlin.collections.Collection")  // maps to `java.util.Collection`
val javaSet = resolver.getJavaClassByName("kotlin.collections.Set")  // maps to `java.util.Set`
javaCollection.isAssignableFrom(javaSet)  // Expected: true, actual: false
```
As far as my investigations went, Kotlin compiler basically walks type hierarchy. And the superinterface for `java.util.Set` is `kotlin.collections.Collection` for some reason, not `java.util.Collection`. Thus, `java.util.Collection` is never encountered and check ends with negative result.

neetopia:
Compiler subtype checking does not convert Java types to Kotlin types, while getting super types from a java type does the conversion, therefore resulting in subtype checking for Java types to fail.
Check if candidate super type is a Java type, convert to Kotlin type for subtype checking.
Potentially needs an ultimate fix in compiler for subtype checking.

fixes #890

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
@@ -26,15 +26,14 @@ import com.google.devtools.ksp.symbol.KSTypeArgument
 import com.google.devtools.ksp.symbol.Nullability
 import com.google.devtools.ksp.symbol.Origin
 import com.google.devtools.ksp.symbol.impl.binary.KSTypeArgumentDescriptorImpl
+import com.google.devtools.ksp.symbol.impl.convertKotlinType
 import com.google.devtools.ksp.symbol.impl.replaceTypeArguments
 import org.jetbrains.kotlin.builtins.isFunctionType
 import org.jetbrains.kotlin.builtins.isKFunctionType
 import org.jetbrains.kotlin.builtins.isKSuspendFunctionType
 import org.jetbrains.kotlin.builtins.isSuspendFunctionType
 import org.jetbrains.kotlin.descriptors.NotFoundClasses
-import org.jetbrains.kotlin.types.KotlinType
-import org.jetbrains.kotlin.types.getAbbreviation
-import org.jetbrains.kotlin.types.isError
+import org.jetbrains.kotlin.types.*
 import org.jetbrains.kotlin.types.typeUtil.TypeNullability
 import org.jetbrains.kotlin.types.typeUtil.isSubtypeOf
 import org.jetbrains.kotlin.types.typeUtil.makeNotNullable
@@ -75,9 +74,10 @@ class KSTypeImpl private constructor(
     }
 
     override fun isAssignableFrom(that: KSType): Boolean {
-        val subType = (that as? KSTypeImpl)?.kotlinType ?: return false
+        val subType = (that as? KSTypeImpl)?.kotlinType?.convertKotlinType() ?: return false
         ResolverImpl.instance.incrementalContext.recordLookupWithSupertypes(subType)
-        return subType.isSubtypeOf(kotlinType)
+        val thisType = (this as? KSTypeImpl)?.kotlinType?.convertKotlinType() ?: return false
+        return subType.isSubtypeOf(thisType)
     }
 
     // TODO: find a better way to reuse the logic in [DescriptorRendererImpl.renderFlexibleType].
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
@@ -22,6 +22,7 @@ import com.google.devtools.ksp.KspExperimental
 import com.google.devtools.ksp.findPsi
 import com.google.devtools.ksp.processing.impl.ResolverImpl
 import com.google.devtools.ksp.symbol.*
+import com.google.devtools.ksp.symbol.Variance
 import com.google.devtools.ksp.symbol.impl.binary.KSClassDeclarationDescriptorImpl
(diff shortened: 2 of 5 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
```diff
@@ -77,7 +81,18 @@ class KSTypeImpl private constructor(
...
-        return subType.isSubtypeOf(kotlinType)
+        val declarationDescriptor = kotlinType.constructor.declarationDescriptor
+        // Compiler subtype checking does not convert Java types to Kotlin types, while getting super types
+        // from a java type does the conversion, therefore resulting in subtype checking for Java types to fail.
+        // Check if candidate super type is a Java type, convert to Kotlin type for subtype checking.
+        val thisType = if (declarationDescriptor is JavaClassDescriptor) {
+            JavaToKotlinClassMapper
+                .mapJavaToKotlin(declarationDescriptor.fqNameSafe, ResolverImpl.instance.module.builtIns)
+                ?.defaultType
+                ?.replace(kotlinType.arguments)
```
ting-yuan, line 92: Does this need to be done recursively, e.g., is `java.util.Container<java.util.Set>` assignable from `kotlin.collections.Set<kotlin.collections.Set>`?

ting-yuan (commented):
1. I just recalled the discussion in #125, where whether java types and kotlin types should be assignable to each other were debated. Although I'm not too worried about the behavior change (which needs to be announced / emphasized in release notes though), all places that involve type checking need to be consistent. For example, overrides. How many places will need to be updated?

2. If I'm reading this right, does it mean that, before this PR, kotlin.collections.Set is assignable from java.util.Set but not the other way around?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
neetopia: I thought about this case when writing test, then realized it is impossible to do so, because we are only getting Java types via `getJavaClassByName`, from where we can only get a `KSType`, and because we only support build type argument with a type reference, therefore KSType can't be used to build a type argument. Which in turn, means that I can only try to get such test case in code, and while resolving types in source, all these types in the type arguments are already converted to kotlin types, therefore it is impossible to reproduce this in test case, nor in real practice.

neetopia:
> 1. I just recalled the discussion in #125, where whether java types and kotlin types should be assignable to each other were debated. Although I'm not too worried about the behavior change (which needs to be announced / emphasized in release notes though), all places that involve type checking need to be consistent. For example, overrides. How many places will need to be updated?

I think the major issue here is that when getting a java class, we are not converting them into kotlin types, therefore left a door open for checking java types against kotlin types and breaks compiler subtype checking logic. Both compiler and KSP has been converting these Java types into Kotlin types and I think they are mostly working fine, we are only getting this issue in #890 because we allowed a non-converted type to be returned via `getJavaClassByName`

> 2. If I'm reading this right, does it mean that, before this PR, kotlin.collections.Set is assignable from java.util.Set but not the other way around?
Yes, actually this PR does not fix this case either, I will update PR to fix this.

ting-yuan:
> > 1. I just recalled the discussion in [KSType assignability between java string and kotlin string returns false #125](https://github.com/google/ksp/issues/125), where whether java types and kotlin types should be assignable to each other were debated. Although I'm not too worried about the behavior change (which needs to be announced / emphasized in release notes though), all places that involve type checking need to be consistent. For example, overrides. How many places will need to be updated?
> 
> I think the major issue here is that when getting a java class, we are not converting them into kotlin types, therefore left a door open for checking java types against kotlin types and breaks compiler subtype checking logic. Both compiler and KSP has been converting these Java types into Kotlin types and I think they are mostly working fine, we are only getting this issue in #890 because we allowed a non-converted type to be returned via `getJavaClassByName`
> 

Allow me to rephrase: can you check places where `KSType` is used as a parameter and see whether they also need conversion or not?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
ting-yuan: Except that we have `Resolver.createKSTypeReferenceFromKSType()` 😢

neetopia:
> > > 1. I just recalled the discussion in [KSType assignability between java string and kotlin string returns false #125](https://github.com/google/ksp/issues/125), where whether java types and kotlin types should be assignable to each other were debated. Although I'm not too worried about the behavior change (which needs to be announced / emphasized in release notes though), all places that involve type checking need to be consistent. For example, overrides. How many places will need to be updated?
> > 
> > 
> > I think the major issue here is that when getting a java class, we are not converting them into kotlin types, therefore left a door open for checking java types against kotlin types and breaks compiler subtype checking logic. Both compiler and KSP has been converting these Java types into Kotlin types and I think they are mostly working fine, we are only getting this issue in #890 because we allowed a non-converted type to be returned via `getJavaClassByName`
> 
> Allow me to rephrase: can you check places where `KSType` is used as a parameter and see whether they also need conversion or not?


There are usages in asMemberOf, I will add a check there as well.

neetopia:
Actually, because the implementation in `asMemberOf` is only checking super types already, not the type itself, same issue won't be happening for `asMemberOf`, although I do see a minor concern for `asMemberOf` not supporting the containing class to be itself, it does not seem to be a valid use case for `asMemberOf` anyway. I would fix it for `asMemberOf` if the subtype checking logic would include the type being checked itself.

ting-yuan (approved):

Issue #881, opened 2022-03-02
KSP Processing order doesn't process room class first before typealias


cp-yfukuda:
With room being KSP dependent, if I have this code, 

```
pacakge com.mypackage

@Entity(tableName = "my_data_table")
data class MyData(
}
```

And in other file in the same package, But different fille

```
pacakge com.mypackage

import com.mypackage.DataProvider

data class Identifier<T> private constructor(private val uuid: UUID) {
}

...

typealias DataProviderIdentifier = Identifier<MyData>

```

This compiration with ksp with room gives me 

`
[ksp] java.lang.IllegalStateException: Unexpected type name for KSType: com.mypackage.Identifier<com.mypackage.DataProvider>

`

This makes me think that alias processing happenes before ksp is processing room annotation.

ting-yuan:
Can you share the stacktrace and/or a sample project that we can reproduce it?

cp-yfukuda:
Hi Ting-Yuan 

I made a simple version of the app, with typealias using Room data, but 
it works. 
https://drive.google.com/file/d/1Dbp217h-I9_M25BYNng5slfR9V2Y7SI3/view?usp=sharing
But here is actual stacktrace. 

````
[ksp] java.lang.IllegalStateException: Unexpected type name for KSType: org.test.Identifier< com.mypackage.DataProvider>
	at androidx.room.compiler.processing.ksp.KSTypeExtKt.typeName(KSTypeExt.kt:214)
	at androidx.room.compiler.processing.ksp.KSTypeExtKt.typeName(KSTypeExt.kt:71)
	at androidx.room.compiler.processing.ksp.KSTypeExtKt.typeName$resolveTypeName(KSTypeExt.kt:153)
	at androidx.room.compiler.processing.ksp.KSTypeExtKt.typeName(KSTypeExt.kt:171)
	at androidx.room.compiler.processing.ksp.KSTypeExtKt.typeName(KSTypeExt.kt:193)
	at androidx.room.compiler.processing.ksp.KSTypeExtKt.typeName(KSTypeExt.kt:181)
	at androidx.room.compiler.processing.ksp.DefaultKspType.resolveTypeName(DefaultKspType.kt:32)
	at androidx.room.compiler.processing.ksp.KspType$typeName$2.invoke(KspType.kt:54)
	at androidx.room.compiler.processing.ksp.KspType$typeName$2.invoke(KspType.kt:51)
[...]

PR #928, opened 2022-04-03
Use abbreviation when returning arguments for a KSType

yigit:
This PR fixes a problem in KSType where it was not returning the right type arguments for TypeAliases.

For input:
```
typealias MyList = MyList<Int>
val x: MyList = TODO()
```

Querying the typeArguments of `x` would return `[Int]` while also returning its declaration as `MyList`.
This makes it looks like `MyList<Int><Int>`.
To fix it, I updated `KSType.arguments` to use the abbreviation type instead of the KotlinType.
Also, updated the test to add more cases and make it print the arguments for each type. I refactored the
test a bit to print types of properties instead of all found types to make the output more readable (happy to revert that part).

Fixes: #881
Test: typeAlias.kt


ting-yuan (approved):
Thank you soooo much!

PR #939, opened 2022-04-07
Enable CI on Windows

ting-yuan:


gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
```diff
@@ -120,6 +121,8 @@ class GradleCompilationTest {
 
     @Test
     fun testCommandLineArgumentProvider() {
```
neetopia, line 123: Can we use the same `DisabledOnOs` annotation?

gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/SourceSetConfigurationsTest.kt
```diff
@@ -235,19 +236,19 @@ class SourceSetConfigurationsTest {
             }
         ).containsExactly(
             SourceFolder(
-                "debug", "SRC:generated/ksp/debug/java"
```
neetopia, line 238: maybe consider write an overloaded version of SourceFolder() that will handle the File separator by itself instead of repeating the same string replacing operation every time calling it?

neetopia: Or maybe add a handy extension function for this, given this is repeated everywhere, something like `String().normalize()`?

gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
ting-yuan: It's only available in JUnit 5. Upgrading to it is non-trivial.

gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/SourceSetConfigurationsTest.kt
ting-yuan: Done.

neetopia (approved):
WIndows CI seems to be particularly slow btw.

Issue #674, opened 2021-10-13
Add ability to retrieve default values from annotation members


jameskleeh:
I'm in the beginning stages of writing an implementation of Micronaut's compile time features using KSP. Micronaut stores annotation data in classes that get created at compile time through a custom API called AnnotationMetadata. That API allows us to read annotations at runtime without using runtime reflection.

 To reduce memory usage at runtime, default values of annotation members are not stored every time the annotation is used. Instead the default values are stored a single time per annotation. 

To support this use case I would need the following:

- A way to retrieve the default values for a given annotation
- The ability determine if the value coming from `KSValueArgument` is because of the default value or because it is actually set

Please let me know if I can be of any assistance. Thank you!

Jeffset:
Would this be solved by the API requested in #542?

jameskleeh:
@Jeffset Perhaps partially, however that wouldn't cover the case of needing to know if the value in an annotation was set or is the default value.

Benjoyo:
I am looking for the same feature. I need to know whether the annotation property is set or not. Right now I have to compare the value to the default value.

PR #943, opened 2022-04-08
Add support for knowing if an annotation value was declared or not

jameskleeh:


jameskleeh:
This completes part of https://github.com/google/ksp/issues/674

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSValueArgumentImpl.kt
```diff
@@ -25,6 +25,7 @@ class KSValueArgumentImpl(private val namedAnnotationValue: KtNamedAnnotationVal
         KSNameImpl(namedAnnotationValue.name.asString())
     }
     override val isSpread: Boolean = false
+    override val isDefault: Boolean = false
```
jameskleeh, line 28: I'm not sure if this is correct. I don't understand what this module is used for or where the data comes from

neetopia: You can put a TODO() here, this is a WIP development module.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSValueArgumentImpl.kt
```diff
@@ -26,15 +26,22 @@ import com.google.devtools.ksp.symbol.KSVisitor
...
     override val value: Any?,
-    override val parent: KSNode?
+    override val parent: KSNode?,
+    override val isDefault: Boolean
 ) : KSValueArgumentImpl() {
-    companion object : KSObjectCache<Triple<KSName, Any?, KSNode>, KSValueArgumentLiteImpl>() {
-        fun getCached(name: KSName, value: Any?, parent: KSNode) = cache
-            .getOrPut(Triple(name, value, parent)) { KSValueArgumentLiteImpl(name, value, parent) }
+    companion object : KSObjectCache<Int, KSValueArgumentLiteImpl>() {
+        private fun getKey(name: KSName, value: Any?, parent: KSNode, isDefault: Boolean): Int {
```
neetopia, line 38: I think it is ok to keep the old hashing input, they are enough to distinguish each other.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSValueArgumentImpl.kt
```diff
@@ -26,15 +26,22 @@ import com.google.devtools.ksp.symbol.KSVisitor
...
+            return Objects.hash(name, value, parent, isDefault)
+        }
+        fun getCached(name: KSName, value: Any?, parent: KSNode, isDefault: Boolean) = cache
+            .getOrPut(getKey(name, value, parent, isDefault)) {
+                KSValueArgumentLiteImpl(name, value, parent, isDefault)
+            }
     }
 
     override val origin = Origin.KOTLIN
```
neetopia, line 47: On a side note, i am thinking whether we should put `Origin.SYNTHETIC` for default annotation arguments, as they are not technically present in the source code, I can do it myself afterwards though, feel free to ignore.

neetopia (commented):
Alternatively, if we want to also fix origin for default annotation arguments, maybe we can put correct origin in the constructor as use that to determine the value os `isDefault`?

neetopia:
After thinking again, I feel it makes more sense to fix the `origin` for default annotation arguments, and then add a utility function in `util.kt` where you can simply write `fun KSValueArgument.isDefault() = origin == Origin.SYNTHETIC", therefore we don't need to add an API to KSValueArgument

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSValueArgumentImpl.kt
jameskleeh: OK

jameskleeh:
@neetopia That is OK with me. I'll push out an update

neetopia (approved):
LGTM, can you add a test case that checks a value argument returns `isDefault() == false` while it has default value in declaration site, but we applied a value to it anyway? So that the test makes sure that we are returning if a value argument **is** a default value, rather than returning if a value argument **could** have default value.

I can also add the test myself if you feel like.

jameskleeh:
@neetopia I'm not sure I totally understand what the test you're suggesting should verify, but I took a stab at it by modifying the existing test to set a value equal to the default value and assert isDefault returns false.

compiler-plugin/testData/api/annotationWithDefault.kt
```diff
@@ -146,7 +146,7 @@ public @interface JavaAnnotation2 {
 
 // FILE: JavaAnnotated.java
 
-@KotlinAnnotation(ConstKt.DebugKt)
```
neetopia, line 149: Can you keep this as is? This is probably caused during your rebase with latest main branch.

neetopia (commented):
Yes that's what I mean. I see there is conflict with the main branch, can you resolve it so that I can merge this PR.

jameskleeh:
@neetopia Fixed that merge issue and I merged main into this branch, however it showed no conflicts

```
ksp annotation_isdefault $ git fetch --all
Fetching origin
Fetching upstream
ksp annotation_isdefault $ git merge upstream/main
Already up to date.
```

neetopia:
Manually pushed in 5f89aa3b889ee0c37080e6492b0485ee8383d959 Thanks for PR!

BTW I think you should do a rebase with main since we are doing rebase when merging PRs, I can get the conflict when executing `git rebase main`

PR #954, opened 2022-04-13
preprocess file classes into binding trace

### LINKED ISSUE #839: `null` value in annotation arguments if annotation originates from Java and argument is Kotlin's `const val`
Assume we have a Kotlin source, named `Constants.kt`, as follows:
```Kotlin
const val FOO = "foo"
```
And a Java source:
```Java
class Bar {
    @Named(ConstantsKt.FOO) void set(/* */) { /* */ }
}
```
When `@Named` annotation (`KSAnnotationJavaImpl`) is resolved, it yields `["value:null"]` from its `arguments` property, with no way to retrieve the `"foo"` value.

*Named = javax.inject.Named* for completeness.

So, basically, Kotlin constants are not properly resolved in Java annotation arguments.

neetopia:
fixes #839


ting-yuan (approved):
Can you open an issue in 1.0.6 as a reminder to check the performance change?

neetopia:
tracking created, will merge after final confirmation with JB.

PR #977, opened 2022-05-05
1.0.5 release

neetopia:


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/OnErrorIT.kt
```diff
@@ -105,8 +105,6 @@ class OnErrorIT {
                     it.startsWith("e: [ksp] kotlin.io.FileAlreadyExistsException:")
                 }
             )
-
-            Assert.assertTrue(result.output.contains("e: java.lang.IllegalStateException: Should not be called!"))
```
jsjeon, line 109: Was this from AA, and somehow the path to that is fixed?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -641,10 +631,35 @@ abstract class KspTaskJvm @Inject constructor(
...
+            commonSourceSet,
+            classpathSnapshotProperties.classpath,
+            classpathSnapshotProperties.classpathSnapshot
+        )
+
+    @get:InputFiles
+    @get:SkipWhenEmpty
+    @get:IgnoreEmptyDirectories
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    override val sources: FileCollection = super.sources.filter {
```
jsjeon, line 674: Nit (i.e., not necessary to fix now or follow-up): Some parameter below is renamed to `kotlinSources` (from `sourceRoots`); also there is `javaSources` below. So... I guess this is `kotlinSources` as well? Not familiar what `super` is here, and maybe need to be renamed from there. Or... is there a convention about this?

gradle.properties
```diff
@@ -1,7 +1,7 @@
 # Copied from kotlinc
 org.gradle.jvmargs=-Duser.country=US -Dkotlin.daemon.jvm.options=-Xmx2200m -Dfile.encoding=UTF-8
 
-kotlinBaseVersion=1.7.0-dev-3025
+kotlinBaseVersion=1.7.0-Beta
```
jsjeon, line 4: Hm, even beta has version numbers. Does it pick the latest one?

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/SealedClassProcessor.kt
```diff
@@ -31,7 +31,7 @@ class SealedClassProcessor : AbstractTestProcessor() {
         resolver.getNewFiles().forEach { f ->
             f.declarations.forEach {
                 if (it is KSClassDeclaration) {
-                    val subs = it.getSealedSubclasses().map { it.simpleName.asString() }.toList()
+                    val subs = it.getSealedSubclasses().map { it.simpleName.asString() }.toList().sorted()
```
jsjeon, line 34: Just curious; why wasn't it sorted before, and why is it sorted now? :)

jsjeon (approved):

gradle.properties
neetopia: I'm releasing for this release: https://github.com/JetBrains/kotlin/releases/tag/v1.7.0-Beta it does seems to not have version numbers other than just beta.

link to maven artrifact: https://search.maven.org/artifact/org.jetbrains.kotlin/kotlin-gradle-plugin/1.7.0-Beta/jar

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/SealedClassProcessor.kt
neetopia: Because of the dependency to compiler version, behavior of KSP API's returning order can actually change if there is an upstream change, and therefore we will just use sort() to get rid of the checking on the ordering if it doesn't matter.

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/OnErrorIT.kt
neetopia: This is testing for old implementation, which was fixed later in upstream.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
neetopia: It is actually quite messy here.. the names are mostly from upstream kotlin gradle plugin therefore we need to follow these naming.

gradle.properties
jsjeon: Interesting. Beta versions I saw are here: https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide-plugin-dependencies/org/jetbrains/kotlin/analysis-api-standalone-for-ide/

neetopia: that one is some other JetBrains managed dev repositories, the versioning is quite confusing tbh

PR #981, opened 2022-05-10
try to get psi before creating sealed sub class declarations

### LINKED ISSUE #936: Could not extract location from getSealedSubclasses declaration
We use KSP to analyze all classes in module and file location for each class is crucial. 
However, if we try to do something like: 
```kotlin
classDeclaration
  .getSealedSubclasses()
  .forEach { sealedClass ->
    sealedClass.location //or sealedClass.parentDeclaration.containingFile
  }
```
We will get `NonExistLocation` or `null`. 
Is it an expected behavior or could it be fixed in the future?

neetopia:
fixes #936 

Cause is that we did not check the psi for the sealed sub classes descriptors.

Testing `toLocation()` is not possible in our test infrastructure, hence test the origin instead, when origin is `KOTLIN` there will be a location value from the psi.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
@@ -23,15 +23,12 @@ import com.google.devtools.ksp.symbol.*
 import com.google.devtools.ksp.symbol.impl.*
 import com.google.devtools.ksp.symbol.impl.java.KSFunctionDeclarationJavaImpl
 import com.google.devtools.ksp.symbol.impl.java.KSPropertyDeclarationJavaImpl
-import com.google.devtools.ksp.symbol.impl.kotlin.KSErrorType
-import com.google.devtools.ksp.symbol.impl.kotlin.KSFunctionDeclarationImpl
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
```diff
@@ -189,5 +186,11 @@ internal fun ClassDescriptor.sealedSubclassesSequence(): Sequence<KSClassDeclara
     // TODO record incremental subclass lookups in Kotlin 1.5.x?
     return sealedSubclasses
         .asSequence()
-        .map { KSClassDeclarationDescriptorImpl.getCached(it) }
+        .map { sealedSubClass ->
+            if (sealedSubClass.findPsi() is KtClassOrObject) {
+                KSClassDeclarationImpl.getCached(sealedSubClass.findPsi() as KtClassOrObject)
+            } else {
+                KSClassDeclarationDescriptorImpl.getCached(sealedSubClass)
+            }
```
jsjeon, line 194: Nit: can be written to:
```kt
when (val psi = sealedSubClass.findPsi()) {
  is KtClassOrObject -> ...(psi)
  else -> ...(sealedSubClass)
}
```

jsjeon (approved):

PR #988, opened 2022-05-13
Get symbols with annotation

neetopia:
aimed for passing test `hello.kt`, blocked due to no annotations available on java filed symbols, local test result as following:

expected:
```
8
test.Foo
test.ITF
Bar.list
Bar.BBB
AClass
C
C.f
C.javaFun
```

actual
```
7
test.Foo
test.ITF
Bar.list
Bar.BBB
AClass
C
C.javaFun
```


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    override val qualifiedName: KSName by lazy {
+        KSNameImpl.getCached(ktJavaFieldSymbol.callableIdIfNonLocal?.asSingleFqName()?.asString() ?: "")
+    }
+
+    override val typeParameters: List<KSTypeParameter>
+        get() = emptyList()
+
+    override val packageName: KSName
+        get() = TODO("Not yet implemented")
```
jsjeon, line 74: `callableIdIfNonLocal` used above has package name in it.

For `KSPropertyDeclarationImpl`, you can do the same retrieval (at `qualifiedName` and here `packageName`). Maybe, one day, it's better to introduce `abstract` impl of `KSPropertyDeclaration` for `KtVariableSymbol` where some logics can be easily shared.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.google.devtools.ksp.symbol.Location
+import com.google.devtools.ksp.symbol.Modifier
+import com.google.devtools.ksp.symbol.Origin
+import com.google.devtools.ksp.toKSModifiers
+import org.jetbrains.kotlin.analysis.api.symbols.KtJavaFieldSymbol
+import org.jetbrains.kotlin.psi.KtModifierListOwner
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+class KSPropertyDeclarationJavaImpl private constructor(
+    private val ktJavaFieldSymbol: KtJavaFieldSymbol
```
jsjeon, line 25: Just asking :) What's the plan for `KtSyntheticJavaPropertySymbol`? `KSPropertyDeclarationSyntheticJavaPropertyImpl` or `KSSyntheticJavaPropertyDeclarationImpl`? Maybe this one should emphasize java "field" somehow?

jsjeon (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
neetopia: This is essentially just field, the name `property` is quite confusing tbf, but since KSP models everything in the Kotlin way, we just use `property` to describe java fields.

PR #993, opened 2022-05-17
use correct type for vararg value parameter descriptor implementation

### LINKED ISSUE #958: Inconsistency in type of `vararg` parameter of overriden method
Say I have these 2 classes:

```
@ProcessThis
interface Sub : Super

@ProcessThis
interface Super {
  fun foo(vararg values: String)
}
```

When processing any of the KSClassDeclarations with `getAllFunctions().filter { it.isAbstract }.forEach { ... }`, the `values` parameter for `Super` has 
```
  KSValueParameter.isVararg == true
  KSValueParameter.type.resolve() == 'String'
```
but `Sub` has
```
  KSValueParameter.isVararg == true
  KSValueParameter.type.resolve() == 'Array<String>'
```
which is very inconsistent.

They should either both be, `vararg=true type=String` or both be `vararg=true type=Array<String>`

neetopia:
compiler implementation of `isVararg` is essentially checking `varargElementType != null`.

fixes #958

### DIFF
--- a/compiler-plugin/testData/api/allFunctions.kt
+++ b/compiler-plugin/testData/api/allFunctions.kt
@@ -75,6 +75,28 @@
 // equals(kotlin.Any): kotlin.Boolean
 // hashCode(): kotlin.Int
 // toString(): kotlin.String
+// class: Sub
+// equals(kotlin.Any): kotlin.Boolean
+// foo(kotlin.String ...): kotlin.Unit
+// hashCode(): kotlin.Int
+// toString(): kotlin.String
+// class: SubAbstract
+// <init>(): SubAbstract
+// equals(kotlin.Any): kotlin.Boolean
+// foo(kotlin.String ...): kotlin.Unit
+// hashCode(): kotlin.Int
+// toString(): kotlin.String
+// class: Super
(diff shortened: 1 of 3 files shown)


compiler-plugin/testData/api/allFunctions.kt
```diff
@@ -75,6 +75,28 @@
...
+// toString(): kotlin.String
+// class: SubAbstract
+// <init>(): SubAbstract
+// equals(kotlin.Any): kotlin.Boolean
+// foo(kotlin.String): kotlin.Unit
+// hashCode(): kotlin.Int
+// toString(): kotlin.String
+// class: Super
+// equals(kotlin.Any): kotlin.Boolean
+// foo(kotlin.String): kotlin.Unit
```
jsjeon, line 91: It might be good to render this as `foo(kotlin.String...)` or `foo(vararg kotlin.String)` to indicate that value parameter is `vararg`?

jsjeon (approved):

compiler-plugin/testData/api/allFunctions.kt
neetopia: Good point

PR #994, opened 2022-05-19
First verbose mode support.

### LINKED ISSUE #927: Any verbosity options available?
Is there any equivalent to Java annotation processing's `-XprintRounds` or Kapt's `KaptTask.verbose` to get extra debugging information about KSP's processing or rounds?

neetopia:
* Logs current round number of processing.
* Logs loaded providers.
* Logs an error when there is no providers found in AP classpath.

fixes #927 

also fulfills a providers related logging request from Donald.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
@@ -76,6 +76,14 @@ class KotlinSymbolProcessingExtension(
 
                 ServiceLoaderLite.loadImplementations(SymbolProcessorProvider::class.java, classLoader)
             }
+            if (providers.isEmpty()) {
+                logger.error("No providers found in processor classpath.")
+            } else {
+                logger.info(
+                    "loaded provider(s): " +
+                        "${providers.joinToString(separator = ", ", prefix = "[", postfix = "]") { it.javaClass.name }}"
+                )
+            }
         }
         return providers
     }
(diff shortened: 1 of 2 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -76,6 +76,13 @@ class KotlinSymbolProcessingExtension(
 
                 ServiceLoaderLite.loadImplementations(SymbolProcessorProvider::class.java, classLoader)
             }
+            if (providers.isEmpty()) {
+                logger.error("No providers found in processor classpath.")
+            } else {
+                providers.forEach {
+                    logger.info("loaded provider: ${it.javaClass.name}")
+                }
```
jsjeon, line 84: Nit: not sure we prefer repeated prefix "loaded provider" for multiple providers? How about just:
```kt
logger.info(
    "loaded provider(s): ${providers.joinToString(separator = ", ", prefix = "[", postfix = "]") { it.javaClass.name }}"
)
```
(or without brackets)

jsjeon (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
neetopia: that was actually my original implementation (without brackets), I am open to all kinds of suggestions on this matter.

PR #1005, opened 2022-05-26
fix annotation value resolution for annotations declared in Java source.

### LINKED ISSUE #1004: Annotation method return type resolved as property for java sources.
When processing Java source for an annotation like:

```
@interface MyAnnotation {
    boolean booleanParam();
}
```

I'm hitting a `ClassCastException` when trying to get the return type of the booleanParam() method.

```
Caused by: java.lang.ClassCastException: class org.jetbrains.kotlin.load.java.descriptors.JavaPropertyDescriptor cannot be cast to class org.jetbrains.kotlin.descriptors.FunctionDescriptor (org.jetbrains.kotlin.load.java.descriptors.JavaPropertyDescriptor and org.jetbrains.kotlin.descriptors.FunctionDescriptor are in unnamed module of loader 'app')
    at com.google.devtools.ksp.processing.impl.ResolverImpl.resolveFunctionDeclaration(ResolverImpl.kt:640)
    at com.google.devtools.ksp.symbol.impl.java.KSFunctionDeclarationJavaImpl.findOverridee(KSFunctionDeclarationJavaImpl.kt:55)
    at androidx.room.compiler.processing.ksp.KSFunctionExtKt.getOriginatingReference(KSFunctionExt.kt:57)
    at androidx.room.compiler.processing.ksp.KSFunctionExtKt.returnKspType(KSFunctionExt.kt:35)
    at androidx.room.compiler.processing.ksp.KspMethodElement$KspNormalMethodElement$returnType$2.invoke(KspMethodElement.kt:165)
    at androidx.room.compiler.processing.ksp.KspMethodElement$KspNormalMethodElement$returnType$2.invoke(KspMethodElement.kt:164)
    at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
    at androidx.room.compiler.processing.ksp.KspMethodElement$KspNormalMethodElement.getReturnType(KspMethodElement.kt:164)
    at androidx.room.compiler.processing.XAnnotationValueTest$tmp$1.invoke(XAnnotationValueTest.kt:762)
    at androidx.room.compiler.processing.XAnnotationValueTest$tmp$1.invoke(XAnnotationValueTest.kt:724)
```

Also note that it's only when processing the annotation as a source -- it works as expected when the using the precompiled class

neetopia:
fixes #1004 

Currently this issue seems to only happen for annotations declared in Java source.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -352,7 +352,11 @@ class ResolverImpl(
     internal fun mapToJvmSignatureInternal(declaration: KSDeclaration): String? = when (declaration) {
         is KSClassDeclaration -> resolveClassDeclaration(declaration)?.let { typeMapper.mapType(it).descriptor }
         is KSFunctionDeclaration -> resolveFunctionDeclaration(declaration)?.let {
-            typeMapper.mapAsmMethod(it).descriptor
+            when (it) {
+                is FunctionDescriptor -> typeMapper.mapAsmMethod(it).descriptor
+                is PropertyDescriptor -> typeMapper.mapFieldSignature(it.type, it) ?: typeMapper.mapType(it).descriptor
+                else -> throw IllegalStateException("Unexpected descriptor type for declaration: $declaration")
+            }
         }
         is KSPropertyDeclaration -> resolvePropertyDeclaration(declaration)?.let {
             typeMapper.mapFieldSignature(it.type, it) ?: typeMapper.mapType(it).descriptor
@@ -414,7 +418,8 @@ class ResolverImpl(
         incrementalContext.recordLookupForDeclaration(candidate)
         val originalDescriptor = when (original) {
             is KSPropertyDeclaration -> resolvePropertyDeclaration(original)
-            is KSFunctionDeclaration -> resolveFunctionDeclaration(original)?.propertyIfAccessor
+            is KSFunctionDeclaration ->
+                resolveFunctionDeclaration(original).safeAs<FunctionDescriptor>()?.propertyIfAccessor
             else -> return false
         }
 
@@ -557,7 +562,7 @@ class ResolverImpl(
         } as ClassDescriptor?
     }
 
-    fun resolveFunctionDeclaration(function: KSFunctionDeclaration): FunctionDescriptor? {
+    fun resolveFunctionDeclaration(function: KSFunctionDeclaration): CallableDescriptor? {
         return when (function) {
             is KSFunctionDeclarationImpl -> resolveDeclaration(function.ktFunction)
             is KSFunctionDeclarationDescriptorImpl -> function.descriptor
@@ -581,7 +586,7 @@ class ResolverImpl(
                 resolved?.unsubstitutedPrimaryConstructor ?: resolved?.constructors?.singleOrNull()
             }
             else -> throw IllegalStateException("unexpected class: ${function.javaClass}")
-        } as FunctionDescriptor?
+        } as? CallableDescriptor
     }
 
     fun resolvePropertyDeclaration(property: KSPropertyDeclaration): PropertyDescriptor? {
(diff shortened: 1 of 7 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -581,7 +586,7 @@ class ResolverImpl(
                 resolved?.unsubstitutedPrimaryConstructor ?: resolved?.constructors?.singleOrNull()
             }
             else -> throw IllegalStateException("unexpected class: ${function.javaClass}")
-        } as FunctionDescriptor?
```
jsjeon, line 584: There are more places in this file that used this style of type casting. You can clean up those in a follow-up PR.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -352,7 +352,11 @@ class ResolverImpl(
     internal fun mapToJvmSignatureInternal(declaration: KSDeclaration): String? = when (declaration) {
         is KSClassDeclaration -> resolveClassDeclaration(declaration)?.let { typeMapper.mapType(it).descriptor }
         is KSFunctionDeclaration -> resolveFunctionDeclaration(declaration)?.let {
-            typeMapper.mapAsmMethod(it).descriptor
```
jsjeon, line 355: This just caught my eyes. Note that, as per https://youtrack.jetbrains.com/issue/KT-48532, old backend will be deprecated, and so will be `KotlinTypeMapper` (separately discussed at https://youtrack.jetbrains.com/issue/KT-48773/Investigate-the-possibility-of-removing-dependency-on-old-JVM-ba). You can also refer to http://b/186742973 or https://github.com/JetBrains/intellij-community/commit/4f41d9888c5b9f71d3e8b7864879eb7eec8704d4 to see how FE1.0 UAST avoids using it (well, replicate it w/ simpler/necessary part of generic rewriting).

compiler-plugin/testData/api/overridee.kt
```diff
@@ -85,6 +85,8 @@
 // JavaAccessorImpl.getX() -> KtInterfaceWithProperty.x
 // JavaAccessorImpl.getY() -> KtInterfaceWithProperty.y
 // JavaAccessorImpl.setY(value:Int) -> KtInterfaceWithProperty.y
+// JavaAnno:
+// JavaAnno.intParam() -> null
```
jsjeon, line 89: Just to double-check: `-> null` means that it doesn't override anything?

compiler-plugin/testData/api/overridee.kt
```diff
@@ -295,3 +297,8 @@ class JavaAccessorImpl implements KtInterfaceWithProperty {
     public void setY(int value) {
     }
 }
+
+// FILE: JavaAnno.java
+@interface JavaAnno {
```
jsjeon, line 302: It might be good to test repeating annotations as well, but can be done in a follow-up PR.

jsjeon (approved):

compiler-plugin/testData/api/overridee.kt
neetopia: yes

compiler-plugin/testData/api/overridee.kt
neetopia: added, tbh I still don't know what differs by being repeating.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
neetopia: Thanks for the heads up, follow up in #1006

compiler-plugin/testData/api/overridee.kt
jsjeon: I expected to see array return type. :)

neetopia: Ugh... That should be in another test that is checking for jvm name. This test only checks for override.

PR #1014, opened 2022-06-07
introduce processorName option for KSP.

neetopia:
This option will allow users to specify certain processors(s) to be
executed, if not specified, all processors found in apclasspath will be
executed.

### DIFF
--- a/common-util/src/main/kotlin/com/google/devtools/ksp/KspOptions.kt
+++ b/common-util/src/main/kotlin/com/google/devtools/ksp/KspOptions.kt
@@ -189,6 +189,13 @@ enum class KspCliOption(
         false
     ),
 
+    PROCESSOR_NAME_OPTION(
+        "processorNames",
+        "<processorNames>",
+        "only executes following processor(s)",
+        false
+    ),
+
     KNOWN_MODIFIED_OPTION(
         "knownModified",
         "<knownModified>",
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspExtension.kt
```diff
@@ -33,6 +33,8 @@ open class KspExtension {
         apOptions.put(k, v)
     }
 
+    open var processorName: List<String> = emptyList()
```
gavra0, line 36: Rename to `processorNames`, considering there are 1+ you can specify?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -165,6 +165,9 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kspExtension.apOptions.forEach {
                 options += SubpluginOption("apoption", "${it.key}=${it.value}")
             }
+            kspExtension.processorName.forEach {
+                options += SubpluginOption("processorName", "$it")
```
gavra0, line 169: Is this a task input? Also, consider adding an integration test in which you modify processor names and assert that task is not up-to-date.

neetopia: done, thanks for pointing this out!

ting-yuan:
Setting the processor names in KspExtension makes it hard to track in Gradle tooling, which would need to depend on KSP implementation, instead of the common approach of reading configuration / dependencies. OTOH, if users only need certain processors to be loaded, they should probably not include other dependencies in the beginning. Could we remove the support in Gradle, i.e., remove the option in KspExtension? So as to keep it only available in command line for now.

ting-yuan (approved):

PR #1018, opened 2022-06-09
disable symbol link check on Windows

neetopia:
Previous symbolic link fix is causing issues for incremental on Windows, disabling this fix for Windows for now.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
@@ -172,7 +172,11 @@ abstract class AbstractKotlinSymbolProcessingExtension(
                         is KSFileJavaImpl -> it.psi
                         else -> null
                     }?.virtualFile?.let { virtualFile ->
-                        File(virtualFile.path).canonicalPath
+                        if (System.getProperty("os.name").startsWith("windows", ignoreCase = true)) {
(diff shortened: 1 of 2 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -172,7 +172,11 @@ abstract class AbstractKotlinSymbolProcessingExtension(
                         is KSFileJavaImpl -> it.psi
                         else -> null
                     }?.virtualFile?.let { virtualFile ->
-                        File(virtualFile.path).canonicalPath
+                        if (System.getProperty("os.name").startsWith("windows", true)) {
```
jsjeon, line 175: I was about to ask whether it is `Windows` or `windows` and suggest to use `lowerCase()`. It looks like the 2nd parameter is about that. Could you add param name `ignoreCase =` explicitly?

jsjeon (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
neetopia: Done. also included an additional commit for changing the build platform of release artifact. It seems that the command for building on linux can't be directly reused on Windows, we changed build platform after last release and did not test the release action until today.

PR #1026, opened 2022-06-23
AA: Implement KSErrorType

neetopia:
also includes an implementation for KSTypeAlias, as well as some refactoring for common logics for KSDeclaration.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSErrorType.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSErrorType.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2022 Google LLC
+ * Copyright 2010-2022 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.devtools.ksp.impl.symbol.kotlin
+
+import com.google.devtools.ksp.symbol.KSAnnotation
+import com.google.devtools.ksp.symbol.KSDeclaration
+import com.google.devtools.ksp.symbol.KSType
+import com.google.devtools.ksp.symbol.KSTypeArgument
+import com.google.devtools.ksp.symbol.Nullability
+
+object KSErrorType : KSType {
+    override val declaration: KSDeclaration
+        get() = KSErrorTypeClassDeclaration
+
+    override val nullability: Nullability
+        get() = Nullability.NULLABLE
+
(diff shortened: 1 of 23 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSValueArgumentImpl.kt
```diff
@@ -51,4 +51,8 @@ class KSValueArgumentImpl private constructor(
     override fun <D, R> accept(visitor: KSVisitor<D, R>, data: D): R {
         return visitor.visitValueArgument(this, data)
     }
+
+    override fun toString(): String {
+        return "${name?.asString() ?: ""}:$value"
```
jsjeon, line 56: If `name` is `null`, we'll get something like `:v`?

Also, if it is a named argument, I expected "name = value" instead.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSErrorType.kt
```diff
@@ -0,0 +1,43 @@
+package com.google.devtools.ksp.impl.symbol.kotlin
```
jsjeon, line 1: Copyright

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSErrorTypeClassDeclaration.kt
```diff
@@ -0,0 +1,94 @@
...
+    override val simpleName: KSName = KSNameImpl.getCached("<Error>")
+
+    override val superTypes: Sequence<KSTypeReference> = emptySequence()
+
+    override val typeParameters: List<KSTypeParameter> = emptyList()
+
+    override fun getSealedSubclasses(): Sequence<KSClassDeclaration> = emptySequence()
+
+    override fun asStarProjectedType(): KSType {
+        TODO()
```
jsjeon, line 62: Nit: seems safe to use `error(...)`, like a message with "Should not be called"? Below too.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
```diff
@@ -29,18 +29,22 @@ import org.jetbrains.kotlin.analysis.api.types.KtFunctionalType
 import org.jetbrains.kotlin.analysis.api.types.KtNonErrorClassType
 import org.jetbrains.kotlin.analysis.api.types.KtType
 import org.jetbrains.kotlin.analysis.api.types.KtTypeNullability
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
-class KSTypeImpl private constructor(private val type: KtType) : KSType {
+class KSTypeImpl private constructor(val type: KtType) : KSType {
```
jsjeon, line 34: Isn't `internal` good enough? I'm afraid it's okay to expose element from Analysis API.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSValueArgumentImpl.kt
neetopia: yes we will get something like you said. the `toString()` method currently tries to align with the compiler plugin implementation, might change it later if this is better representation.

jsjeon (approved):

PR #1030, opened 2022-06-27
Recover `lateinit` modifier for descriptor-based implementations.

Jeffset:
Very similar to what I did with `const` modifier once.

### DIFF
--- a/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/impl/test/KSPAATest.kt
+++ b/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/impl/test/KSPAATest.kt
@@ -322,6 +322,13 @@ class KSPAATest : AbstractKSPAATest() {
     }
 
     @Disabled
+    @TestMetadata("lateinitProperties.kt")
+    @Test
+    fun testLateinitProperties() {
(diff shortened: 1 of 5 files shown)


kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/impl/test/KSPAATest.kt
```diff
@@ -321,6 +321,13 @@ class KSPAATest : AbstractKSPAATest() {
         runTest("../compiler-plugin/testData/api/javaWildcards2.kt")
     }
 
+    @Disabled
+    @TestMetadata("lateinitProperties.kt")
+    @Test
+    fun testLateinitProperties() {
+        runTest("testData/api/lateinitProperties.kt")
```
neetopia, line 328: should be `"../compiler-plugin/testData/api/lateinitProperties.kt"`

neetopia (approved):

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/impl/test/KSPAATest.kt
Jeffset: Done

neetopia:
Thanks for PR!

PR #1031, opened 2022-06-28
`KotlinType.unwrap()` has to be used for correct `is` checks.

Jeffset:
Sometimes raw types may be wrapped into "enhanced" types by the compiler,
 so the old naive check didn't work on them.

 `UnwrappedType`'s doc states, that it's only safe to use `instanceof` checks on unwrapped types.

I ran into this issue in my closed-source project yet was unable to reproduce it with KSP testing infra,
 so no test changes here.


neetopia (approved):
Thanks for the fix.

PR #1033, opened 2022-07-02
Fix assignable check for function types with deserialized descriptors.

Jeffset:
E.g. a `kotlin.Function0` type is a Kotlin "fictional" type;
the real JVM type it is backed by for JVM is `kotlin.jvm.functions.Function0`.
Inside the compiler the backing type is always converted into Kotlin's one.

Yet for KSP the `kotlin.jvm.functions.Function0` with its deserialized descriptor
 can be obtained directly (e.g. via `getJavaClassByName`) and used along with
 the `kotlin.Function0` type.

This CL fixes the assignability check between such types.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
@@ -507,9 +507,11 @@ internal val KSFunctionDeclaration.jvmAccessFlag: Int
 // Compiler subtype checking does not convert Java types to Kotlin types, while getting super types
 // from a java type does the conversion, therefore resulting in subtype checking for Java types to fail.
 // Check if candidate super type is a Java type, convert to Kotlin type for subtype checking.
+// Also, if the type is a generic deserialized type, that actually represents a function type,
+// a conversion is also required to yield a type with a correctly recognised descriptor.
 internal fun KotlinType.convertKotlinType(): KotlinType {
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
```diff
@@ -528,3 +530,16 @@ internal fun KotlinType.convertKotlinType(): KotlinType {
         upperBound
     )
 }
+
+private fun ClassifierDescriptor.shouldMapToKotlinForAssignabilityCheck(): Boolean {
```
neetopia, line 534: you can make it extension of nullable type to avoid null check before calling this. Alternatively, you can use `?.foo() == true`

neetopia (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/utils.kt
Jeffset: I did consider this, but then the smart cast for the `classifierDescriptor` will not be possible and a redundant `!!` will be required.

Jeffset: Changed to `declarationDescriptor?.shouldMapToKotlinForAssignabilityCheck() == true`

Jeffset:
@neetopia PTAL

neetopia:
Oops, missed previous notification. Thanks for your PR!

PR #1041, opened 2022-07-13
A new CodeGenerator.createFile method with more flexibility

jameskleeh:
Problem:
The current API assumes classes are being created by enforcing package name and extension arguments. Many use cases for creating files include creating resources. This requires passing in a dummy "package name" and an empty extension in many cases. The processing of the package name argument assumes that dots should be replaced with slashes to represent separate directories. This prevents the creation of directories with dots in their name.

Solution:
This PR maintains the previous behavior with the existing API and introduces a new method to allow users more flexibility in the path being created. Suggestions welcome


neetopia (commented):
Overall this looks good, but I would refrain from introducing a new FileType enum, this is helpful but will make the 2 `createNewFile()` apis inconsistent in terms of the way they handle file types. If necessary we can introduce FileType later in a uniformed way for all APIs, for now just checking the extension name should help, unless we have a use case for example, generating `.java` files into resource directory?

jameskleeh:
@neetopia Having to pass in an extension as a mechanism to choose a directory doesn't make sense in the cases where there is no file extension. In addition, if the fileType was changed to `extensionName: String`, wouldn't that clash with the existing API where the last argument has a default value?

neetopia:
I am not completely against the idea of `FileType`, but do not want to make it consistent for 2 file creation APIs, and change the existing one will be a breaking change therefore we want to evaluate it later to see if we want to go that approach. 

That being said, even if we want to use `FileType` for the existing file creation API, it will still cause same signature clash issue, to avoid signature clash, an option is to name the function differently.

jameskleeh:
@neetopia I've made the changes we discussed on slack

api/src/main/kotlin/com/google/devtools/ksp/processing/FileType.kt
```diff
@@ -0,0 +1,12 @@
+package com.google.devtools.ksp.processing
+
+/**
+ * Used to determine where files should be stored when being
+ * created through the [CodeGenerator].
+ */
+enum class FileType {
```
neetopia, line 7: This `FileType` should be unnecessary then? I see that you do have usages in your code relying on this enum, but since we do not have any APIs exposing this enum, maybe we can put it in our implementations, if this enum helps with your implementation.

neetopia (commented):
Can you please fix the lint error and the failing test?

api/src/main/kotlin/com/google/devtools/ksp/processing/FileType.kt
jameskleeh: Yeah, got rid of it

jameskleeh:
@neetopia Tests are passing

neetopia:
you can run `./gradlew ktlint` to see the failure on lint.

jameskleeh:
@neetopia The lint issue is resolved

neetopia (approved):
LGTM, please fix the api compatibility check so I can merge this PR.

neetopia:
Merged, thanks for your PR!

PR #1053, opened 2022-08-02
AA: use AAsession, test infra fix, error type fix

neetopia:
I am open to suggestions on the error type test. Basically AA is capable of resolving a function call with error argument, which actually makes more sense to me, but at this point it might not make a lot of sense to fix old frontend (assume changes in upstream). 

An alternative is to have 2 versions of test result for comparing, as we are expecting more inconsistency between old and new frontend, this also makes sense in terms of long term investment.

### DIFF
--- a/compiler-plugin/testData/api/errorTypes.kt
+++ b/compiler-plugin/testData/api/errorTypes.kt
@@ -32,7 +32,7 @@
 // END
 // FILE: a.kt
 class C {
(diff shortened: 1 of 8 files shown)


ting-yuan (approved):

jsjeon:
Good to see that KSP is switched to use Analysis API session/builder. 🎉 

cc @darthorimar

Issue #1063, opened 2022-08-13
Memory leak: org.sqlite.JDBC holds classloader of KSP


ting-yuan:
This happens when running Room:
1. Kotlin compiler daemon loads compiler plugins (KSP) with a new class loader every time it gets a request from Gradle. 
2. KSP loads processors(Room) with a child classloader.
3. Room uses `org.sqlite.JDBC`, which is also loaded with Room's classloader.
4. `org.sqlite.JDBC`'s static initializer calls `java.sql.DriverManager.registerDriver`, which stores the class `org.sqlite.JDBC` in a static member of `java.sql.DriverManager`.
6. `java.sql.DriverManager` is loaded by the root classloader.
7. Therefore all everything loaded with KSP's classloader can also be reached by the root classloader and never be GC'ed.

Here is [a sample project](https://github.com/BugsBunnyBR/KspMemory) thanks to @BugsBunnyBR.

![Screenshot 2022-08-13 01 37 09](https://user-images.githubusercontent.com/3227560/184476092-9f308d6f-fcc3-4245-8aa2-a54f6a00cdcc.png)

BugsBunnyBR:
@ting-yuan @neetopia 
Thank you for your analysis and mitigation efforts!
It seems that more could be done on Room side of things, so I opened [this issue](https://issuetracker.google.com/issues/242575306) on Room issue tracker. Maybe you folks could contribute with your thoughts there ?

BugsBunnyBR:
Should we try to open an issue on Sqlite too?

ting-yuan:
This is fixed in Room already. It is also mitigated in #1067 in case there are other processors with similar issue.

BugsBunnyBR:
To the people watching this issue...

I ran the benchmark in a [branch with the latest version of room ](https://github.com/BugsBunnyBR/KspMemory/tree/ksp-room-alpha) and these are the results.
![Screenshot 2022-08-25 at 15 49 40](https://user-images.githubusercontent.com/70573/186682916-237cbd86-46b8-4c62-aa7d-fd2e3fa2ff7e.png)

This is way better than before, even though it seems to use lots of memory I was able to run 50 iterations and didn't get an out of memory. 👏 👏 
I will be waiting a version of ksp to verify the final results.

BugsBunnyBR:
[...]

PR #1067, opened 2022-08-15
A mitigation for processors with memory leaks

ting-yuan:
by keeping static variables as few as possible, and set them to null after each round.

See #1063 for details.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
@@ -221,7 +221,10 @@ abstract class AbstractKotlinSymbolProcessingExtension(
                             findTargetInfos(module)
                         )
                     )
-                }?.let { analysisResult -> return@doAnalysis analysisResult }
+                }?.let { analysisResult ->
+                    resolver.tearDown()
+                    return@doAnalysis analysisResult
+                }
                 if (logger.hasError()) {
                     return@mapNotNull null
                 }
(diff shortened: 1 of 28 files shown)


neetopia (commented):
Have you verified it with the reproduce project? 

What I didn't get is that, in #1063 it says the the reference to the processors comes from `instance` object, however the processor loading happens in `KotlinSymbolProcessingExtension`.

Can you elaborate how un-assigning `instance` for `ResolverImpl` can help with this case when the actual reference to processors are in `KotlinSymbolProcessingExtension`?

ting-yuan:
Yes it does help. Without this it'd OOM at ~30th iteration. Also verified with VisualVM and the leak is much slower now.

`ResolverImpl.instance` holds all the objects that are reachable from `ResolverImpl`. Cutting it makes most of the compiler data structures (indirectly) used by KSP unreachable.

There is no reference in static fields of `KotlinSymbolProcessingExtension`. Objects referenced by the instance of `KotlinSymbolProcessingExtension` should become collectible once the compilation finishes, because nothing holds the instance. Non-static fields are collectable.

The only thing left un-collectable are classes themselves (which are not trivial unfortunately), as well as objects reachable from static fields / companion objects.

neetopia (approved):
I get your point on the reference to compiler data structures. Just to confirm, does this PR solves the classloader issue caused by Room? My question was mainly on that part since the linked issue mentioned Room in particular.

ting-yuan:
It mitigates it. This PR makes most of the objects in KSP and kotlinc collectable. There is still leak, but much smaller now. The thorough fix needs to be done in Room or more likely sqlite. This PR is the best we can do in KSP in case of leaking processors like this.

PR #1073, opened 2022-08-19
AA: add visibility check for getAllFunctions/getAllProperties 

neetopia:
* getAllSymbols returns all symbols including invisible ones, check vibility for the target class to filter out unneeded symbols.
* Include a temporary workaround for checking containing class due to FIR compiler limitation.
* unmutes allFunctions_kotlin allFunctions_java_inherits_kt test.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
@@ -28,6 +28,7 @@ import org.jetbrains.kotlin.analysis.api.analyze
 import org.jetbrains.kotlin.analysis.api.annotations.KtAnnotated
 import org.jetbrains.kotlin.analysis.api.annotations.annotations
 import org.jetbrains.kotlin.analysis.api.lifetime.KtAlwaysAccessibleLifetimeTokenFactory
+import org.jetbrains.kotlin.analysis.api.symbols.KtClassOrObjectSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KtEnumEntrySymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KtFunctionLikeSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KtJavaFieldSymbol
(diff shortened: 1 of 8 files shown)


neetopia:
also added missing `@TestMetadata` annotation for `abstractFunctions.kt`, somehow this test is missing this annotation.

ting-yuan (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
```diff
@@ -48,7 +49,8 @@ internal val ktSymbolOriginToOrigin = mapOf(
     KtSymbolOrigin.JAVA_SYNTHETIC_PROPERTY to Origin.SYNTHETIC,
     KtSymbolOrigin.INTERSECTION_OVERRIDE to Origin.KOTLIN,
     // TODO: distinguish between kotlin library and java library.
-    KtSymbolOrigin.LIBRARY to Origin.JAVA_LIB
+    KtSymbolOrigin.LIBRARY to Origin.KOTLIN_LIB,
+    KtSymbolOrigin.SUBSTITUTION_OVERRIDE to Origin.JAVA_LIB
```
jsjeon, line 53: You can't tell the origin of (type) substituted override unless you check the original symbol.

See the example I added at https://github.com/JetBrains/intellij-community/commit/0c4ec1476ec22cab5ee8fbfb50fb535ddc3073af. The fake override added to `SubBox` where type parameter `t` is substituted with `String` will have `SUBSTITUTION_OVERRIDE`. And both of `SubBox` and `Box<T>` are in kotlin source in this case.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
```diff
@@ -99,6 +101,10 @@ internal fun KtSymbolWithMembers.declarations(): Sequence<KSDeclaration> {
 internal fun KtSymbolWithMembers.getAllProperties(): Sequence<KSPropertyDeclaration> {
     return analyze {
         this@getAllProperties.getMemberScope().getCallableSymbols()
+            .filter {
+                it.isVisibleInClass(this@getAllProperties as KtClassOrObjectSymbol) ||
```
jsjeon, line 105: This one is `KtOverrideInfoProvider#isVisible`, which is checking if the given (possibly inherited) member symbol is visible in the given (possibly subtype) class. They're originally used by FIR checker, `FirNotImplementedOverrideChecker`, and even didn't care containing class membership.

I think your intention is rather `FirVisibilityChecker#isVisible` (note that that util has an option to `skipCheckForContainingClassVisibility` for simplicity!), which is wrapped as `KtVisibilityChecker#isVisible`.

PR #1083, opened 2022-08-29
AA: abstraction for converting KSTypeArgument to KtTypeArgument

neetopia:
unmutes 2 tests along the way.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
@@ -77,7 +77,11 @@ class KSClassDeclarationImpl private constructor(private val ktClassOrObjectSymb
     }
 
     override fun asType(typeArguments: List<KSTypeArgument>): KSType {
-        TODO("Not yet implemented")
+        return analyze {
+            analysisSession.buildClassType(ktClassOrObjectSymbol) {
+                typeArguments.forEach { argument(it.toKtTypeArgument()) }
+            }.let { KSTypeImpl.getCached(it) }
+        }
     }
 
     override fun asStarProjectedType(): KSType {
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
@@ -79,6 +79,6 @@ class KSTypeReferenceImpl(private val ktType: KtType) : KSTypeReference {
         }
 
     override fun toString(): String {
-        return ktType.toString().substringAfterLast('/')
+        return ktType.render()
     }
 }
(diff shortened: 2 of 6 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -77,7 +77,11 @@ class KSClassDeclarationImpl private constructor(private val ktClassOrObjectSymb
     }
 
     override fun asType(typeArguments: List<KSTypeArgument>): KSType {
-        TODO("Not yet implemented")
+        return analyze {
+            analysisSession.buildClassType(ktClassOrObjectSymbol) {
```
ting-yuan, line 81: Is `buildClassType` cached? The type arguments are not cached so it looks like the type isn't cached as well. Do two different instances compare to equal?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
```diff
@@ -71,7 +72,11 @@ class KSTypeReferenceImpl(private val ktType: KtType) : KSTypeReference {
     }
 
     override val modifiers: Set<Modifier>
-        get() = TODO("Not yet implemented")
+        get() = if (ktType is KtFunctionalType && ktType.isSuspend) {
+            setOf(Modifier.SUSPEND)
```
ting-yuan, line 76: Can we leave a todo in the comment to address other possibilities? Looks like it only handled a special case for the test.

neetopia: We only treat `SUSPEND` modifier for type reference in FE1.0 implementation.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
neetopia: calling with same argument will result in same return result by reference. But it is not documented behavior, might be subject to change in upstream, let me think about how to cache for this.

ting-yuan: I'm afraid that it is always called with different arguments: `KSTypeArgument.toKtTypeArgument()` always returns new instances.

neetopia: it's true that the result for `toKtTypeArgument` returns a different instance, but somehow the end result of `buildClassType` can still return same instance, which was my point in the comment, my guess is that there is some underlying caching that checks for this, therefore different instance can still result in same result if they are equal, it does look odd to me tbh.

btw the implementation is not cached in FE1.0 as well, which might be yet a bigger concern for us.

ting-yuan: It's fine as long as either `KtType` is cached or handles `equals` / `hashCode`. Can you confirm that this case is covered in our tests before merging? I.e., calling `asType` twice results in instances, same or not, that compare to equal.

ting-yuan (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
neetopia: after some deep dive, my current understanding of how everything works for both implementations:

* for AA, during type creation, AA first looks up class symbol by `ClassId` which will ensure same instance of type constructor, then for the type arguments, AA will unwrap `KtTypeArgument`s to build FIR type arguments, therefore eliminating instance differences for the inputs. Then for the AA types, its `hashCode` and `equals` functions are overridden to use the FIR's `coneKotlinType`'s, which in turn eliminates all input difference. 

* for FE1.0, types related computations seems to go through a storage manager which is in turn a global cache therefore cached automatically.

ting-yuan (approved):

PR #1085, opened 2022-08-30
UPDATE_KOTLIN_VERSION: 1.8.0-dev-2355

ting-yuan:
* implement LookupTracker.clear()
* filter `entries` out in enums

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
@@ -27,11 +27,13 @@ import com.google.devtools.ksp.symbol.impl.kotlin.*
 import com.google.devtools.ksp.symbol.impl.replaceTypeArguments
 import com.intellij.psi.PsiField
 import com.intellij.psi.PsiMethod
+import org.jetbrains.kotlin.builtins.StandardNames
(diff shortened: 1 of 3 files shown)


ting-yuan:
More context for Enum.entries: https://youtrack.jetbrains.com/issue/KT-48872

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
```diff
@@ -97,7 +98,8 @@ class KSClassDeclarationDescriptorImpl private constructor(val descriptor: Class
     override val declarations: Sequence<KSDeclaration> by lazy {
         sequenceOf(
             descriptor.unsubstitutedMemberScope.getDescriptorsFiltered(),
-            descriptor.staticScope.getDescriptorsFiltered(),
+            // FIXME: Support static, synthetic `entries` for enums when the language feature is enabled.
+            descriptor.staticScope.getDescriptorsFiltered { it.identifier != StandardNames.ENUM_ENTRIES.identifier },
```
neetopia, line 102: what if a non enum class has a static declaration with name `entries`?

ting-yuan: Good catch! Updated.

neetopia (approved):

PR #1088, opened 2022-08-31
change the modifiers to private from public and lift return out of 'if'

heli-os:
Hi i am Heli who using ksp.

~First, changed some codes to not use the platform type. platform type requires unnecessary boilerplate(!!, let trick, ...) And open source is a guide to writing code.~

~so i think bad code examples should be deleted.~

Second, Lift return out of 'if'

Third, change the modifiers to private from public

thanks. 🙇

--
There is no change that users can feel.


heli-os:
Can anyone please review it?

neetopia:
I think the change for combining return statement and private val make sense to me, but I don't get the benefit of adding explicit types to the rest of your change, if no specific reasons, maybe let's not touch them.

heli-os:
@neetopia 
okay, It would be better not to touch it.

I removed the added explicit type.

neetopia:
Thanks for your PR!

PR #1101, opened 2022-09-09
retrieve annotations for KSType from KotlinType instead of KSTypeReference

### LINKED ISSUE #985: No "kotlin.ExtensionFunctionType" annotation on functional types with receivers
When you have a `KSType` that represents a functional type with a receiver (e.g. `String.() -> Unit`), and you are trying to get the list of all annotations of the type, there is no `@kotlin.ExtensionFunctionType`. **But**, `functionalType.toString()` call returns this:
```
"[@kotlin.ExtensionFunctionType] Function1<String, Unit>"
```
In the debugger, I see this:

> 

Somehow, this annotation is omitted by ksp.

However, without it seems **impossible** to determine, whether any functional type has a receiver or not.
Maybe there is another way?

I created a [branch](https://github.com/popovanton0/kira/tree/ksp_issue_with_functional_types) in my [repo](https://github.com/popovanton0/kira) that reproduces this behaviour. To reproduce, run [this](https://github.com/popovanton0/kira/blob/ksp_issue_with_functional_types/kira-processor/src/test/java/com/popovanton0/kira/processor/KiraProcessorTest.kt) unit test. Input to the test is [here](https://github.com/popovanton0/kira/blob/ksp_issue_with_functional_types/kira-processor/src/test/resources/KiraProcessorTest/no%20extension%20annotation%20of%20functional%20type/input/test.kt). Code of the processor is [here](https://github.com/popovanton0/kira/blob/ksp_issue_with_functional_types/kira-processor/src/main/java/com/popovanton0/kira/processing/KiraProcessor.kt).

neetopia:
fixes #985 

several open questions:

* defining annotations origin for KSType is tricky, as type is not part of source code. with this PR actually all annotation origins from KSType becomes library, which I actually found more meaningful, but this can be breaking change.
*  since `KotlinType.annotations` contains the full picture of the actual annotations of a kotlin type, the `annotations` parameter in `getKSTypeCached` is no longer necessary, might want to clean up in a future patch.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
@@ -25,6 +25,7 @@ import com.google.devtools.ksp.symbol.KSType
 import com.google.devtools.ksp.symbol.KSTypeArgument
 import com.google.devtools.ksp.symbol.Nullability
 import com.google.devtools.ksp.symbol.Origin
+import com.google.devtools.ksp.symbol.impl.binary.KSAnnotationDescriptorImpl
 import com.google.devtools.ksp.symbol.impl.binary.KSTypeArgumentDescriptorImpl
 import com.google.devtools.ksp.symbol.impl.convertKotlinType
 import com.google.devtools.ksp.symbol.impl.replaceTypeArguments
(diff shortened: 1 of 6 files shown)


ting-yuan:
IIRC the first point you mentioned is the reason why the annotations are passed in. Another reason is that it was a (premature)  optimization. To keep the behavior untouched, there are two approaches:
1. Fix call sites of `getKSTypeCached`.
2. Try to find PSIs for those annotations.

neetopia:
sounds good, updated to not add from descriptor for annotations with actual source.

ting-yuan (approved):

PR #1119, opened 2022-09-24
Support running KSP via KotlinCompilerRunner

ting-yuan:
which is based on KGP's GradleCompilerRunnerWithWorker.

Also refactored the code to allow selection between the old and new
ways to create KSP tasks and call the compiler.


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunner.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.gradle.api.tasks.Internal
+import org.jetbrains.kotlin.cli.common.arguments.K2JSCompilerArguments
+import org.jetbrains.kotlin.cli.common.arguments.K2JVMCompilerArguments
+import org.jetbrains.kotlin.cli.common.arguments.K2MetadataCompilerArguments
+import org.jetbrains.kotlin.cli.common.arguments.K2NativeCompilerArguments
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompilerExecutionStrategy
+import org.jetbrains.kotlin.gradle.utils.newInstance
+import java.io.File
+
+interface KotlinCompilerRunner {
```
gavra0, line 37: Do you actually needs these properties? My understanding is that you only need
```
fun runJvmCompilerAsync(
        args: K2JVMCompilerArguments,
        sources: List<File>,
        commonSources: List<File>,
        outputs: List<File>
    )
```

In fact, you probably want to keep the same ones as KGP in order to ensure Kotlin Compiler daemon is reused.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunner.kt
```diff
@@ -0,0 +1,113 @@
...
+fun createKotlinJvmCompilerRunner(
+    task: Task,
+    objectFactory: ObjectFactory,
+): KotlinJvmCompilerRunner {
+    return objectFactory.newInstance<KotlinJvmCompilerRunnerImpl>(task)
+}
+
+fun createKotlinJsCompilerRunner(
+    task: Task,
+    objectFactory: ObjectFactory,
```
gavra0, line 96: Just passing the task should be enough from the API point of view.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunner.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Classpath
+import org.gradle.api.tasks.Internal
+import org.jetbrains.kotlin.cli.common.arguments.K2JSCompilerArguments
+import org.jetbrains.kotlin.cli.common.arguments.K2JVMCompilerArguments
+import org.jetbrains.kotlin.cli.common.arguments.K2MetadataCompilerArguments
+import org.jetbrains.kotlin.cli.common.arguments.K2NativeCompilerArguments
```
gavra0, line 32: This requires moving arguments to `-api` module, we need to check if that's possible and if they match the API guarantees.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunnerImpls.kt
```diff
@@ -0,0 +1,233 @@
...
+internal inline fun <reified T : Any?> ObjectFactory.property() = property(T::class.java)
+internal inline fun <reified T : Any?> ObjectFactory.property(initialValue: T) = property<T>().value(initialValue)
+
+abstract class KotlinCompilerRunnerImpl @Inject constructor(
+    task: Task,
+    objectFactory: ObjectFactory,
+    @get:Internal val workerExecutor: WorkerExecutor
+) : KotlinCompilerRunner {
+    @get:Internal
+    internal val taskProvider: Provider<GradleCompileTaskProvider> = objectFactory.property(
```
gavra0, line 60: I'm positively surprised that `GradleCompileTaskProvider` does not require Kotlin-specific task type, only Gradle Task is enough :)

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunnerImpls.kt
```diff
@@ -0,0 +1,233 @@
...
+        objectFactory.newInstance<GradleCompileTaskProvider>(task.project.gradle, task, task.project)
+    )
+
+    @get:Internal
+    internal val defaultKotlinJavaToolchain: Provider<DefaultKotlinJavaToolchain> =
+        objectFactory.propertyWithNewInstance({ null })
+
+    @get:Internal
+    internal val metrics: Property<BuildMetricsReporter> = objectFactory.property(BuildMetricsReporterImpl())
```
gavra0, line 69: Once this get's to KGP, all of these should be configured by the KGP (from its DSL/API etc.).

gavra0 (commented):
Overall, it seems ok. The API surface is not too big:
- given a task you'll get a compiler runner
- compiler runner should allow you to specify compiler args and input/output/common sources

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunner.kt
ting-yuan: Will add TODOs to remove them when getting into KGP. They are needed for now to allow KSP copy those settings from the compilation task.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunner.kt
ting-yuan: Will address in upstream patches. I'll probably also move those functions into proper KGP classes or factories.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunnerImpls.kt
ting-yuan: Will add TODOs in a new commit.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunner.kt
ting-yuan: Good catch. I'll create a simplified / public version of those.

ting-yuan: Addressed with a simplified version of the arguments in 8782e89.

Tapchicoma: these particular flags are the part of compiler and will never be in KGP-api artifact. On the other side - we are exposing currently `kotlinOptions` and from `1.8.0` - `compilerOptions` generated DSL. And we have plans to add more missing options to `compilerOptions` in followup releases.

Probably we should think about another approach that will not require to use Kotlin compiler arguments from KSP side. Also this arguments are not an official compiler api and could have breaking changes between minor releases.

ting-yuan:
The proposed `KotlinCompilerArguments` mostly comes from `kotlinOptions`, `compilerOptions` (compiler args annotated with `GradleOptions`) and fields in `BaseKotlinCompile` and `KotlinCompileTool`:

```
    // From compilerOptions / kotlinOptions:
    var verbose: Boolean = false
    var allWarningsAsErrors: Boolean = false
    var languageVersion: String? = null
    var apiVersion: String? = null
    var useK2: Boolean = false
    var multiPlatform: Boolean = false
    var moduleName: String? = null
    var jvmTarget: String? = null // JVM
    var noJdk: Boolean = false // JVM
    var noStdlib: Boolean = false // JS
    var target: String = "v5"  // JS
    var target: String? = null // K/N

    // From BaseKotlinCompile / KotlinCompileTool / KotlinTaskConfigs.kt:
    var friendPaths: List<File> = emptyList()
    var libraries: List<File> = emptyList()
    var destination: File? = null
    var pluginOptions: List<String> = emptyList()
    var pluginClasspaths: List<File> = emptyList()
```

With a few exceptions. They are unlikely to change and may be stable enough to get into the KGP-api artifact.
```
    // JVM
    var noStdlib: Boolean = false
    var noReflect: Boolean = false
    var allowNoSourceFiles: Boolean = false
```

**Proposal**
1. Add the above JVM options (`noStdlib`, `noReflect`, `allowNoSourceFiles`) into `CompilerJvmOptions`. And replace `KotlinCompilerArguments` (defined in this patch) with `compilerOptions`. Because `CompilerCommonOptions` / `CompilerJvmOptions` are configurable in Kotlin DSL, it should make sense to use it to configure / setup the custom compilation.
2. Require the entry point in the proposed (e.g., `createKotlinJvmCompilerRunner`) API to accept a subtype of `BaseKotlinCompile` or `KotlinJvmCompile`, so that `runJvmCompilerAsync` can read the options defined in `KotlinJvmCompile` / `BaseKotlinCompile`.
3. Add `freeArgs` to `runJvmCompilerAsync`. KSP needs this to pass modified / removed files in its plugin options at execution time. Also it can cover some of the unstable compiler arguments (at users' discretion). Please see comments below.

Admittedly, there are still a few options that may be subject to change in the future. It might be better to handle those in `freeArgs`, at users' risk / discretion, until they become stable.
```
    // Used to select between the new and old expect actual linker
    var expectActualLinker: Boolean = false

    // These are used to select among JS backends.
    // In fact, in KGP, they are never set up; The free format args, e.g. -Xir-only are used instead.
    var irOnly: Boolean = false
    var irProduceJs: Boolean = false
    var irProduceKlibDir: Boolean = false
    var irProduceKlibFile: Boolean = false
    var irBuildCache: Boolean = false
    var wasm: Boolean = false
```

To summarize, with the newly proposed `KotlinCompilerRunner`:
1. It accepts subtypes of `BaseKotlinCompile` and returns an instance of `KotlinCompilerRunner`, which has a member function `run<target>CompilerAsync`.
2. Users (e.g., KSP) fill `CompilerCommonOptions` / `CompilerJvmOptions` / etc and configure `BaseKotlinCompile`, and call `run<target>CompilerAsync` with the options, `freeArgs` and sources / destination.

@Tapchicoma what do you think? The only new interface introduced is `KotlinCompilerRunner`. Nothing new around compiler args will need to be maintained.

ting-yuan:
@Tapchicoma I've uploaded an implementation that makes use of `compilerOptions` instead of `CommonCompilerArguments`. The use of `freeArgs` is limited and most of them can be added to the API.
```
"-Xallow-no-source-files", "-no-stdlib", "-Xmulti-platform", "-Xexpect-actual-linker",
"-Xplugin=...", "-P plugin:..."
```

Currently I have to copy some free args from JS compilation task to select the backend and filter the libraries. Because `IR_LEGACY` is being deprecated, these will go away soon:
```
"-Xir-only", "-Xir-produce-js", "-Xir-produce-klib-file", "-Xir-produce-klib-dir", "-Xir-build-cache", "-Xwasm"
```

Paths are moved into the parameters of `run*CompilerAsync`. The API interface now becomes, for example,
```
interface KotlinJvmCompilerRunner : KotlinCompilerRunner {
    fun runJvmCompilerAsync(
        options: CompilerJvmOptions,
        freeArgs: List<String>,
        sources: List<File>,
        commonSources: List<File>,
        friendPaths: List<File>,
        libraries: List<File>,
        outputs: List<File>,
        destination: File
    )
}

fun createKotlinJvmCompilerRunner(
    task: Task,
    objectFactory: ObjectFactory,
): KotlinJvmCompilerRunner {
    return objectFactory.newInstance<KotlinJvmCompilerRunnerImpl>(task)
}
```

gradle-plugin/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ plugins {
 dependencies {
     implementation("org.jetbrains.kotlin:kotlin-gradle-plugin-api:$kotlinBaseVersion")
     implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinBaseVersion")
+    implementation("org.jetbrains.kotlin:kotlin-compiler-runner:$kotlinBaseVersion")
```
Tapchicoma, line 27: KGP has `compileOnly` dependency on this library as all classes should come from `kotlin-compiler-embeddable` on runtime. Probably you should do the same

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/CompilerOptionsFactory.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.jetbrains.kotlin.gradle.dsl.CompilerJsOptions
+import org.jetbrains.kotlin.gradle.dsl.CompilerJsOptionsDefault
+import org.jetbrains.kotlin.gradle.dsl.CompilerJvmOptions
+import org.jetbrains.kotlin.gradle.dsl.CompilerJvmOptionsDefault
+import org.jetbrains.kotlin.gradle.utils.newInstance
+
+// TODO: to be replaced by KotlinJvmFactory, etc.
+class CompilerOptionsFactory {
+    companion object {
+        fun createCompilerJvmOptions(objectFactory: ObjectFactory): CompilerJvmOptions =
```
Tapchicoma, line 34: If you are targeting Kotlin 1.8.0 - you could already use `KotlinJvmFactory` to create `CompilerJvmOptions`

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/CompilerOptionsFactory.kt
```diff
@@ -0,0 +1,43 @@
...
+            objectFactory.newInstance<CompilerJvmOptionsDefault>()
+
+        fun createCompilerJsOptions(objectFactory: ObjectFactory): CompilerJsOptions =
+            objectFactory.newInstance<CompilerJsOptionsDefault>()
+
+        fun createCompilerCommonOptions(objectFactory: ObjectFactory): CompilerCommonOptions =
+            objectFactory.newInstance<CompilerCommonOptionsDefault>()
```
Tapchicoma, line 41: For `KotlinJsFactory` and `KotlinCommonFactory` please open new issues, possibly it is still ok to add them into `1.8.0`

Tapchicoma: If you need same way to create for soon-to-added `CompilerNativeOptions`interface  - pleases create an issue as well

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinCompilerRunnerImpls.kt
Tapchicoma: Because it is declared on `AbstractKotlinCompile` and is used to provide generic compilation configuration that is same for all targets :slightly_smiling_face:

Tapchicoma:
> Add freeArgs to runJvmCompilerAsync. KSP needs this to pass modified / removed files in its plugin options at execution time. Also it can cover some of the unstable compiler arguments (at users' discretion). Please see comments below.

FYI: KGP will go away from `freeCompilerArgs` and try to provide all user-required via `compilerOptions`.

jaschdoc:
@ting-yuan, can I close this PR or is it still relevant? Does it contain work-in-progress features that we want to support in another way?

PR #1120, opened 2022-09-25
Bump Gradle wrapper to 7.5.1

3flex:


neetopia:
Thanks for the PR, however, we intend to upgrade gradle versions only when needed.

PR #1123, opened 2022-09-25
Commit kotlin-js-store

3flex:
This is recommended practice for projects using Kotlin/JS: https://kotlinlang.org/docs/js-project-setup.html#version-locking-via-kotlin-js-store


neetopia:
The example is mainly for showing KSP usage, not for KMP in general.

PR #1126, opened 2022-09-27
Block other compiler plugins by default

### LINKED ISSUE #897: Compatibility with Kotlin/Native compiler plugins
We are [seeing](https://github.com/rickclephas/KMP-NativeCoroutines/issues/50) a similar error as in #207 where required plugin options aren't provided to the compiler plugin by KSP.
Looking at that issue this should be fixed by #270, though it looks like that fix isn't applied to the [KspTaskNative](https://github.com/ting-yuan/ksp/blob/37fd5af015dd7292fdc12d648ea1b83e627d4b34/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L783) task.

ting-yuan:
and deprecate `blockOtherPlugins`.

Fixes #897 

There's still no real test yet because no built-in plugin runs before `AnalysisHandlerExtension`; The existing test only checks whether it builds.

### DIFF
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/PlaygroundIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/PlaygroundIT.kt
@@ -59,7 +59,7 @@ class PlaygroundIT {
         val gradleRunner = GradleRunner.create().withProjectDir(project.root)
 
         File(project.root, "workload/build.gradle.kts")
-            .appendText("\nksp {\n  blockOtherCompilerPlugins = true\n}\n")
+            .appendText("\nksp {\n  blockOtherCompilerPlugins = false\n}\n")
         gradleRunner.buildAndCheck("clean", "build")
         gradleRunner.buildAndCheck("clean", "build")
         project.restore("workload/build.gradle.kts")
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/PlaygroundIT.kt
```diff
@@ -59,7 +59,7 @@ class PlaygroundIT {
         val gradleRunner = GradleRunner.create().withProjectDir(project.root)
 
         File(project.root, "workload/build.gradle.kts")
-            .appendText("\nksp {\n  blockOtherCompilerPlugins = true\n}\n")
+            .appendText("\nksp {\n  blockOtherCompilerPlugins = false\n}\n")
```
neetopia, line 62: what is this for?

neetopia (approved):

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/PlaygroundIT.kt
ting-yuan: I'd like to deprecate the flag gradually. The default becomes true effectively. Let's not break `blockOtherCompilerPlugins = false` too soon.

PR #1130, opened 2022-09-29
AA: type alias and reference elements.

neetopia:
* Handled type aliases better 
* Added parent information to some non source based KtSymbols 
* Added more rendering logic for types
* Use the correct object cache for type references.

This is not a final implementation for some of the added classes, some of the TODOs will be addressed in a follow up PR, to avoid too large PR.

TODOs:
* should use `KtTypeReference` as the wrapping element for kotlin source based type references.
* AA actually returns type annotations in more scenarios compared to FE 1.0 causing nestedClassType test to fail, need to address it in a way that tests framework should be able to handle 2 sets of expected results.
* The current ClassifierReference implementation still has some way to go in terms of correctness.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
@@ -27,10 +27,7 @@ import com.google.devtools.ksp.symbol.KSVisitor
 import com.google.devtools.ksp.symbol.Location
 import com.google.devtools.ksp.symbol.Modifier
 import com.google.devtools.ksp.symbol.Origin
-import org.jetbrains.kotlin.analysis.api.types.KtClassErrorType
-import org.jetbrains.kotlin.analysis.api.types.KtFunctionalType
-import org.jetbrains.kotlin.analysis.api.types.KtNonErrorClassType
-import org.jetbrains.kotlin.analysis.api.types.KtType
+import org.jetbrains.kotlin.analysis.api.types.*
 
 class KSTypeReferenceImpl(private val ktType: KtType) : KSTypeReference {
     companion object : KSObjectCache<KtType, KSTypeReference>() {
@@ -43,10 +40,7 @@ class KSTypeReferenceImpl(private val ktType: KtType) : KSTypeReference {
         // TODO: non exist type returns KtNonErrorClassType, check upstream for KtClassErrorType usage.
         return if (
             analyze {
-                ktType is KtClassErrorType || (
-                    (ktType is KtNonErrorClassType) &&
-                        ktType.classId.toKtClassSymbol() == null
-                    )
+                ktType is KtClassErrorType || (ktType.classifierSymbol() == null)
             }
         ) {
             KSErrorType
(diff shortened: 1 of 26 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
```diff
@@ -234,3 +214,22 @@ internal fun ClassId.toKtClassSymbol(): KtClassOrObjectSymbol? {
...
+            is KtFunctionalType -> classSymbol
+            is KtUsualClassType -> classSymbol
+            is KtDefinitelyNotNullType -> original.classifierSymbol()
+            is KtDynamicType -> null
+            is KtFlexibleType -> lowerBound.classifierSymbol()
+            is KtIntegerLiteralType -> null
+            is KtIntersectionType -> null
+        }
+    } catch (e: KotlinExceptionWithAttachments) {
+        null
```
ting-yuan, line 276: Can you leave a comment to explain why muting the exception here?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
```diff
@@ -234,3 +214,22 @@ internal fun ClassId.toKtClassSymbol(): KtClassOrObjectSymbol? {
...
+            is KtTypeParameterType -> this.symbol
+            is KtCapturedType -> TODO("fix in upstream")
+            is KtClassErrorType -> null
+            is KtFunctionalType -> classSymbol
+            is KtUsualClassType -> classSymbol
+            is KtDefinitelyNotNullType -> original.classifierSymbol()
+            is KtDynamicType -> null
+            is KtFlexibleType -> lowerBound.classifierSymbol()
+            is KtIntegerLiteralType -> null
+            is KtIntersectionType -> null
```
ting-yuan, line 273: This is probably overly simplified for `KtFlexibleType` and `KtIntersectionType`, especially for the latter. Can you comment why returning `lowerBound` for the flexible, and make the intersection `TODO()`? Or, comment this function to make it explicit that some types are not supported / not applicable.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
```diff
@@ -234,3 +214,22 @@ internal fun ClassId.toKtClassSymbol(): KtClassOrObjectSymbol? {
         }
     }
 }
+
+internal fun KtType.classifierSymbol(): KtClassifierSymbol? {
+    return try {
+        when (this) {
+            is KtTypeParameterType -> this.symbol
+            is KtCapturedType -> TODO("fix in upstream")
```
ting-yuan, line 265: A comment that explains the TODO would be really useful.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeReferenceImpl.kt
```diff
@@ -38,7 +39,24 @@ class KSTypeReferenceImpl private constructor(
     }
 
     // FIXME: return correct reference element.
```
ting-yuan, line 41: nit: remove the FIXME

ting-yuan (approved):
LGTM, with some nits on comments.

PR #1138, opened 2022-10-06
Scan Kotlin constants for JavaPsiFacade

### LINKED ISSUE #997: Incremental KSP fails in Java/Kotlin mixed module
If we have the following files (a Kotlin annotation, a Java annotated class, and a Kotlin constants file):

`com.example.ann.MyAnn`
```kotlin
annotation class MyAnn(val value: String)
```

`com.example.JavaClass`
```java
public class JavaClass {
    @MyAnn(KotlinConsts.ACTION2)
    public void f() {
    }
}
```

`com.example.KotlinConsts`
```kotlin
class KotlinConsts {
    companion object {
        const val ACTION = "REPLACE"
        const val ACTION2 = "REPLACE"
    }
}
```

With a processor that looks for `MyAnn` and throws an error if the annotation value isn't `"REPLACE"`:
```kotlin
resolver
    .getSymbolsWithAnnotation("com.example.ann.MyAnn")
    .filterIsInstance<KSFunctionDeclaration>()
    .forEach { func ->
        val arg = func.annotations.first().arguments.first().value.toString()
        if (!arg.startsWith("REPLACE"))
            throw IllegalStateException(arg)
    }
```

Then if we run the processor once, KSP runs fine. Then if we change `@MyAnn(KotlinConsts.ACTION2)` to `@MyAnn(KotlinConsts.ACTION)`, and run KSP again, it will run under incremental mode, then we can no longer read the argument value, the condition isn't true, and we throw the `IllegalStateException`.

With the attached project:
[playground3.zip](https://github.com/google/ksp/files/8744036/playground3.zip)
```bash
./gradlew  :workload:kspKotlin
# Edit JavaClass to reference KotlinConsts.ACTION instead
./gradlew  :workload:kspKotlin
# Observe that KSP was not able to read the annotation value
```

This prevents us from enabling incremental mode, which  prevents us from using KSP as without incremental, our time in processing is higher than with Kapt where we can use incremental compilation.

ting-yuan:
so that constants in light classes can be resolved by JavaPsiFacade.

Fixes #997.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
@@ -31,8 +31,14 @@ import com.google.devtools.ksp.processing.impl.NativePlatformInfoImpl
 import com.google.devtools.ksp.processing.impl.ResolverImpl
 import com.google.devtools.ksp.processing.impl.UnknownPlatformInfoImpl
 import com.google.devtools.ksp.symbol.KSAnnotated
+import com.google.devtools.ksp.symbol.KSClassDeclaration
+import com.google.devtools.ksp.symbol.KSDeclarationContainer
 import com.google.devtools.ksp.symbol.KSFile
+import com.google.devtools.ksp.symbol.KSPropertyDeclaration
+import com.google.devtools.ksp.symbol.KSVisitorVoid
+import com.google.devtools.ksp.symbol.Modifier
 import com.google.devtools.ksp.symbol.Origin
+import com.google.devtools.ksp.symbol.Visibility
 import com.google.devtools.ksp.symbol.impl.java.KSFileJavaImpl
 import com.google.devtools.ksp.symbol.impl.kotlin.KSFileImpl
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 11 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -198,6 +204,42 @@ abstract class AbstractKotlinSymbolProcessingExtension(
...
+
+                            override fun visitPropertyDeclaration(property: KSPropertyDeclaration, data: Unit) {
+                                if (property.modifiers.contains(Modifier.CONST)) {
+                                    property.getter // force resolution
+                                }
+                            }
+                        },
+                        Unit
+                    )
+                } catch (_: Exception) {
```
neetopia, line 237: what's the case for exception here and why it is simply ignored for the general `Exception` rather than more specific exception types?

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/Playground3IT.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.Assert
+import org.junit.Assume
+import org.junit.Rule
+import org.junit.Test
+import java.io.File
+
+class Playground3IT {
```
neetopia, line 12: nit: better naming.

neetopia (commented):
Is the issue only relates to incremental? If so maybe we can put a check to see if incremental is enabled before proceeding to the pre-processing.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
ting-yuan: It's similar to the initialization blocks protected by `handleException` in the above. Exceptions can arise from, e.g., broken inputs. This is the best effort to make constants in Kotlin available to JavaPsiFacade and try not to error out if possible. It's up to the processors to decide whether to error out or not.

ting-yuan:
The issue only comes from clean files. The check would be redundant because only clean files are (needed to be) visited.

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/Playground3IT.kt
ting-yuan: Done.

PR #1143, opened 2022-10-11
fix relative path calculation

### LINKED ISSUE #1079: [ksp] java.lang.IllegalArgumentException: this and base files have different roots
KSP didn't respect gradle `buildDir` variable.

Using RAM disk for build fodlers:

**USER_HOME/.gradle/init.gradle**
```
allprojects {
    buildDir="Z:/build/$group/$name/"
}
```

**app/build.gradle**
```
android {
    sourceSets {
        androidTest.assets.srcDirs += files("$projectDir/schemas".toString())

        main.kotlin.srcDirs += '$buildDir/generated/ksp/$name/kotlin/'
        test.kotlin.srcDirs += '$buildDir/generated/ksp/$name/kotlin/'
    }

    buildTypes {
       // https://github.com/InsertKoinIO/koin-annotations
        applicationVariants.all { variant ->
            variant.sourceSets.java.each {
                it.srcDirs += "$buildDir/generated/ksp/${variant.name}/kotlin/"
            }
        }
   }
}
```

But build failed with:

```
e: [ksp] java.lang.IllegalArgumentException: this and base files have different roots: Z:\build\ProjectX\app\generated\ksp\stagingDebug\java\cz\packagename\data\database\dao\UserOauthDao_Impl.java and D:\Projects\company\ProjectX\app.
	at kotlin.io.FilesKt__UtilsKt.toRelativeString(Utils.kt:117)
	at kotlin.io.FilesKt__UtilsKt.relativeTo(Utils.kt:128)
	at com.google.devtools.ksp.processing.impl.CodeGeneratorImpl.associate(CodeGeneratorImpl.kt:119)
	at com.google.devtools.ksp.processing.impl.CodeGeneratorImpl.createNewFile(CodeGeneratorImpl.kt:92)
	at androidx.room.compiler.processing.ksp.KspFiler.createNewFile(KspFiler.kt:106)
	at androidx.room.compiler.processing.ksp.KspFiler.write(KspFiler.kt:41)
	at androidx.room.compiler.processing.XFilerKt.writeTo(XFiler.kt:44)
	at androidx.room.compiler.processing.XFilerKt.writeTo$default(XFiler.kt:43)
	at androidx.room.writer.ClassWriter.write(ClassWriter.kt:75)
	at androidx.room.DatabaseProcessingStep.process(DatabaseProcessingStep.kt:111)
	at androidx.room.DatabaseProcessingStep.process(DatabaseProcessingStep.kt:42)
```

neetopia:
My concern on the `startsWith` check is true, we have to fallback to try catch for better accuracy.

Open to discussion on better approach other than try catch, hence left one place unchanged, should change to try catch if we deciced to proceed with this approach.

fixes #1079

### DIFF
--- a/common-util/src/main/kotlin/com/google/devtools/ksp/processing/impl/CodeGeneratorImpl.kt
+++ b/common-util/src/main/kotlin/com/google/devtools/ksp/processing/impl/CodeGeneratorImpl.kt
@@ -156,11 +156,13 @@ class CodeGeneratorImpl(
     }
 
     private val File.relativeFile: File
-        get() =
-            if (this.startsWith(buildDir))
+        get() {
+            val buildDirPrefix = if (buildDir.path.startsWith("/")) buildDir.path else buildDir.path.replace("\\", "/")
+            return if (this.startsWith(buildDirPrefix))
                 relativeTo(buildDir)
             else
                 relativeTo(projectBase)
+        }
 
     private fun associate(sources: List<KSFile>, outputPath: File) {
         if (!isIncremental)
--- a/common-util/src/main/kotlin/com/google/devtools/ksp/processing/impl/CodeGeneratorImpl.kt
+++ b/common-util/src/main/kotlin/com/google/devtools/ksp/processing/impl/CodeGeneratorImpl.kt
@@ -32,6 +32,7 @@ class CodeGeneratorImpl(
     private val javaDir: File,
     private val kotlinDir: File,
     private val resourcesDir: File,
+    private val projectBase: File,
     private val buildDir: File,
     private val anyChangesWildcard: KSFile,
     private val allSources: List<KSFile>,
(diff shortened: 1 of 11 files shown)


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/PlaygroundIT.kt
```diff
@@ -65,6 +65,28 @@ class PlaygroundIT {
         project.restore("workload/build.gradle.kts")
     }
 
+    @Test
+    fun testArbitraryBuildDir() {
+        Assume.assumeTrue(System.getProperty("os.name").startsWith("Windows", ignoreCase = true))
+        val gradleRunner = GradleRunner.create().withProjectDir(project.root)
+
+        File(project.root, "workload/build.gradle.kts")
+            .appendText("project.buildDir = File(\"D:/build/\")")
```
neetopia, line 74: note that it does seem like there is only one C drive on github windows runners, therefore this test might not be testable on github runners.

common-util/src/main/kotlin/com/google/devtools/ksp/processing/impl/CodeGeneratorImpl.kt
```diff
@@ -154,18 +155,25 @@ class CodeGeneratorImpl(
         associate(sources, file)
     }
 
+    private val File.relativeFile: File
+        get() =
+            if (this.startsWith(buildDir))
```
neetopia, line 160: I found the reason this fails on Windows.

Windows uses counter slash for directory separator, but in Java API it seems to be intepreting the path as URI, which uses normal slash as directory separator, and this causes lots of issues that will break simple string check.

One workaround is to manually replace counter slash with normal slash when doing check, but counter slash is a valid file name character on unix, uncommon but valid, which messes things up again... 

Maybe also put an OS check before checking the strings, but that looks just too bad as a code to me.

PR #1148, opened 2022-10-16
Explicitly set root project name

3flex:
This recommended practice also removes a warning in certain cases involving composite builds.


neetopia (approved):
LGTM, thanks for your PR!

PR #1157, opened 2022-10-24
Compare PSIs by PsiManager.areElementsEquivalent

ting-yuan:
instead of `equals`, for those PSIs returned from
`DeclarationDescripor.findPsi()`. PSIs from descriptors are different instances than PSIs from KSFileJavaImpl. Although they are structurally equivalent, `equals` returns false.


neetopia (commented):
rather than overriding equals check, how about replace with the correct PSI at the result result of `findPsi()`?

ting-yuan:
How would you locate the desired PSI? That's why I ended up replacing the comparison.

neetopia (approved):

PR #1160, opened 2022-10-26
Bump 1438

neetopia:


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -367,13 +412,17 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                         kspTask.compilerPluginClasspath = project.configurations.getByName(pluginConfigurationName)
                         kspTask.commonSources.from(kotlinCompileTask.commonSources)
                         kspTask.compilerPluginOptions.addPluginArgument(kotlinCompileTask.compilerPluginOptions)
+                        val kspOptions = kspTask.options.get().flatMap { listOf("-P", it.toArg()) }
```
ting-yuan, line 415: Comment: this needs to happen after `configureAsKspTask` is called, where `kspTask.options` is set up.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -367,13 +412,17 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
...
+                        kspTask.compilerOptions.freeCompilerArgs.value(
+                            kspOptions + kotlinCompileTask.compilerOptions.freeCompilerArgs.get()
+                        )
```
ting-yuan, line 418: If `additionalCompilerOptions` was removed, are there any alternatives? Ideally, these are best set at execution time.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -488,7 +537,12 @@ interface KspTask : Task {
 abstract class KspTaskJvm @Inject constructor(
     workerExecutor: WorkerExecutor,
     objectFactory: ObjectFactory
-) : KotlinCompile(KotlinJvmOptionsImpl(), workerExecutor, objectFactory), KspTask {
+) : KotlinCompile(
+        objectFactory.newInstance(KotlinJvmCompilerOptionsDefault::class.java),
```
ting-yuan, line 541: You can rely on injection so as to avoid referencing the KGP internal `KotlinJvmCompilerOptionsDefault`. Example: https://github.com/ting-yuan/ksp/blob/gradle/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/tasks/inherited/InheritedTasks.kt#L268

ting-yuan (commented):
Although legacy js is being deprecated, it is still available if user really wants it. Can we keep testing for it?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
neetopia: `configureAsKspTask` is at line 409 which is before this line?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
neetopia: the deprecation message says `Replaced with compilerOptions.freeCompilerArgs`

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
neetopia: putting it in a constructor seems to expose internal classes (`KotlinJvmCompilerOptionsDefault` here) therefore is causing error.

gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/ProcessorClasspathConfigurationsTest.kt
```diff
@@ -96,7 +96,7 @@ class ProcessorClasspathConfigurationsTest {
             """
                 kotlin {
                     jvm { }
-                    js { browser() }
+                    js(BOTH) { browser() }
```
neetopia, line 99: Can't figure out a way to pass compiler free arguments to kmp project with android plugin applied

ting-yuan (approved):

Issue #1167, opened 2022-11-01
CommandLineArgument not included as SubpluginOption


cdsap:
Given a `CommandLineArgumentProvider` defined in the build configuration:
```kotlin
class Provider(roomOutputDir: File) : CommandLineArgumentProvider {
    @OutputDirectory
    val outputDir = roomOutputDir

    override fun asArguments(): Iterable<String> {
        return listOf(
            "room.schemaLocation=${outputDir.path}"
        )
    }
}
```

When we are adding a `commandLineArgument` using `KspTask` like:
```
tasks.withType<KspTask>().configureEach {
    commandLineArgumentProviders.add(Provider(project.layout.projectDirectory.dir("schemas-${name}").asFile))
}
```
The argument is not included as `apoption`. 

Using a task graph listener to print the apoptions and commandline arguments:
```kotlin
gradle.taskGraph.whenReady {
    this.allTasks.forEach {
        if(it is com.google.devtools.ksp.gradle.KspTask) {
            it.options.get().filter { it.key == "apoption" }.forEach { option ->
                println("${option.key}=${option.value}")
            }
            it.commandLineArgumentProviders.get().forEach { commandLine ->
                println("CommandLineArgument - ${commandLine.asArguments()}")
            }
        }
    }
}
```
the current result is:
```
CommandLineArgument - [room.schemaLocation=/MyApplication6/app/schemas-kspDebugKotlin]
```
when the expected result is:
```
apoption=room.schemaLocation=/MyApplication6/app/schemas-kspDebugKotlin
CommandLineArgument - [room.schemaLocation=/MyApplication6/app/schemas-kspDebugKotlin]
```

PR #1168, opened 2022-11-01
Including CommandLineArgument as SubpluginOption with KspTask

cdsap:
This PR proposes a solution for the issue explained in #1167

CommandLineArguments are included as SubpluginOptions through the KspExtension. With this change, we include the CommandLineArguments from the KspTask allowing ad-hoc configurations for arguments in the task:

```kotlin
tasks.withType<KspTask>().configureEach {
    commandLineArgumentProviders.add(Provider(project.layout.projectDirectory.dir("schemas-${name}").asFile))
}
```


ting-yuan (approved):
Thanks for the patch! Could you fix the lint error so that it won't break the CI after we merge it?

PR #1176, opened 2022-11-04
AA: support use site annotations

neetopia:
there is some code copied with minor modification from FE1.0 implementation, unfortunately not reusable due to relying on implementation details, with more similar case discovered, it is also worth considering how to better abstract common logics.


ting-yuan (approved):
Before figuring out a way to hoist common logic, how about adding comments on both sides to associate them together if
1. they are not covered by tests well enough, AND
2. the logic is non-trivial

PR #1177, opened 2022-11-05
AA: getDeclarationsFromPackage 

neetopia:
also includes additional works to get modifiers for library symbols since they have no psi at this time. Ideally library symbols should also have PSI but it can be either expensive (likely building a stubs with more details not needed for KSP) or take a long time to be available, therefore put the current logic in place.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
@@ -22,7 +22,11 @@ import com.google.devtools.ksp.impl.symbol.kotlin.KSClassDeclarationEnumEntryImp
 import com.google.devtools.ksp.impl.symbol.kotlin.KSClassDeclarationImpl
 import com.google.devtools.ksp.impl.symbol.kotlin.KSFileImpl
 import com.google.devtools.ksp.impl.symbol.kotlin.KSFileJavaImpl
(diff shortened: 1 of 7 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -134,8 +142,29 @@ class ResolverAAImpl(
...
+        return analyze {
+            val packageNames = packageName.split(".")
+            var packages = listOf(analysisSession.ROOT_PACKAGE_SYMBOL)
+            for (curName in packageNames) {
+                packages = packages
+                    .flatMap { it.getPackageScope().getPackageSymbols { it.asString() == curName } }
+                    .distinct()
+            }
+            packages.flatMap {
+                it.getPackageScope().getAllSymbols().distinct().mapNotNull { symbol ->
```
ting-yuan, line 156: Why does it need `distinct()`?

ting-yuan (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
neetopia: AA returns duplicate package symbols for reasons I don't know. Probably need a fix in upstream but put distinct() here for now.

PR #1187, opened 2022-11-10
Introduce `kspOnlyVersion` to root build script

ting-yuan:
so that `kspVersion` can be constructed by

  `"$kotlinBaseVersion-$kspOnlyVersion"`

where `kotlinBaseVersion` is available in gradle.properties.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,12 @@ import com.google.devtools.ksp.configureKtlintApplyToIdea
 val sonatypeUserName: String? by project
 val sonatypePassword: String? by project
 
+val kotlinBaseVersion: String? by project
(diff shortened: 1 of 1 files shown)


neetopia:
what the purpose of this property since we already have `kspVersion`? or just replace the existing `kspVersion` with this one?

ting-yuan:
Supplying Kotlin version from command line is error prone; It should always match gradle.properties / the single source of truth. Ideally, making `kspVersion` contain only KSP version and creating another `kotlinKspVersion` to replace the current usages of `kspVersion` would be the way to go, but that would be a huge patch.

neetopia (approved):

PR #1201, opened 2022-11-16
Add `mapJavaNameToKotlinMutable` API to Resolver.

Jeffset:
This new function is complementary to the `mapJavaNameToKotlin`, which returns immutable Kotlin platform types.
There was no conventional way to obtain mutable types based on Java ones.


neetopia:
overall lgtm, my major concern here is if we want a more versatile API to avoid future addition to our APIs, we are still thinking about it therefore want to hold the PR for a while.

Jeffset:
> overall lgtm, my major concern here is if we want a more versatile API to avoid future addition to our APIs, we are still thinking about it therefore want to hold the PR for a while.

Okay, thank you for the response!

PR #1204, opened 2022-11-18
AA java refs workaround

neetopia:
In FIR java types are first resolved to `JavaTypeRef` which is incompatible with `FirResolvedTypeRef` used in `FirCallableSymbol.returnType` whichj is in turn used by AA for getting symbol return types.

In FIR, java symbols are constructed to default be in last FIR resolve phase, therefore trying a workaround where we manually convert the java type refs `o` resolved type refs with internal functions from fir.java module should be safe for KSP.

Using this workaround before upstream fixes this issue

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
@@ -15,13 +15,20 @@
  * limitations under the License.
  */
 
+@file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
 package com.google.devtools.ksp.impl.symbol.kotlin
 
 import com.google.devtools.ksp.KSObjectCache
 import com.google.devtools.ksp.symbol.*
+import org.jetbrains.kotlin.analysis.api.fir.symbols.KtFirFunctionSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.*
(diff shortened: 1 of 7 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -59,7 +66,19 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
         }
     }
 
+    @OptIn(SymbolInternals::class)
     override val returnType: KSTypeReference? by lazy {
+        // FIXME: temporary workaround before upstream fixes java type refs.
+        if (origin == Origin.JAVA) {
+            if (ktFunctionSymbol is KtFirFunctionSymbol) {
+                (ktFunctionSymbol.firSymbol.fir as? FirJavaMethod)?.also {
```
ting-yuan, line 74: Can you replace the type checks with `KtFirSymbol` and `FirCallableDeclaration`, hoist the workaround and share among the 3 places?

If possible, please keep those in a `WorkaroundsUsingInternals.kt` so that we won't forget to clean them up before formal release.

neetopia: The reason I put this at different places is that, unfortunately `FirCallableDeclaration.returnTypeRef` is declared as a `val`, and it only gets overridden to `var` in `FirJavaMethod` , `FirJavaField` and `FirJavaValueParameter` (somehow makes total sense since they are the ones need to be transformed)

ting-yuan (approved):

PR #1212, opened 2022-11-26
Configure KSP tasks with modern KGP APIs

ting-yuan:
and minimize customizations in subclasses.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -30,7 +30,6 @@ import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.FileCollection
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.ListProperty
-import org.gradle.api.provider.MapProperty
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -365,69 +282,131 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                     kspTask.setSource(kotlinCompilation.defaultSourceSet.kotlin)
                 }
             }
+            kspTask.exclude { kspOutputDir.isParentOf(it.file) }
```
neetopia, line 285: this line seems to be new, can you elaborate it?

neetopia (approved):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
ting-yuan: Generated sources do not participate in processing. Previously, this was done in overridden `sources` in KSP tasks. BTW, `javaSources` is still needed because the `patternFilterable` isn't implemented properly in KGP.

PR #1246, opened 2022-12-20
Deprecate KspTask.commandLineArgumentProviders.

ting-yuan:
KSP will delegate task creation to KGP and drop support of the per-task commandLineArgumentProviders.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinFactories.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinFactories.kt
@@ -128,6 +128,7 @@ interface KspTask : Task {
     @get:Internal
     val options: ListProperty<SubpluginOption>
 
+    @Deprecated("KSP is going to use compilation tasks created by KGP and no longer support this property.")
     @get:Nested
     val commandLineArgumentProviders: ListProperty<CommandLineArgumentProvider>
 


ting-yuan:
Hello @cdsap, we are replacing the hacky KspTask, which inherits and overrides internal implementations in kotlin-gradle-plugin, with tasks created by public APIs in kotlin-gradle-plugin-api. Because there will be no longer  `commandLineArgumentProviders` property, projects relying on it will be affected. May I know your use cases and see how we can find a mitigation?

cdsap:
hi @ting-yuan, thanks for the heads up. It makes sense.

The main use case is the Ksp implementation of the `RoomWorkaround` for [Android Cache fix plugin](https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy).  
Room with Ksp has the same issue as Kapt(https://github.com/google/ksp/issues/861). We use  `commandLineArgumentProviders `  to add a variant-specific output of the room schema for the task. This task is currently in progress but we can discuss different options.


How looks the new approach?

runningcode:
I think this is the main use case from the Google documentation: https://developer.android.com/training/data-storage/room/migrating-db-versions#export-schemas

There needs to be an alternative way to provide task specific inputs and outputs to the annotation processor arguments. The deprecation warning should point to this.

neetopia (approved):

PR #1257, opened 2023-01-07
Use correct psi implementation for finding type parameter declaration.

### LINKED ISSUE #1210: Resolved KSTypeParameter doesn't equal the source
For example,
```
interface Foo<T> {
    val t: T
}
```

I handle it below
```
val klassFoo = resolver.getClassDeclarationByName("Foo")!!
Log.w(klassFoo.typeParameters.first() == klassFoo.getDeclaredProperties().first().type.resolve().declaration)
```
The result is ```false```, which is unexpected because it's always ```true``` when that symbol is not ```KSTypeParameter```.

neetopia:
Fixes #1210.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -742,11 +742,12 @@ class ResolverImpl(
     fun findDeclaration(kotlinType: KotlinType): KSDeclaration {
         val descriptor = kotlinType.constructor.declarationDescriptor
         val psi = descriptor?.findPsi()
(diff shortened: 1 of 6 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -742,11 +742,12 @@ class ResolverImpl(
     fun findDeclaration(kotlinType: KotlinType): KSDeclaration {
         val descriptor = kotlinType.constructor.declarationDescriptor
         val psi = descriptor?.findPsi()
-        return if (psi != null && psi !is KtTypeParameter) {
```
neetopia, line 745: I can't recall why we filtered out type parameter here..

ting-yuan (approved):

PR #1282, opened 2023-01-24
Exclude generated libraries from deps

### LINKED ISSUE #1280: Incremental compilation fails with "Number of loaded files in snapshots differs"
We are building support for KSP in Micronaut (a popular Java/Kotlin framework). The initial implementation is done but it fails with KSP incremental compilation enabled with an error "Number of loaded files in snapshots differs" on the second compile:

```
./gradlew test
> Task :kspKotlin FAILED

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':kspKotlin'.
> Number of loaded files in snapshots differs. Reported changed files: []
```

To reproduce download this example project and run `./gradlew cTK` twice.

[ksp-test2.zip](https://github.com/google/ksp/files/10480874/ksp-test2.zip)

ting-yuan:
Fixes #1280

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -324,8 +324,8 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                 resourceOutputDir
             )
 
+            val kotlinCompileTask = kotlinCompileProvider.get()
             if (kspExtension.allowSourcesFromOtherPlugins) {
-                val kotlinCompileTask = kotlinCompileProvider.get()
                 fun FileCollection.nonSelfDeps(): List<Task> =
                     buildDependencies.getDependencies(null).filterNot {
                         it.name == kspTaskName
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -514,6 +514,7 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                         kspTask.compilerOptions.freeCompilerArgs.value(
                             kspOptions + kotlinCompileTask.compilerOptions.freeCompilerArgs.get()
                         )
+                        kspTask.compilerOptions.useK2.value(false)
```
neetopia, line 517: is this related to this fix?

neetopia (approved):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
ting-yuan: Not specific to #1280, but a fix that is small enough to come with other commits.

Issue #983, opened 2022-05-11
support KSP compiler plugin with K2 compiler


neetopia:
follow up of #982

neetopia:
Well actually might not need this. K2 compiler should use the analysis API implementation of KSP which won't be a compiler plugin anymore.

zuisong:
Can we use ksp with k2 compiler now?

I got this error   , kotlin version `1.7.20Beta` ksp version `1.7.20-Beta-1.0.6`
```
[ERROR] Failed to execute goal org.jetbrains.kotlin:kotlin-maven-plugin:1.7.20-Beta:compile (compile) on project common: Compilation failure
[ERROR] There are some plugins incompatible with K2 compiler:
[ERROR]   com.google.devtools.ksp.KotlinSymbolProcessingComponentRegistrar
[ERROR] Please remove -Xuse-k2
```

neetopia:
It is not supported at this moment, currently upstream disabled compiler plugin for K2 compiler, and current KSP implementation is built on old frontend therefore not working with K2 compiler.

We are working on a whole new implementation built on top of K2 compiler, please stay tuned.

Anivie:
Wen will it work?
The 1.8.0 Beta is comming, but ksp still can't working with k2.

ting-yuan:
If you are using Gradle, KSP should work with K2 enabled. KSP has its own Gradle task that always calls K1 while the compilation can use K2.

KSP turns off K2 explicitly in its own task but bugs are always a possibility. How do you turn on K2?

dave08:
I just got (on gradle 7.6 and kotlin 1.8.0 with ksp for that version):
`There are some plugins incompatible with K2 compiler:
  com.google.devtools.ksp.KotlinSymbolProcessingComponentRegistrar
  org.jetbrains.kotlinx.serialization.compiler.extensions.SerializationComponentRegistrar
Please remove -Xuse-k2`

I enabled k2 like this:
```kotlin
tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
    kotlinOptions {
        freeCompilerArgs += "-opt-in=kotlinx.coroutines.ExperimentalCoroutinesApi"
        useK2 = true
    }
}
```

apatrida:
I just tried it and I get an error as well, same as above.

sureshg:
Tried the latest `1.8.20-Beta` version and having the same issue.
```kotlin
kotlin= "1.8.20-Beta"
[...]

PR #1322, opened 2023-02-10
set language version to allow KSP with FIR compiler

neetopia:
`use-k2` was deprecated.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -22,6 +22,7 @@ import org.gradle.api.Action
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.UnknownTaskException
+import org.gradle.api.artifacts.Configuration
 import org.gradle.api.attributes.Attribute
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.FileCollection
@@ -35,6 +36,8 @@ import org.gradle.process.CommandLineArgumentProvider
 import org.gradle.tooling.provider.model.ToolingModelBuilderRegistry
 import org.gradle.util.GradleVersion
 import org.jetbrains.kotlin.config.ApiVersion
+import org.jetbrains.kotlin.config.LanguageVersion.Companion.LATEST_STABLE
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 import org.jetbrains.kotlin.gradle.dsl.kotlinExtension
 import org.jetbrains.kotlin.gradle.internal.kapt.incremental.CLASS_STRUCTURE_ARTIFACT_TYPE
 import org.jetbrains.kotlin.gradle.internal.kapt.incremental.ClasspathSnapshot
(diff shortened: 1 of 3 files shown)


neetopia:
fixes #983

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -83,6 +84,7 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
         const val KSP_GROUP_ID = "com.google.devtools.ksp"
         const val KSP_PLUGIN_CLASSPATH_CONFIGURATION_NAME = "kspPluginClasspath"
         const val KSP_PLUGIN_CLASSPATH_CONFIGURATION_NAME_NON_EMBEDDABLE = "kspPluginClasspathNonEmbeddable"
+        val LANGUAGE_VERSION = KotlinVersion.KOTLIN_1_8
```
ting-yuan, line 87: You can use `org.jetbrains.kotlin.config.LanguageVersion.LATEST_STABLE` until bumping to 2.0.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -415,7 +417,11 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                         configureAsAbstractKotlinCompileTool(kspTask as AbstractKotlinCompileTool<*>)
                         configurePluginOptions(kspTask)
                         kspTask.compilerOptions.noJdk.value(kotlinCompileTask.compilerOptions.noJdk)
-                        kspTask.compilerOptions.useK2.value(false)
+                        kspTask.compilerOptions.languageVersion.orNull?.let { version ->
```
ting-yuan, line 420: Configuration of shared compilerOptions can be moved into a local function:
```
fun configureKotlinCompilationTask(task: KotlinCompilationTask) {
  ...
}
```

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/PlaygroundIT.kt
```diff
@@ -147,19 +146,25 @@ class PlaygroundIT {
         }
     }
 
-    // Disabled for now: ERROR: K2 does not support plugins yet, so please remove -Xuse-k2 flag
-    // Test -Xuse-fir for compilation; KSP still uses FE1.0
-    @Ignore
     @Test
     fun testFirPreview() {
-        val gradleProperties = File(project.root, "gradle.properties")
-        gradleProperties.appendText("\nkotlin.useK2=true")
```
ting-yuan, line 156: Can you confirm what would happen if a user specifies `-Pkotlin.useK2=true` w. and w.o. `languageVersion`? So as to make sure removing `kspTask.compilerOptions.useK2.value(false)` in the above is safe. No need to put it into a test case.

neetopia: This option is deprecated and therefore the errors we saw, it should not be set to true anymore if users want to use FIR.

ting-yuan: I'd rather spend a few minutes beforehand than hours triage and blaming users' faults later.

neetopia: I meant the behavior when user tries to use this flag in the setup you mentioned in a normal compilation, which is not KSP related. For KSP it doesn't matter, KSP won't get invoked in both scenarios.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
neetopia: LATEST_STABLE is 1.9 in main branch.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
neetopia: done

ting-yuan (approved):

PR #1334, opened 2023-03-02
UPDATE_KOTLIN_VERSION: 1.9.0-dev-2695

neetopia:
updating to latest compiler bootstrap.

Main changes: Kotlin/Native gradle task moved compiler options into constructor, added requirement for `moduleName`, changed accordingly in KSP gradle plugin.

Muted annotationOnConstructorParameter test for AA for now, reason: after this update, analysis API is no longer returning annotations on properties declared in constructor at least for the case declared in this test, need to figure out what's the cause afterwards.


neetopia:
Looks like there is a lint error. Make sure at least lint passes, you can check by running `./gradlew ktlint` prior to pushing. Ideally, a full run of `./gradlew build` will include all necessary checks.

jsjeon (approved):

Issue #1335, opened 2023-03-02
mapToJvmSignature crash on local class


bubenheimer:
Given this top-level function and local class:

```
private fun simple() {
    class VerySimple
}
```

Calling `Resolver.mapToJvmSignature()` on the `KSClassDeclaration` corresponding to `class VerySimple` crashes ksp:

```
java.lang.IllegalArgumentException: Unexpected container: private fun simple(): kotlin.Unit defined in com.example in file Sample.kt[SimpleFunctionDescriptorImpl@6668c51f] for class VerySimple
	at org.jetbrains.kotlin.load.kotlin.DescriptorBasedTypeSignatureMappingKt.computeInternalName(descriptorBasedTypeSignatureMapping.kt:190)
	at org.jetbrains.kotlin.load.kotlin.DescriptorBasedTypeSignatureMappingKt.mapType(descriptorBasedTypeSignatureMapping.kt:141)
	at org.jetbrains.kotlin.codegen.state.KotlinTypeMapper.mapType(KotlinTypeMapper.kt:276)
	at org.jetbrains.kotlin.codegen.state.KotlinTypeMapper.mapType$default(KotlinTypeMapper.kt:267)
	at org.jetbrains.kotlin.codegen.state.KotlinTypeMapper.mapType(KotlinTypeMapper.kt:263)
	at com.google.devtools.ksp.processing.impl.ResolverImpl.mapToJvmSignatureInternal$compiler_plugin(ResolverImpl.kt:356)
	at com.google.devtools.ksp.processing.impl.ResolverImpl.mapToJvmSignature(ResolverImpl.kt:353)
	at org.bubenheimer.instancelimit.InstanceLimitProcessor.createClassInfo(InstanceLimitProcessor.kt:111)
	at org.bubenheimer.instancelimit.InstanceLimitProcessor.addInstanceLimit(InstanceLimitProcessor.kt:64)
	at org.bubenheimer.instancelimit.InstanceLimitProcessor.process(InstanceLimitProcessor.kt:214)
	at org.bubenheimer.instancelimit.InstanceLimitProcessor.process(InstanceLimitProcessor.kt:140)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension$doAnalysis$6$1.invoke(KotlinSymbolProcessingExtension.kt:291)
[...]

PR #1338, opened 2023-03-02
Add Resolver.mapToJvmClassName() for KSClassDeclaration

### LINKED ISSUE #1336: Provide binary name for KSClassDeclaration in Resolver
I need the binary name corresponding to `KSClassDeclaration` in my processor, and would like to see this functionality added to ksp. The functionality is readily available from asm, but not currently exposed at the ksp API level. It is similar to what `Resolver.mapToJvmSignature()` provides, but for class binary names. It also is similar to `Resolver.getOwnerJvmClassName()`, without the part of finding the owner.

It may make sense to instead mirror `Resolver.mapToJvmSignature()` with `Resolver.mapToJvmClassName()`, making this functionality available for `KSClassDeclaration`, `KSFunctionDeclaration`, and `KSPropertyDeclaration`. I am not clear on the implementation in the latter two cases, while the first one is trivial.

I can provide a PR for the `KSClassDeclaration` case. I am not as clear on the `KSFunctionDeclaration` and `KSPropertyDeclaration` cases.

bubenheimer:
Resolves #1336

Focusing on `KSClassDeclaration` for now. Other types of `KSDeclaration` may be added later by changing the parameter type to `KSDeclaration`.

### DIFF
--- a/test-utils/testData/api/jvmClassNameMapper.kt
+++ b/test-utils/testData/api/jvmClassNameMapper.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 Google LLC
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TEST PROCESSOR: MapJvmClassNameProcessor
+// EXPECTED:
+// Cls1
+// Cls1$Cls2
+// Cls1$Cls3
+// Cls1$Cls4
+// JavaInterfaceWithVoid
+// JavaClass1
+// JavaClass1$JavaClass2
+// JavaClass1$JavaClass3
+// JavaClass1$JavaClass4
+// JavaClass5
+// JavaAnno
+// END
+
+// FILE: Cls1.kt
+class Cls1 {
+    class Cls2
+
+    open class Cls3
+
+    inner class Cls4
+}
+
+// FILE: JavaInterfaceWithVoid.java
+interface JavaInterfaceWithVoid {}
+
+// FILE: JavaClass1.java
+class JavaClass1 {
+    static final class JavaClass2 {}
+    static class JavaClass3 {}
+    class JavaClass4 {}
+}
+
+class JavaClass5 {}
+
+// FILE: JavaAnno.java
+@interface JavaAnno {}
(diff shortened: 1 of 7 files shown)


test-utils/testData/api/jvmClassNameMapper.kt
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TEST PROCESSOR: MapJvmClassNameProcessor
+// EXPECTED:
+// Cls1
+// Cls1$Cls2
```
neetopia, line 21: Is there any more complicated class names? It looks to me that if only nested class matters, we can actually write this in a format of extension function by recursively checking the parent declaration and build a name?

bubenheimer:
The CI build seems to trip over ResolverAAImpl. Not sure what that means.

neetopia:
`ResolverAAImpl` is a WIP project. Since it inherits `Resolver` interface you need to implement it to pass compilation, the simplest way is to just put a `TODO()` in the implementation.

Although, if the logic is simple enough as I asked in the comment, maybe it is easier to just add an extension function in [util](https://github.com/google/ksp/blob/main/api/src/main/kotlin/com/google/devtools/ksp/utils.kt) so that there is no need to add an api to `Resolver`?

test-utils/testData/api/jvmClassNameMapper.kt
bubenheimer: I used signatureMapper.kt (and MapSignatureProcessor) as a template, for the equivalent Resolver.mapToJvmSignature() functionality. Both tests are equally underpowered. However, I don't think can really do much right now to make my test a lot more expressive until #1335 is fixed, which crashes the new Resolver.mapToJvmClassName() in the same manner, so I expect that a fix for #1335 will similarly clear the landscape here.

Once the fix is in place, I'd think the following additional tests will be particularly useful:

- Local class inside internal function (function name mangling)
- Local class inside function annotated with @JvmName("xyz")
- Local class inside overloaded function

Anything else of particular interest that you can think of?

bubenheimer:
Gotcha. Logic is not so simple, unfortunately.

bubenheimer:
I'm unable to build on my local MacOS for some reason, so I'll have to rely on CI for guidance I guess.

neetopia:
Based on some observations on the behavior of compiled class files, the name mostly gets complicated when it comes to local classes, which is unfortunately not supported as the moment even with this PR. 

With local classes though, the behavior seems to be about adding a number specifying the functions that a local class belongs to distinguish and avoid name clashes among multiple local classes, it it seems doable with our `getDeclarationsInSourceOrder` API. If that is the case (name only matters with the ordering of which the parent function belong to), it might still be fine to just implement as an extension function. 

My concern is that without an explicit spec, aforementioned approach won't work, then we still need to rely on compiler to do the work. But at the same time, compiler is throwing exceptions for such queries, we might need to fix compiler before going with current approach in this PR.

bubenheimer:
I'm ok with the idea of a recursive extension function for this. However, it does not seem trivial to me to get all possible cases right at each level, and dealing with correct naming & name mangling. Correct implementation may also be tied to specific compiler versions, naming & name mangling might change between versions. Perhaps even specific to JVM version, if it turned out to not be part of an official specification.

Perhaps it seems more daunting to me because I lack a good grasp of it. Are you or KSP team willing to implement this kind of function? I don't think I will dare to do it myself, even though my processor won't work with local classes without this functionality.

I guess, like you said, the next step in any case would be to find out if local class binary name structure is documented/specified.

jaschdoc:
Closing this PR since it is no longer relevant.

PR #1344, opened 2023-03-11
AA: reference element rework

neetopia:
* reimplement reference element to align with upstream's change to `KtType.qualifiers`
* Support def non null types in reference element
   *  tested for symbols from source, symbols from binaries blocked due to get top level callables not working for libraries.
* some other minor fixes.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
@@ -1,37 +1,77 @@
 package com.google.devtools.ksp.impl.symbol.kotlin
 
 import com.google.devtools.ksp.IdKeyPair
+import com.google.devtools.ksp.IdKeyTriple
 import com.google.devtools.ksp.KSObjectCache
 import com.google.devtools.ksp.symbol.*
-import org.jetbrains.kotlin.analysis.api.components.buildClassType
-import org.jetbrains.kotlin.analysis.api.types.KtUsualClassType
+import org.jetbrains.kotlin.analysis.api.types.KtClassType
+import org.jetbrains.kotlin.analysis.api.types.KtClassTypeQualifier
+import org.jetbrains.kotlin.analysis.api.types.KtTypeParameterType
 
 class KSClassifierReferenceImpl private constructor(
-    internal val ktType: KtUsualClassType,
+    internal val ktType: KtClassType,
+    internal val index: Int,
     override val parent: KSTypeReference?
 ) : KSClassifierReference {
-    companion object : KSObjectCache<IdKeyPair<KtUsualClassType, KSTypeReference?>, KSClassifierReferenceImpl>() {
-        fun getCached(ktType: KtUsualClassType, parent: KSTypeReference?) =
-            cache.getOrPut(IdKeyPair(ktType, parent)) { KSClassifierReferenceImpl(ktType, parent) }
+    companion object : KSObjectCache<IdKeyTriple<KtClassType, Int, KSTypeReference?>, KSClassifierReferenceImpl>() {
+        fun getCached(ktType: KtClassType, index: Int, parent: KSTypeReference?) =
+            cache.getOrPut(IdKeyTriple(ktType, index, parent)) { KSClassifierReferenceImpl(ktType, index, parent) }
     }
+
+    private val classifierReference: KtClassTypeQualifier
+        get() = ktType.qualifiers[index]
+
     override val qualifier: KSClassifierReference? by lazy {
-        ktType.classId.outerClassId?.let {
-            analyze {
-                buildClassType(it)
-            }
-        }?.let { getCached(it as KtUsualClassType, parent) }
+        if (index == 0) {
+            null
+        } else {
+            getCached(ktType, index - 1, parent)
+        }
     }
 
     override fun referencedName(): String {
-        return ktType.classId.asFqNameString()
+        return classifierReference.name.asString()
     }
 
     override val typeArguments: List<KSTypeArgument> by lazy {
-        ktType.typeArguments().map { KSTypeArgumentImpl.getCached(it, this) }
+        classifierReference.typeArguments.map { KSTypeArgumentImpl.getCached(it, this) }
     }
 
     override val origin: Origin = parent?.origin ?: Origin.SYNTHETIC
 
     override val location: Location
         get() = parent?.location ?: NonExistLocation
+
+    override fun toString(): String {
+        return referencedName()
+    }
+}
+
+class KSClassifierParameterImpl private constructor(
+    internal val ktType: KtTypeParameterType,
+    override val parent: KSTypeReference?
+) : KSClassifierReference {
+    companion object : KSObjectCache<IdKeyPair<KtTypeParameterType, KSTypeReference?>, KSClassifierParameterImpl>() {
+        fun getCached(ktType: KtTypeParameterType, parent: KSTypeReference?) =
+            KSClassifierParameterImpl.cache.getOrPut(IdKeyPair(ktType, parent)) {
+                KSClassifierParameterImpl(ktType, parent)
+            }
+    }
+
+    override val qualifier: KSClassifierReference? = null
+
+    override fun referencedName(): String {
+        return ktType.name.asString()
+    }
+
+    override val typeArguments: List<KSTypeArgument>
+        get() = emptyList()
+    override val origin: Origin
+        get() = parent?.origin ?: Origin.SYNTHETIC
+    override val location: Location
+        get() = parent?.location ?: NonExistLocation
+
+    override fun toString(): String {
+        return referencedName()
+    }
 }
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
@@ -62,10 +62,10 @@ class KSTypeImpl private constructor(internal val type: KtType) : KSType {
     }
 
     override val nullability: Nullability by lazy {
-        if (type.nullability == KtTypeNullability.NON_NULLABLE) {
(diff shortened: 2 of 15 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
```diff
@@ -62,10 +63,26 @@ class KSTypeImpl private constructor(internal val type: KtType) : KSType {
...
+            }
+            if (type is KtFlexibleType && isNullable(type.upperBound)) {
+                return true
+            }
+            if (type is KtDefinitelyNotNullType) {
+                return false
+            }
+            if (type is KtTypeParameterType) {
+                return KSTypeParameterImpl.getCached(type.symbol).bounds.map { it.resolve() }
+                    .contains(ResolverAAImpl.instance.builtIns.anyType.makeNullable())
```
jsjeon, line 78: 1. This goes through KSP's `KSType` creation / lookup
2. This checks if the type parameter has nullable `Any?` as an upper bound, but that misses a lot of nullable type parameters.
3. The whole logic misses `KtIntersectionType`

To mitigate everything all together, how about just using `KtTypeInfoProviderMixIn#canBeNull`?

Re: 2, note that type parameter without any bound is nullable (like just `<T>`); type parameter is nullable if _all_ of its upper bound is nullable (like `<T> ... where T : Int?, T : Long?`, which eventually means T : Number?). The current logic thought the following is nullable `<T> ... where T : Any?, T : Any` which is actually not nullable since upper bounds are intersected. That's the same for intersection type.

Again, `KtTypeInfoProviderMixIn#canBeNull` will take care of everything.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
```diff
@@ -1,37 +1,77 @@
...
-    internal val ktType: KtUsualClassType,
+    internal val ktType: KtClassType,
+    internal val index: Int,
     override val parent: KSTypeReference?
 ) : KSClassifierReference {
-    companion object : KSObjectCache<IdKeyPair<KtUsualClassType, KSTypeReference?>, KSClassifierReferenceImpl>() {
-        fun getCached(ktType: KtUsualClassType, parent: KSTypeReference?) =
-            cache.getOrPut(IdKeyPair(ktType, parent)) { KSClassifierReferenceImpl(ktType, parent) }
+    companion object : KSObjectCache<IdKeyTriple<KtClassType, Int, KSTypeReference?>, KSClassifierReferenceImpl>() {
+        fun getCached(ktType: KtClassType, parent: KSTypeReference?, index: Int) =
```
jsjeon, line 17: Nit: you can match the order of value parameters here to the order of type parameters of `KSObjectCache` above? I.e., `ktType, index, parent`?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
```diff
@@ -1,37 +1,77 @@
...
     }
 
     override fun referencedName(): String {
-        return ktType.classId.asFqNameString()
+        return classifierReference.name.asString()
     }
 
     override val typeArguments: List<KSTypeArgument> by lazy {
-        ktType.typeArguments().map { KSTypeArgumentImpl.getCached(it, this) }
+        classifierReference.typeArguments.map { KSTypeArgumentImpl.getCached(it, this) }
```
jsjeon, line 37: I wonder if `KtNonErrorClassType.typeArguments` is just renamed to `KtNonErrorClassType.ownTypeArguments`?

This change seems like, picking the specific classifier reference at `index` and retrieving _that_ one's type arguments, instead of loading the type arguments of `ktType`.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
```diff
@@ -1,37 +1,77 @@
...
+        get() = ktType.qualifiers[index]
+
     override val qualifier: KSClassifierReference? by lazy {
-        ktType.classId.outerClassId?.let {
-            analyze {
-                buildClassType(it)
-            }
-        }?.let { getCached(it as KtUsualClassType, parent) }
+        if (index == 0) {
+            null
```
jsjeon, line 26: I'm not sure I understand what this means: for index 0, `null` qualifier. Is it like a `<root>` package? Could you clarify this with an example?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
```diff
@@ -1,37 +1,77 @@
...
 
     override val location: Location
         get() = parent?.location ?: NonExistLocation
+
+    override fun toString(): String {
+        return referencedName()
+    }
+}
+
+class KSClassifierParameterImpl private constructor(
```
jsjeon, line 50: At a glance, everything else is in its own file, but why not this one?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
neetopia: Basically it is trying to check if there is more qualifiers, therefore, when the index is already at 0, there can't be more qualifiers before current qualifier.

e.g. for a reference of `A.B` where B is a class defined inside A, index 0 means the reference element part at `A`, and 1 means element at `B`, the test covers it but unfortunately the whole test is blocked by some other unrelated bugs (get top level callables).

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
neetopia: My thought on this is that this implementation still is a classifier reference element implementation, in FE 1.0 this is not separated out because it was PSI based, therefore I want to keep logically same reference elements in same file.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
neetopia: The logic is mainly reworked from FE1.0, I guess at least in FE1.0 the check for implicit upper bound is managed somewhere else. But if there is already a utility in AA available, then I am more than happy to use that :)

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
neetopia: Yes, IIRC the change to rename `KtNonErrorClassType.typeArguments` to `KtNonErrorClassType.ownTypeArguments` was introduced together with the qualifier.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
jsjeon: Yeah, AA FE1.0 uses `TypeUtils#isNullableType`: https://github.com/JetBrains/kotlin/blob/master/analysis/analysis-api-fe10/src/org/jetbrains/kotlin/analysis/api/descriptors/components/KtFe10TypeInfoProvider.kt#L41, which will eventually handle implicit upper bound here: https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/types/TypeUtils.java#L350

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
jsjeon: So, my question was whether this is supposed to be `ktType.ownTypeArguments.map { ... }`, not `classifierReference.typeArguments.map { ... }`.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
jsjeon: Sorry, just to make sure one more time. So, in your example, for a reference of `A` then, index 0 returns `null` that indicates this one is the top-most qualifier?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
neetopia: No, it should be the latter because qualifier's argument only look at the arguments of its part. actually this `classifierReference` is not of a `KtType` class, but its own one.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassifierReferenceImpl.kt
neetopia: index 0 returns `A`, when trying to call `qualifier` on `A`, it triggers this if check and returns `null` to indicate there is no further qualifiers.

jsjeon (approved):

PR #1348, opened 2023-03-14
remove module name mangling workaround

neetopia:
per comment in #649 , as the [linking issue](https://youtrack.jetbrains.com/issue/KT-45777) has been fixed, we should be able to remove this name mangling logic now.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -421,7 +421,14 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
                         configurePluginOptions(kspTask)
                         kspTask.compilerOptions.noJdk.value(kotlinCompileTask.compilerOptions.noJdk)
                         configureLanguageVersion(kspTask)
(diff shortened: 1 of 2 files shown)


gavra0:
We still need this if we want to support the old IC i.e. `kotlin.incremental.useClasspathSnapshot=false`. Do we want to support it?

neetopia:
I guess we still need it. Will it be feasible to do the module name mangling based on this property?

neetopia:
Well, I guess I don't have full context on this property, is `false` the default value? Then we might have bigger issue for the issue we want to fix.

gavra0:
In KGP 1.8.20, default value is `kotlin.incremental.useClasspathSnapshot=true`. However, users can set `kotlin.incremental.useClasspathSnapshot=false`.

I think we can check for this property, and if it is set to false (assuming KGP is 1.8.20+), then keep the workaround.

SimonMarquis:
👋 Hi, I've been linked to this issue from here: https://issuetracker.google.com/issues/274117700
I've created a [reproducer project](https://github.com/SimonMarquis/IssueTracker-274117700) where using `internal` visibility modifier in conjunction with KSP 1.8.20-RC-1.0.9 and Jetpack Room 2.5.0.
Hopefully, a fix will land in the next stable release 🤞

gavra0 (approved):
Looking good, thanks!

PR #1350, opened 2023-03-23
set trackChangesInLookupCache for Incremental context

neetopia:
per the documentation for this property: `Controls whether changes in lookup cache should be tracked. Required for the classpath snapshots based IC approach`


gavra0 (commented):
LGTM

PR #1381, opened 2023-05-17
UPDATE_KOTLIN_VERSION: 1.9.0-dev-6976

neetopia:
2 things needs attention:
* there seem to be a known issue with native embeddable compiler, [fix](https://github.com/JetBrains/kotlin/commit/0fb5934ebd05d0433c9cf2a6ffbdb8269a576b45) in upstream not yet in this version of compiler. Disabling test for now, should get a mitigation if one is easy, or we will be enabling it again if the fix in compiler gets shipped with 1.9.0 (currently the fix is only in 1.9.20-dev- builds).
* The information for java synthetic classifier reference is lost in intellij upgrade, and even trying converting it to kotlin type (i.e. descriptor based solution) can't get the information back. The logic in descriptor looks making more sense to me, tentatively changing the test as a breaking change for this version upgrade.

Let me know any concerns with these 2 items.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.google.devtools.ksp.gradle
 
 import com.google.devtools.ksp.gradle.model.builder.KspModelBuilder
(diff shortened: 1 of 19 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+@file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
```
ting-yuan, line 17: per discussed, please keep these within KotlinFactories.kt

ting-yuan (approved):

PR #1398, opened 2023-06-02
Set up jvm toolchain requirements for tests

ribafish:
After loosing some time debugging why some tests were failing I noticed that tests require Java 11, with some in `:integration-tests:test` also require `JDK_9` to be set up and pointing to a Java 9 installation. These changes declare these requirements.

I also updated jUnit to latest jUnit 4 (4.13.1).


neetopia:
Thanks for the PR, can you resolve the conflict on the PR so we can proceed?

ribafish:
> Thanks for the PR, can you resolve the conflict on the PR so we can proceed?

Done

PR #1436, opened 2023-07-03
Fix test failures on Windows

lukellmann:
I ran into these failures while trying to build ksp on my Windows machine:

* `:api:checkApi` failed because of the wrong classpath separator, see https://www.baeldung.com/java-classpath-syntax and https://bugs.openjdk.org/browse/JDK-8262004
* `GradleCompilationTest.commandLineArgumentIsIncludedInApoptionsWhenAddedInKspTask` failed because the output had `\` where `/` was expected.


neetopia (approved):
Thanks for the fix.

Issue #1453, opened 2023-07-13
getSymbolsWithAnnotation does not return functions declared in anonymous objects


fan87:
Calling `Resolver.getSymbolsWithAnnotation` does not return functions annotated with that annotation in anonymous objects (e.g. `object : Something`)

Source Code:
```kt
open class Something

annotation class ExampleAnnotation

fun main() {
    println(object : Something() {
        @ExampleAnnotation fun testFunction() {}
    })
}
```

Annotation Processor:
```kt
resolver.getSymbolsWithAnnotation("ExampleAnnotation") // nothing is found
resolver.getSymbolsWithAnnotation("ExampleAnnotation", true) // as well as this one - if it even helps
```

If it works as intended, how am I supposed to do annotation processing on anonymous objects?

neetopia:
This is beyond the scope of KSP. The anonymous object you are checking here is inside a lambda which is part of an expression and KSP does not look into expressions.

PR #1458, opened 2023-07-18
fix java array argument variance

### LINKED ISSUE #1456: KSTypeArgument missing `out` variance.
This is reproducible in the following example.

First, define a Kotlin interface with `out` declaration-site variance:

```kotlin
// Kotlin source file
interface MyInterface<out T>
```

Next, define a Java class that references the interface and uses `? extends` in its type argument:

```java
// Java source file
class MyClass {            
  MyInterface<? extends String> field;
}
```

Now, if we look at the `KSTypeReference`/`KSReferenceElement` the variance of the type argument looks correct:

```
fieldDeclaration.type: MyInterface<? extends String>
fieldDeclaration.type.element?.typeArguments?.single(): COVARIANT String
```

However, if we look at the resolved type the variance of the type argument is incorrect:

```
fieldDeclaration.type.resolve(): (MyInterface<(String..String?)>..MyInterface<(String..String?)>?)
fieldDeclaration.type.resolve().arguments.single(): INVARIANT String
```

neetopia:
fixes #1456 

It is a little bit messed up regarding flexible type.

By Java specs, array should have a single variance of covariance for its type arguments. However, from kotlin compiler, the said type is resolved to a flexible type with a lower bound of invariant type argument and an upper bound of covariant type argument. It might make sense from kotlinc since invariant is more restrictive hence works for a lower bound. It is just that when compiler is dealing with `KoltinType.arguments`, it only looks at lower bound which causes trouble because the lowerbound argument of an array actually discarded the covariant part.

It might look to be better respect Java spec by making it lower bound covariant as well, but attempting to do that caused a lot of failures elsewhere and trying to work against compiler seems a bad idea, following kotlinc's result is easier to implement here. 

However it is still reasonable to add a special handling in reference element to reflect the covariant nature of array.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -738,7 +738,16 @@ class ResolverImpl(
                     )
                     javaTypeParameterMap[typeParameterDescriptor] = psi
                 }
-                return getKSTypeCached(resolveJavaType(type.psi, type), type.element.typeArguments, type.annotations)
+                val resolved = getKSTypeCached(
+                    resolveJavaType(type.psi, type),
+                    type.element.typeArguments,
+                    type.annotations
+                )
+                return if (type.psi is PsiArrayType) {
+                    resolved
+                } else {
+                    resolved.replace(type.element.typeArguments)
+                }
             }
             else -> throw IllegalStateException("Unable to resolve type for $type, $ExceptionMessage")
         }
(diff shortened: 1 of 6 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -738,7 +738,16 @@ class ResolverImpl(
...
-                return getKSTypeCached(resolveJavaType(type.psi, type), type.element.typeArguments, type.annotations)
+                val resolved = getKSTypeCached(
+                    resolveJavaType(type.psi, type),
+                    type.element.typeArguments,
+                    type.annotations
+                )
+                return if (type.psi is PsiArrayType) {
+                    resolved
+                } else {
+                    resolved.replace(type.element.typeArguments)
```
neetopia, line 749: fix for #1453 is at this line. However introducing this line causes issues elsewhere due to array issue mentioned in this PR. As the issue on array turned out to be a bigger issue the PR is named after that.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -738,7 +738,16 @@ class ResolverImpl(
...
+                }
```
ting-yuan, line 750: I'm a little bit confused. Why do you special-case Array here?

Also, I'm worried that replacing the type arguments in this way would effectively change the semantics of platform types.

ting-yuan (approved):

Issue #1136, opened 2022-10-05
Support analyzing context 


ShawxingKwok:
The syntax 'context' has been provided for a long time but can't be analyzed yet.

neetopia:
KSP should support context receiver feature.

neetopia:
As context receiver is still an experiment feature, KSP will hold on the implementation until it becomes stable in Kotlin.

ting-yuan:
We plan to support this later. Could we leave this open?

dtdharan:
Poor

mos8502:
As it seem that context receivers is going to be supported in Kotlin 2.x and as the feature has been out for preview for quite some time, it would be great to start "experimenting" with it in KSP before the design of the feature is finalised

apatrida:
This is indeed nearing stable release and frameworks could start developing against it using KSP if it became available.

PR #1475, opened 2023-07-22
[WIP] Support analyzing context receivers

apatrida:
Add analysis for context receivers (experimental up thought Kotlin 1.9 but should be final in Kotlin 2.0)

This basically clones code for extension receivers and adds support for context receivers in the same classes and interfaces.  The testing is mirroring the same level of support for extension receivers (which is minimally covered in tests).  

This is to address issue #1136 

TODO:

* [X] constructors with context receivers
* [ ] open questions below
* [ ] top level functions are not currently covered by tests for Extension Receiver nor for this new context receivers, should they be?

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSClassDeclaration.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSClassDeclaration.kt
@@ -74,4 +74,9 @@ interface KSClassDeclaration : KSDeclaration, KSDeclarationContainer {
      * @return A type with all type parameters applied with star projection.
      */
     fun asStarProjectedType(): KSType
+
+    /**
+     * The class may have context receivers at the class level, which affect all constructors
+     */
+    val contextReceivers: List<KSTypeReference>
 }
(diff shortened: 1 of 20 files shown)


apatrida:
Question for opinions:

A context receiver on the class level applies to the constructors, but is written at the class level.  So should the context receivers be exposed in the class, or via each constructor as they would be on other member functions or top-level functions?

apatrida:
Question for opinions:

A context receiver can be on functions, methods, classes.  Should they therefore be available from `KSDeclaration` and return emptyList for properties and typealiases, or should they bely on classes and function declarations.

apatrida:
Question to resolve:

Does https://github.com/tschuchortdev/kotlin-compile-testing have a way to set freeArgs or `-Xcontext-receivers` in some way for the compiler?  This change here could impact that until this feature is stable and the flag is removed (Kotlin 2.0)

apatrida:
Hello, any feedback on this approach?  Nearly stable release for feature in Kotlin, and this is probably pretty close to what is needed.

Ferdia-Fagan:
I am just going to poke this because I have been waiting for this for so long ... Any success with this? or is it still in experimental?

jaschdoc:
Closing this, but context parameters are planned.

PR #1480, opened 2023-07-28
Handle Kotlin source sets from Kotlin compilations more lazily

sellmair:


neetopia (approved):
verified with https://github.com/JuBan1/ksp-repro

PR #1494, opened 2023-08-15
multiple-round: implement incremental providers

ting-yuan:
and replace StandaloneAnalysisAPISessionBuilder, which hides KotlinCoreProjectEnvironment, which is needed to update KotlinCliJavaFileManagerImpl for new files.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/analysisapi/providers/IncrementalKotlinDeclarationProvider.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/analysisapi/providers/IncrementalKotlinDeclarationProvider.kt
@@ -1,0 +1,105 @@
+package com.google.devtools.ksp.analysisapi.providers
+
+import com.intellij.openapi.project.Project
+import com.intellij.psi.search.GlobalSearchScope
+import org.jetbrains.kotlin.analysis.project.structure.KtModule
+import org.jetbrains.kotlin.analysis.providers.KotlinDeclarationProvider
+import org.jetbrains.kotlin.analysis.providers.KotlinDeclarationProviderFactory
+import org.jetbrains.kotlin.analysis.providers.impl.KotlinStaticDeclarationProviderFactory
+import org.jetbrains.kotlin.name.CallableId
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.psi.KtClassLikeDeclaration
+import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.KtScript
+import org.jetbrains.kotlin.psi.KtTypeAlias
+
+class IncrementalKotlinDeclarationProvider(var del: KotlinDeclarationProvider) : KotlinDeclarationProvider() {
+    override fun computePackageSetWithTopLevelCallableDeclarations(): Set<String> {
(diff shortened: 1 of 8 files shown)


test-utils/src/main/kotlin/com/google/devtools/ksp/processor/ConstructorDeclarationsProcessor.kt
```diff
@@ -24,9 +24,10 @@ import com.google.devtools.ksp.symbol.*
 
 class ConstructorDeclarationsProcessor : AbstractTestProcessor() {
     val visitor = ConstructorsVisitor()
+    lateinit var result: List<String>
 
     override fun toResult(): List<String> {
-        return visitor.toResult()
+        return result
```
neetopia, line 30: what's this change for?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/analysisapi/providers/IncrementalKotlinDeclarationProvider.kt
```diff
@@ -0,0 +1,105 @@
...
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.psi.KtClassLikeDeclaration
+import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.psi.KtScript
+import org.jetbrains.kotlin.psi.KtTypeAlias
+
+class IncrementalKotlinDeclarationProvider(var del: KotlinDeclarationProvider) : KotlinDeclarationProvider() {
```
neetopia, line 21: if we are purely delegating here, it might make more sense to simply use delegation without need to write down all the redundant overrides here.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -71,77 +354,239 @@ class KotlinSymbolProcessing(
...
+        // TODO: performance
+        val project = analysisAPISession.project
+        val psiManager = PsiManager.getInstance(project)
+        var finished = false
+        var initialized = false
+        lateinit var codeGenerator: CodeGeneratorImpl
+        lateinit var processors: List<SymbolProcessor>
+        lateinit var newKSFiles: List<KSFile>
+        lateinit var newFileNames: Set<String>
+        while (!finished) {
```
neetopia, line 372: are we falling back to the previous design for multiple round processing? Just want to know if this conflicts with the plan to do multiple round within Gradle runner.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/ConstructorDeclarationsProcessor.kt
ting-yuan: Some KtSymbols are not available after cache drop

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/analysisapi/providers/IncrementalKotlinDeclarationProvider.kt
ting-yuan: Unfortunately, `by` delegation doesn't allow swapping the delegate.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
ting-yuan: IIRC the plan is to do it here, in the unified entry point of production and test. There will be multilple callers in addition to gradle and we want the setup to be as easy as possible.

neetopia (approved):

PR #1495, opened 2023-08-17
multiple-round: implement deferred symbols 

ting-yuan:
by introducing Deferrable/Restorable and using KtSymbolPointer.
KtSymbolPointer handles quite a lot corner cases that we'd have to do by
ourselves if using psi directly.

TODOs:
1. Java elements are not supported by KtSymbolPointer yet. We either
   need to do it in the upstream or implement with psi in KSP.
2. KSTypeReference and KSTypeArgument are implemented upon KtType, which
   a. can change in the subsequent rounds, and
   b. is not KtSymbol and has no KtSymbolPointer.


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFileImpl.kt
```diff
@@ -93,4 +93,15 @@ class KSFileImpl private constructor(private val ktFileSymbol: KtFileSymbol) : K
     override fun toString(): String {
         return "File: ${this.fileName}"
     }
+
+    override fun defer(): Restorable {
+        val psi = this.psi
+        return Restorable {
+            when (psi) {
+                is KtFile -> analyze { getCached(psi.getFileSymbol()) }
+                is PsiJavaFile -> null
```
neetopia, line 102: when will we hit this line now that we have separate KSFileJavaImpl for java files?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
```diff
@@ -448,3 +448,21 @@ internal fun KtValueParameterSymbol.getDefaultValue(): KtAnnotationValue? {
...
+fun interface Restorable {
+    fun restore(): KSAnnotated?
+}
+
+fun interface Deferrable {
+    fun defer(): Restorable?
+}
+
+fun <T : KtSymbol> T.defer(restore: (T) -> KSAnnotated?): Restorable? {
+    val ptr = analyze { with(this) { this@defer.createPointer() } }
```
neetopia, line 461: IIRC back to early days I had some discussion with JB regarding the usability of symbol pointers and my impression was that this might not suite best for our case, have you chatted with them on using pointers?

ting-yuan: Would you mind elaborating?

neetopia (approved):

PR #1511, opened 2023-08-29
Kotlin: Migrate from ChangedFiles to SourcesChanges

ALikhachev:
In `master` of Kotlin there's a massive change in KGP removing a lot of dependencies KGP -> Kotlin compiler modules (https://github.com/JetBrains/kotlin/compare/e68661fc0dbb2851feff227939a84c9eb0d65517...829e0675f4ebe163506c99b15611b3c64dfab026), removing many of them both from compilation classpath and runtime classpath. One of such changes affecting KSP is replacing `org.jetbrains.kotlin.incremental.ChangedFiles` by `org.jetbrains.kotlin.buildtools.api.SourcesChanges`. It's a similar hierarchy but located in another module which KGP depends on and will depend on. This PR restores compatibility with KGP.

PR #1526, opened 2023-09-07
UPDATE_KOTLIN_VERSION: 1.9.30-dev-2548

neetopia:
first commit is cherry-picked from #1511 with updated commit message to ensure auto-merge works.
while I separated commits for AA and FE1.0, it is worth mentioning that due to the dependency to the `kotlinBaseVersion` in `common-util` for testing framework, updating AA version alone will cause unit test run time error due to version mismatch in test utils.


gradle.properties
```diff
@@ -9,7 +9,7 @@ junit5Version=5.8.2
 junitPlatformVersion=1.8.2
 googleTruthVersion=1.1
 
-aaKotlinBaseVersion=1.9.30-dev-1098
+aaKotlinBaseVersion=1.9.30-dev-2548
```
ting-yuan, line 12: Can you reword the commit message, so that it can be auto-merged?

neetopia: can you elaborate? all 3 commits have same commit message, I tried to run the shell command locally and can have all 3 commits filtered.

ting-yuan: The commit for AA should be cherry-picked automatically. With `UPDATE_KOTLIN_VERSION` in commit message it will be filtered out.

neetopia: unfortunately due to the dependency issues in `common-util` mentioned in the description, AA commit has to be excluded from auto-merge.

ting-yuan: KSPAATest in the release branch is redundant. Let's disable it.

Talking about tests, we need to enable KSP 2 in gradle tests and integration tests. Could you update CI to include KSP 2.0 in the matrix in a separate PR?

neetopia:
updated commit message for aa version update commit.

ting-yuan (approved):

PR #1528, opened 2023-09-11
Refactor dependency

neetopia:


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspOptions.kt
```diff
@@ -0,0 +1,344 @@
+/*
```
ting-yuan, line 1: Can we remove this file and KSPCommandLineProcessor.kt together? The latter is not used and the new entry point has moved away from KspOptions.

ting-yuan (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspOptions.kt
neetopia: Good catch, removed unused classes.

PR #1540, opened 2023-09-19
Qualified name and flexible type rendering for java types.

neetopia:


test-utils/testData/api/resolveJavaType.kt
```diff
@@ -60,12 +60,6 @@
 // END
 // FILE: a.kt
 annotation class Test
-@Test
```
neetopia, line 63: This part seems to be unrelated to the test as removing them does not cause issues. In fact, putting them here actually causes issues in analysis API test because these declarations are a duplicating declaration with the actual classes we want to test in Java sources, FE1.0 was not seeing this issue because it looks legit from PSI manager (duplicate declarations are from Kotlin).

ting-yuan (approved):

PR #1547, opened 2023-09-21
KSP 2: excludeProcessor, JvmPlatformInfo, build cache

ting-yuan:
Implemented the above features

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
@@ -10,13 +10,18 @@ import org.gradle.api.logging.LogLevel
 import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.SetProperty
+import org.gradle.api.tasks.CacheableTask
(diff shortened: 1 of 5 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -128,6 +136,19 @@ abstract class KspAATask @Inject constructor(
                     }
                     cfg.logLevel.value(logLevel)
                     cfg.allWarningsAsErrors.value(kspExtension.allWarningsAsErrors)
+                    cfg.excludedProcessors.value(kspExtension.excludedProcessors)
+
+                    val jvmDefaultMode = project.provider {
+                        compilerOptions.freeCompilerArgs.get().lastOrNull {
```
neetopia, line 142: is `lastOrnull` a standard solution for duplicate arguments?

neetopia (approved):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
ting-yuan: It's a common practice for argument parsing. The latter overrides the former.

PR #1553, opened 2023-09-22
fork test for parent node.

neetopia:
New behavior changes a lot in terms of synthetic members and ordering,
it is infeasible to adjust test processor logic to get tests aligned.


ting-yuan (approved):

neetopia:
Failed test in Linux CI might be JDK dependent, I tried to run same test locally in Linux but was unable to reproduce the error.

ting-yuan:
Looks like an issue of visit order. Sort the files before visiting.

PR #1561, opened 2023-10-06
isolate class loader inside KSPAATask

neetopia:


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -76,7 +81,11 @@ abstract class KspAATask @Inject constructor(
             val sourceSetName = kotlinCompilation.defaultSourceSet.name
             val kspTaskName = kotlinCompileProvider.name.replaceFirst("compile", "ksp")
             val kspAADepCfg = project.configurations.detachedConfiguration(
-                project.dependencies.create("${KspGradleSubplugin.KSP_GROUP_ID}:symbol-processing-aa:$KSP_VERSION")
+                project.dependencies.create("${KspGradleSubplugin.KSP_GROUP_ID}:symbol-processing-aa:$KSP_VERSION"),
+                project.dependencies.create(
+                    "${KspGradleSubplugin.KSP_GROUP_ID}:symbol-processing-gradle-plugin:$KSP_VERSION"
+                ),
+                project.dependencies.create("org.jetbrains.intellij.deps:trove4j:1.0.20200330"),
```
neetopia, line 88: no idea where this comes from, call site looks like intellij deps to me but intellij jar relocated the symbols therefore has to use this weired version.

ting-yuan: Shouldn't trove4j be packed into the fat jar already?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -76,7 +81,11 @@ abstract class KspAATask @Inject constructor(
             val sourceSetName = kotlinCompilation.defaultSourceSet.name
             val kspTaskName = kotlinCompileProvider.name.replaceFirst("compile", "ksp")
             val kspAADepCfg = project.configurations.detachedConfiguration(
-                project.dependencies.create("${KspGradleSubplugin.KSP_GROUP_ID}:symbol-processing-aa:$KSP_VERSION")
+                project.dependencies.create("${KspGradleSubplugin.KSP_GROUP_ID}:symbol-processing-aa:$KSP_VERSION"),
+                project.dependencies.create(
+                    "${KspGradleSubplugin.KSP_GROUP_ID}:symbol-processing-gradle-plugin:$KSP_VERSION"
```
ting-yuan, line 86: Why do you need this? It'll also fetch all runtime dependencies of KSP Gradle plugin and can be a surprise in the future.

ting-yuan: Also, I'd expect Kotlin stdlib and maybe reflection / etc to be listed here now that we're using `getPlatformClassLoader()`. If you're adding KSP Gradle plugin for this purpose, let's state the standard libraries explicitly. They should be available from Gradle and also KGP. Make sure you use either of them so that it doesn't end up downloading the third version.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -59,6 +63,7 @@ abstract class KspAATask @Inject constructor(
         }
```
ting-yuan, line 63: Should be able to replace this with noIsolation now, for better performance.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -307,17 +321,32 @@ abstract class KspAAWorkerAction : WorkAction<KspAAWorkParameter> {
...
             jvmDefaultMode = gradleCfg.jvmDefaultMode.get()
         }.build()
+        val byteArrayOutputStream = ByteArrayOutputStream()
+        val objectOutputStream = ObjectOutputStream(byteArrayOutputStream)
+        objectOutputStream.writeObject(kspConfig)
 
         val exitCode = try {
-            KotlinSymbolProcessing(kspConfig).execute()
+            val kspLoaderClass = isolatedClassLoader.loadClass(KSPLoader::class.java.canonicalName)
+            val kspLoader = kspLoaderClass.newInstance()
```
ting-yuan, line 336: nit: Let's make `loadAndRunKSP` a static function and use `@JvmStatic` to force its name. Then we can avoid instantiating `KspLoader`.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KSPConfig.kt
```diff
@@ -130,7 +121,7 @@ class KSPJvmConfig(
     allWarningsAsErrors,
     mapAnnotationArgumentsInJava,
 ) {
-    class Builder : KSPConfig.Builder() {
+    class Builder : KSPConfig.Builder(), Serializable {
```
ting-yuan, line 124: May I know why `Builder` needs to be serializable too?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
neetopia: No, the one found in fat jar does not sits in `gnu` package name, that's the relocation issue I am referring to.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
neetopia: This is needed because we put the `KSPLoader` in gradle plugin code. `KSPGradleLogger` is used in `KSPLoader` so we need to add it here. Alternatively we can move all `KSPGradleLogger` away if we don't want KSP Gradle plugin in the classpath.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
neetopia: Actually changes to `noIsolation` causes runtime errors inside KSPAATask itself due to classes like `KSPLogger` not found.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KSPConfig.kt
neetopia: Removed.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
ting-yuan: Can you please check how it is included in the compiler. If we need this, we should pack it and remove the one we already have. Having both sounds wrong to me.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
ting-yuan: Should we move KSPLoader into `kotlin-analysis-api` then? Probably into `KotlinSymbolProcessing`.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
ting-yuan: That is actually an indication of missing libraries in the classpath: `symbol-processing-api`

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
neetopia: This class is relocated multiple times by different consumers, e.g. Gradle also relocated this class under its own name. What's causing trouble here is the non relocated version got involved, and is not part of any fat jar since fat jars all do relocations.

ting-yuan (approved):

PR #1566, opened 2023-10-11
introduce gradle-util module for classpath isolation

neetopia:


gradle-plugin/build.gradle.kts
```diff
@@ -28,10 +28,11 @@ dependencies {
     // replace AGP dependency w/ gradle-api when we have source registering API available.
     compileOnly("com.android.tools.build:gradle:$agpBaseVersion")
     compileOnly(gradleApi())
-    implementation(project(":kotlin-analysis-api")) {
+    compileOnly(project(":kotlin-analysis-api")) {
         isTransitive = false
     }
-    implementation(project(":api"))
+    api(project(":api"))
```
ting-yuan, line 34: compileOnly

gradle-plugin/build.gradle.kts
```diff
@@ -28,10 +28,11 @@ dependencies {
     // replace AGP dependency w/ gradle-api when we have source registering API available.
     compileOnly("com.android.tools.build:gradle:$agpBaseVersion")
     compileOnly(gradleApi())
-    implementation(project(":kotlin-analysis-api")) {
+    compileOnly(project(":kotlin-analysis-api")) {
         isTransitive = false
```
ting-yuan, line 32: remove isTransitive

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -58,7 +58,7 @@ abstract class KspAATask @Inject constructor(
         //  1. passing arguments by data structures in stdlib, or
         //  2. hoisting and publishing KspGradleConfig into another package.
 
-        val workerQueue = workerExecutor.noIsolation()
+        val workerQueue = workerExecutor.classLoaderIsolation()
```
ting-yuan, line 61: Isn't the purpose of the extra dependency removing classLoaderIsolation?

ting-yuan (commented):
Can we name the module / artifact / package `common` or `common-deps` or etc, since it is not gradle specific and will be needed in command line and other build systems.

gradle-plugin/build.gradle.kts
neetopia: KSPLogger is still needed at runtime classpath, actually `implementation` makes more sense here.

ting-yuan (approved):

Issue #1568, opened 2023-10-12
KSP resolves dependencies too early


yigit:
We are unable to use KSP in AndroidX Multiplatform project.

Here is the repro:

* Checkout androidx, then apply this CL:

https://android-review.googlesource.com/c/platform/frameworks/support/+/2787012
* Run:
```
ANDROIDX_PROJECTS=INFRAROGUE gw :room:integration-tests:room-nativetestapp:macOsArm64Test
```

[Build scan](https://ge.androidx.dev/s/nhgs4gzu365jk/failure?expanded-stacktrace=WyIwIl0#1)

It will fail with an error which is caused by KSP resolving configurations early ([my guess](https://github.com/google/ksp/blob/main/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L540))

I think KSP should be configuring its tasks with a provider so that the configuration is only resolved when the task is needed, not when it is configured. The early resolution causes some other AndroidX code to fail so I'm not 100% sure if this is just a KSP issue or an integration problem. But it seems like KSP should avoid resolving these collections at configuration time as they may not necessarily have the files ready yet.

ting-yuan:
I haven't been able to reduce the test case to confirm the root cause yet but commenting out the line you pointed out did make the error away.

The fix could be a bit tricky though because, unlike other platforms, `KotlinNativeCompile` doesn't have any execution time function intended for public override. That is why it is resolved at configuration time. I'll need to talk to upstream before applying workarounds.

yigit:
I played with this a bit, have a possible solution:
https://github.com/yigit/ksp/commit/7af4fb58b041a0aaa619dd699ad47519a0d74b48

This branch is off of v13 so it might be a bit different but there were 2 problems:

* `FilesSubpluginOption`: Gets the files in the classpath. I'm not sure if this really resolves but rather it might be just out of date when task runs. Created a `ClasspathFilesSubpluginOption` that will resolve it when needed.
* `val kspOptions = kspTask.options.get().flatMap { listOf("-P", it.toArg()) }`: You can "almost" never call `.get` on a property during configuration as you don't necessarily know that it is complete. The solution is rather easy, freeArgs is already a ListProperty which accepts `provider` so I'm passing a provider instead:
```
kspTask.compilerOptions.freeCompilerArgs.addAll(
                            kspTask.options.map {
                                it.flatMap { listOf("-P", it.toArg()) }
                            }
                        )
```


[...]

PR #1571, opened 2023-10-16
Avoid resolving classpaths eagerly

yigit:
This PR update KspSubplugin to resolve arguments lazily and also adds a workaround for KonanTarget serialization problem.

These are the 3 issues fixed:

* FilesSubpluginOption: Calling `FilesSubpluginOption("apclasspath", processorClasspath.files.toList())` requires resolving that classpath at configuration time, where it may not necessarily be ready for resolution. Replaced it with a `FileCollectionSubpluginOption` that will resolve it lazily. 
* options.get: Ksp was calling `options.get` at configuration time, which again causes early resolution. Furthermore, it won't change if options are later changed, which is a problem. Changed it to use the provider API.
* configuration cache: KSP has a `task.onlyIf` block to check if the KonanTarget can be compiled on this machine. Unfortunately, KonanTarget has a bug which makes it not properly serializable. (https://youtrack.jetbrains.com/issue/KT-61657) . Changed the block to use the name instead.

Fixes: #1568 
Test: testConfigurationsAreNotResolvedAtConfigurationTime for classpath problem
Test: testArgumentsAreNotResolvedAtConfigurationTime for options.get at configuration time
Test: applicationCanAccessGeneratedCode_multiplatform_withConfigCache for the KonanTarget fix

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinFactories.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinFactories.kt
@@ -18,7 +18,6 @@
 @file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
 
 package com.google.devtools.ksp.gradle
-
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.file.FileCollection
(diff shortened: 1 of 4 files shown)


ting-yuan (approved):

ting-yuan:
Thanks a log for the fixes!

Issue #1536, opened 2023-09-14
KSP plugin is incompatible with Kotlin language version 2.0 


ksiarhey:
Steps to reproduce: 
1. Build KSP gradle plugin from main branch and publish to mavenLocal 
2. Apply the built before KSP gradle plugin in a tested project 
3. Update KGP version in the tested project to the latest from master (1.9.20-dev-3093)
4. Enable kotlin languageVersion = 2.0 
5. Build the project 
Actual Results: failure 
```
e: There are some plugins incompatible with language version 2.0:
  com.google.devtools.ksp.KotlinSymbolProcessingComponentRegistrar
Please use language version 1.9 or below
```
Probably [this](https://github.com/JetBrains/kotlin/commit/b0b6a6a05aea614de33e91119d4540755eaa39f4) commit causes such an incompatibility

neetopia:
are you testing with 1.9.30-dev-3093? I don't think there is a 1.9.20-dev-3093.

ting-yuan:
@ksiarhey could you share more error messages? In particular, it'd be helpful to know whether it is failing in ksp's task or kotlinc's.

ksiarhey:
kotlinc's tasks are failing in modules where ksp plugin applied as the compiler with LV 2.0 enabled regards the KSP plugin as compatible anymore.
I hurried to submit the issue prior to the release of the upcoming bootstrap build.

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':service:compileFossDebugKotlin'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:147)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:145)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:133)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
[...]

PR #1574, opened 2023-10-18
Be more compatible with K2

ting-yuan:
The compiler plugins are always loaded unconditioanlly regardless of K1 or K2 and causes FirKotlinToJvmBytecodeCompiler to throw. Because there seems no way to load compiler plugins conditionally on language version, we declare the KSP compiler plugin to be K2 compatible while doing nothing if K2 is detected.

Tested locally with a new enough Kotlin compiler.


ting-yuan:
@Tapchicoma

ting-yuan:
Fixes #1536 for future Kotlin versions

neetopia (approved):

Issue #1569, opened 2023-10-16
1.9.20-RC-1.0.13 breaks Default Kotlin Hierarchy Template 


rickclephas:
`1.9.20-RC-1.0.13` includes the [fix](https://github.com/google/ksp/pull/1558) for #1529. This however breaks the Default Kotlin Hierarchy Template with the following warning:
```
w: The Default Kotlin Hierarchy Template was not applied to 'project ':shared'':
Explicit .dependsOn() edges were configured for the following source sets:
[iosArm64Main, iosSimulatorArm64Main, iosX64Main, macosArm64Main, macosX64Main, tvosArm64Main, tvosSimulatorArm64Main, tvosX64Main, watchosArm32Main, watchosArm64Main, watchosSimulatorArm64Main, watchosX64Main]

Consider removing dependsOn-calls or disabling the default template by adding
    'kotlin.mpp.applyDefaultHierarchyTemplate=false'
to your gradle.properties
```

Resulting in ignored intermediate source-sets (in this case `iosMain`, `appleMain`, etc).

Explicitly applying the default hierarchy template fixes this:
```kotlin
kotlin {
    applyDefaultHierarchyTemplate()
}
```

rickclephas:
Just tested with `1.9.20-RC2-1.0.13` and I can confirm this has been fixed.

PR #1575, opened 2023-10-20
Add generated files to target source sets

ting-yuan:
instead of creating source sets and making default source sets depend on them.

This avoids breaking kotlin.mpp.applyDefaultHierarchyTemplate.

Common source sets are untouched, because otherwise the generated sources for them would be observed by downstream compilations and be inconsistent with current KMP build scheme.

Alternatively, we could add all generated files, common or target, directly to compilation tasks and keep default source sets untouched. The drawback is that generated files won't be seen by IDE. We would have to build IDE plugins and implement our own model builder to register generated files to IDE.

TODO: Add common source sets to their corresponding compilation for the new KMP build scheme.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -37,7 +37,6 @@ import org.gradle.util.GradleVersion
 import org.jetbrains.kotlin.buildtools.api.SourcesChanges
 import org.jetbrains.kotlin.config.ApiVersion
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
-import org.jetbrains.kotlin.gradle.dsl.kotlinExtension
 import org.jetbrains.kotlin.gradle.internal.kapt.incremental.CLASS_STRUCTURE_ARTIFACT_TYPE
 import org.jetbrains.kotlin.gradle.internal.kapt.incremental.ClasspathSnapshot
 import org.jetbrains.kotlin.gradle.internal.kapt.incremental.KaptClasspathChanges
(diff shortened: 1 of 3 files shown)


ting-yuan:
Fixes #1569

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -565,16 +527,19 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
...
-            // They will be observed by downstreams and confuse processors.
-            kotlinCompileProvider.configure {
-                it.source(kspGeneratedSourceSet.kotlin)
-            }
+            // Do not add generated sources to common source sets.
+            // They will be observed by downstreams and violate current build scheme.
+            kotlinCompileProvider.configure { it.source(*generatedSources) }
         } else {
-            kotlinCompilation.defaultSourceSet.dependsOn(kspGeneratedSourceSet)
+            kotlinCompilation.defaultSourceSet.kotlin.srcDirs(*generatedSources)
```
gavra0, line 540: Let's look into adding an API in KGP that allows:
- registering source-generating tasks
- getting all sources (without the sources you've just generated)
?

ting-yuan: That'll be awesome. Before the API is available in KGP, I guess I'll have to choose between:
1. this PR, that needs to iterate through all sources, or
2. give up IDE support

Whether making generated sources available to other Gradle tasks is also a difference.

ting-yuan:
I'm going to merge this as the previous approach has bigger issues. This is also needed in 1.0.13 for the upcoming Kotlin 1.9.20.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
gavra0: Can you please file a YT issue and I'll raise this with JB?

gavra0:
LGTM 👍

PR #1585, opened 2023-10-31
isolate static members into thread local

neetopia:


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -262,14 +262,20 @@ interface KspAAWorkParameter : WorkParameters {
     var kspClassPath: ConfigurableFileCollection
 }
 
+var isolatedClassLoaderCache = mutableMapOf<String, URLClassLoader>()
```
ting-yuan, line 265: should this be synchronized?

ting-yuan (approved):

PR #1592, opened 2023-11-03
Make hierarchy in KSP API sealed (attempt 2)

lukellmann:
Another try of #1380 with additional compilation fixes.


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSDeclarationDescriptorImpl.kt
```diff
@@ -30,21 +30,22 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.parents
 import org.jetbrains.kotlin.resolve.descriptorUtil.parentsWithSelf
 import org.jetbrains.kotlin.utils.addToStdlib.firstIsInstanceOrNull
 
-abstract class KSDeclarationDescriptorImpl(private val descriptor: DeclarationDescriptor) : KSDeclaration {
+abstract class KSDeclarationDescriptorImpl(private val descriptor: DeclarationDescriptor) /*: KSDeclaration*/ {
 
-    override val origin by lazy {
+    open /*override*/ val origin by lazy {
```
ting-yuan, line 35: I'm a bit concerned with this. Having to cast the implementations is not only inconvenient, but also error prone and hard to maintain. Is there a better way to do it?

lukellmann: this approach was the one with minimal changes. another one i can think of is to have these abstract classes but override and upcall in all implementation classes. this would be a bigger change but also less error prone.

lukellmann: what about this: https://github.com/google/ksp/pull/1592/commits/cda80566bd861994cd4208d2feff98ab1d9010a2

ting-yuan:
After thinking this again, I feel it might not be right to make the API sealed. The class hierarchy may be expanded in the future to reflect language changes. Making them sealed will make it very tricky. Could you justify why sealing it is the way to go?

lukellmann:
Like described in https://github.com/google/ksp/issues/1351, the motivation was to be able to have the compiler enforce code changes when the hierchy is changed. So changes in the hierachy leading to compiler errors in sealed `when`s is the desired outcome.

lukellmann:
> [...] The class hierarchy may be expanded in the future to reflect language changes. Making them sealed will make it very tricky. [...]

To be clear, you are talking about the API side facing the KSP user here, right? Or do you also mean something else?

ting-yuan:
Making the API sealed only helps discover source level incompatibility. From library users' point of view, sealed interface / class isn't truly sealed. A unhandled case can still arise in an exhaustive-when in the user code. If `when` is used as a statement, it could also become a silent error. I don't know how to alter the hierarchy of a sealed class in the API without potentially breaking users' code which relies on exhaustive-when.

It'd also be tricky when API users try to be future compatible; If their `when` is already exhaustive, adding an else clause would trigger a compiler warning.

lukellmann:
> Making the API sealed only helps discover source level incompatibility. From library users' point of view, sealed interface / class isn't truly sealed. A unhandled case can still arise in an exhaustive-when in the user code.

And that's fine. The goal is to make writing the code easier. An unhandled case at runtime will throw and clearly tell what went wrong - the processor isn't built for handling this new feature.

> If `when` is used as a statement, it could also become a silent error.

How would it be silent? Non-exhaustive `when` statements on sealed types report an error [since Kotlin 1.7.0](https://kotlinlang.org/docs/compatibility-guide-17.html#make-when-statements-with-enum-sealed-and-boolean-subjects-exhaustive-by-default). So either you run into the `else` you are required to write or you get an exception if you didn't recompile after a new feature was added.

> I don't know how to alter the hierarchy of a sealed class in the API without potentially breaking users' code which relies on exhaustive-when.

Tools working on source code (like KSP processors) will always break on new language features, in one way or another. The whole point is to get help by the compiler when things change. Visitors can still be used when preferred.

> It'd also be tricky when API users try to be future compatible; If their `when` is already exhaustive, adding an else clause would trigger a compiler warning.

I'd argue if a user really wants to be future compatible to this extent, a suppression would be ok.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -318,8 +320,8 @@ class ResolverAAImpl(
         val kotlinClass = classBinaryCache.getKotlinBinaryClass(virtualFile) ?: return container.declarations
         val declarationOrdering = DeclarationOrdering(kotlinClass)
 
-        return (container.declarations as? Sequence<AbstractKSDeclarationImpl>)
-            ?.sortedWith(declarationOrdering.comparator) ?: container.declarations
+        return container.declarations
+            .sortedWith(compareBy(declarationOrdering.comparator) { it as AbstractKSDeclarationImpl })
```
lukellmann, line 324: explanation:
This was an unchecked cast (generics involved) and would always succeed - so replacing `as?` with `as` shouldn't change any behavior.

lukellmann:
What's the state of this PR, @ting-yuan? Does the proposed change have a chance to be merged?

lukellmann:
also: should i merge or rebase + force-push to resolve conflicts?

jaschdoc:
Hi, I'm closing this PR as it is marked stale. From what I understand, we have moved away from a sealed API.

PR #1610, opened 2023-11-19
Optimize

GeorgCantor:
**Using lastIndexOf:**
Instead of splitting the string into an array and then joining it back together, we find the index of the last period (.) in the string using lastIndexOf. This is more efficient as it avoids unnecessary array creation.

**Using substring:**
If a period is found (lastIndex != -1), we use substring to extract everything before the last period. This is a direct operation on the string and avoids the overhead of creating an intermediate list.

**Handling Edge Cases:**
If there is no period in the string, we return an empty string. This ensures that our function behaves predictably regardless of input.


IsakTheHacker (approved):

glureau (approved):

ting-yuan:
Thanks for the patch, @GeorgCantor. Would you mind to rebase it?

PR #1632, opened 2023-12-01
Publish symbol-processing-aa-embeddable

ting-yuan:


ting-yuan:
And also include sources from selected dependencies for the fat jars.

Issue #1482, opened 2023-07-31
KSP processing fails with Java enum


dstepanov:
Originally reported here https://github.com/micronaut-projects/micronaut-core/issues/9642

I found a few different exceptions:
```
e: [ksp] java.lang.IllegalStateException: unexpected class: class com.google.devtools.ksp.symbol.impl.java.KSClassDeclarationJavaEnumEntryImpl
        at com.google.devtools.ksp.processing.impl.ResolverImpl.resolveClassDeclaration(ResolverImpl.kt:568)
        at com.google.devtools.ksp.processing.impl.ResolverImpl.mapToJvmSignatureInternal$compiler_plugin(ResolverImpl.kt:356)
        at com.google.devtools.ksp.processing.impl.ResolverImpl.mapToJvmSignature(ResolverImpl.kt:353)
        at io.micronaut.kotlin.processing.ExtensionsKt.getBinaryName(extensions.kt:47)
        at io.micronaut.kotlin.processing.visitor.KotlinClassElement$KotlinEnclosedElementsQuery.getElementName(KotlinClassElement.kt:628)
        at io.micronaut.kotlin.processing.visitor.KotlinClassElement$KotlinEnclosedElementsQuery.getElementName(KotlinClassElement.kt:619)
        at io.micronaut.inject.ast.utils.EnclosedElementsQuery.getAllElements(EnclosedElementsQuery.java:217)
```

```
e: [ksp] /Users/denisstepanov/dev/micronaut-core/test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/core/beans/JavaEnum.java:3: Member must have a containing class: PsiMethod:valueOf
e: java.lang.AssertionError: Member must have a containing class: PsiMethod:valueOf
        at org.jetbrains.kotlin.load.java.structure.impl.JavaMemberImpl.getContainingClass(JavaMemberImpl.java:63)
        at com.google.devtools.ksp.symbol.impl.UtilsKt.resolveContainingClass(utils.kt:235)
        at com.google.devtools.ksp.processing.impl.ResolverImpl.resolveJavaDeclaration(ResolverImpl.kt:537)
        at com.google.devtools.ksp.processing.impl.ResolverImpl.resolveFunctionDeclaration(ResolverImpl.kt:577)
        at com.google.devtools.ksp.symbol.impl.java.KSFunctionDeclarationJavaImpl.findOverridee(KSFunctionDeclarationJavaImpl.kt:56)
[...]

PR #1635, opened 2023-12-04
Fix resolving JVM signature for Java enum

dstepanov:
Fixes https://github.com/google/ksp/issues/1482


test-utils/src/main/kotlin/com/google/devtools/ksp/processor/MapClassSignatureProcessor.kt
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2020 Google LLC
```
neetopia, line 2: 2023?

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/MapClassSignatureProcessor.kt
```diff
@@ -0,0 +1,42 @@
...
+
+package com.google.devtools.ksp.processor
+
+import com.google.devtools.ksp.KspExperimental
+import com.google.devtools.ksp.getClassDeclarationByName
+import com.google.devtools.ksp.processing.Resolver
+import com.google.devtools.ksp.symbol.KSAnnotated
+
+@KspExperimental
+class MapClassSignatureProcessor : AbstractTestProcessor() {
```
neetopia, line 26: We have `MapSignatureProcessor` where you can simply add an enum class there and add it to the visit list.

dstepanov: That processor also tests methods, and the results don't match between two tests, so I had to create a new one.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/MapClassSignatureProcessor.kt
dstepanov: Fixed

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/MapClassSignatureProcessor.kt
neetopia: Did you mean the result don't match between compiler-plugin and kotlin-analysis-api tests?

dstepanov: Yes

neetopia: some difference between the 2 is acceptable depending on the case, it's better to just split test data rather than writing a seemingly unnecessary new test. IIRC there is some difference around enum member functions, if that's the case, it should be ok to just split test data.

dstepanov: Ok, I did split the test data, please check

neetopia (approved):
Thanks for PR, will merge once CI passes.

neetopia:
you can expect a release in 2 weeks.

PR #1680, opened 2024-01-16
fix java type resolution for checked exception

### LINKED ISSUE #1460: `Resolver#getJvmCheckedException` results in `<ERROR TYPE>` when throwing type variable.
To reproduce this, define the following java class with a method that throws a type variable:

```java
class MyClass {
  <T extends Throwable> void method() throws T {}
}
```

Alternatively, the type variable can also be defined on the class:

```java
class MyClass<T extends Throwable> {
  void method() throws T {}
}
```

Then inspect the thrown types:

```kotlin
val myClass = resolver.getClassDeclarationByName(resolver.getKSNameFromString("MyClass"))
myClass?.getAllFunctions()?.forEach { method ->
  println("${method}: ${resolver.getJvmCheckedException(method).toList()}")
}
```

Result:

```
method:  [<ERROR TYPE>]
```

neetopia:
fixes #1460


neetopia:
PsiTypeParameter resolution requires additional information hence can't be resolved directly with PSI apis, delegating to existing infrastructure.

ting-yuan (approved):

PR #1684, opened 2024-01-18
restore annotations on type alias

### LINKED ISSUE #1679: Annotations missing on `KSTypeArgument` of `typealias`
We ran into a case in Dagger where the `@JvmSuppressWildcards` is missing in the following case:

```kotlin
interface MyType
typealias MyAlias = MyType
typealias MyListAlias = List<@JvmSuppressWildcards MyAlias>
fun method(): MyListAlias
```

In particular, if we try to resolve `MyListAlias` the `@JvmSuppressWildcards` annotation is now missing (I'm just printing the `toString()` here since it's easier, but actually getting the annotations from the type arguments gives the same result):

```kotlin
// Prints: List<MyType> instead of List<[@JvmSuppressWildcards] MyType>
println("${(method.returnType().declaration as KSTypeAlias).type.resolve()}") 
```

I've also checked that all annotations get dropped, not just `@JvmSuppressWildcards`.

I'm guessing this is related to the nested `typealias`, because the following cases with a single `typealias` seem to work fine:

```kotlin
interface MyType
typealias MyListAlias = List<@JvmSuppressWildcards MyType>
fun method(): MyListAlias

// This works!
// Prints: List<[@JvmSuppressWildcards] MyType>
println("${(method.returnType().declaration as KSTypeAlias).type.resolve()}") 
```

```kotlin
interface MyType
typealias MyAlias = MyType
fun method(): List<@JvmSuppressWildcards MyAlias>

// This works!
// Prints: List<[@JvmSuppressWildcards] MyType>
println("${(method.returnType().declaration as KSTypeAlias).type.resolve()}") 
```

neetopia:
fixes #1679

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
@@ -35,7 +35,6 @@ import org.jetbrains.kotlin.builtins.isKFunctionType
 import org.jetbrains.kotlin.builtins.isKSuspendFunctionType
 import org.jetbrains.kotlin.builtins.isSuspendFunctionType
 import org.jetbrains.kotlin.descriptors.NotFoundClasses
-import org.jetbrains.kotlin.descriptors.SourceElement
 import org.jetbrains.kotlin.types.*
 import org.jetbrains.kotlin.types.typeUtil.TypeNullability
 import org.jetbrains.kotlin.types.typeUtil.isSubtypeOf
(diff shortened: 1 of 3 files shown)


ting-yuan (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/kotlin/KSTypeImpl.kt
```diff
@@ -145,8 +144,7 @@ fun getKSTypeCached(
         KSTypeImpl.getCached(
             kotlinType,
             ksTypeArguments,
-            annotations + kotlinType.annotations
-                .filter { it.source == SourceElement.NO_SOURCE }
+            kotlinType.annotations
```
dstepanov, line 147: Removing the filter might have caused https://github.com/google/ksp/issues/1696

dstepanov: Nevermind found the problem

Issue #1665, opened 2023-12-20
NoClassDefFound in ksp.useKSP2=true mode


mazhukinevgeniy:
Initial report: https://youtrack.jetbrains.com/issue/KT-64365/NoClassDefFound-on-upgrading-ksp-to-2.0.0-beta1

I think the issue is caused by adding kotlin-compiler-common-for-ide [without transitive dependencies](https://github.com/google/ksp/blob/88283988491b32a5952f480f031cbeb85893b3c3/kotlin-analysis-api/build.gradle.kts#L59). `io.javaslang:javaslang:2.0.6` should be provided.

ting-yuan:
Thanks for the bug report and suggested fix! Missing dependencies are added in #1749, which is available in the [nightly build](https://oss.sonatype.org/content/repositories/snapshots/com/google/devtools/ksp/symbol-processing-aa-embeddable/2.0.0-Beta3-1.0.18-SNAPSHOT/), and will be included in the future 1.0.18 release.

PR #1732, opened 2024-02-14
fix: `NoClassDefFound` with `ksp.useKSP2=true`

sgammon:
Fixes and closes google/ksp#1665


neetopia:
I've tried this before but was getting corrupted zip file error at runtime, have you tested this fix with some reproducible cases?

PR #1736, opened 2024-02-15
Adjust bundled deps in KSP2

ting-yuan:
1. Do not `bundle kotlin-stdlib`
2. Do not `bundle symbol-processing-api`
3. Make sure all bundled deps are properly renamed

### DIFF
--- a/symbol-processing-aa-embeddable/build.gradle.kts
+++ b/symbol-processing-aa-embeddable/build.gradle.kts
@@ -29,24 +29,23 @@ tasks.withType<Jar> {
 }
 
 val prefixesToRelocate = listOf(
-    "org.jetbrains.",
-    "org.intellij.",
-    "com.intellij.",
-    "it.unimi.dsi.",
+    "com.fasterxml.",
+    "org.codehaus.",
+    "com.github.benmanes.caffeine.",
     "com.google.common.",
     "com.google.errorprone.",
     "com.google.gwt.",
     "com.google.j2objc.",
-    "kotlin.sequences.",
-    "kotlin.text.",
+    "com.intellij.",
+    "it.unimi.dsi.",
+    "kotlinx.collections.immutable.",
     "org.checkerframework.",
-    "com.github.benmanes.caffeine.",
+    "org.intellij.",
+    "org.jetbrains.",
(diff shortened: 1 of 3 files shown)


symbol-processing-aa-embeddable/build.gradle.kts
```diff
@@ -51,6 +70,44 @@ fun String.replaceWithKsp() =
 
 val DEP_SOURCES_DIR = "$buildDir/source-jar"
 
+val validPaths = prefixesToRelocate.map {
+    it.second.split('.').filter { it.isNotEmpty() }.joinToString("/")
```
neetopia, line 74: when will it be empty?

symbol-processing-aa-embeddable/build.gradle.kts
```diff
@@ -51,6 +70,44 @@ fun String.replaceWithKsp() =
...
+    "org/apache/log4j/package-info.class",
+    "ksp/FirNativeForwardDeclarationGetClassCallChecker.class",
+)
+
+class Trie(paths: List<String>) {
+    class TrieNode(val key: String)
+
+    private val terminals = mutableSetOf<TrieNode>()
+
+    private val m = mutableMapOf<Pair<TrieNode?, String>, TrieNode>().apply {
```
neetopia, line 87: I did not get this, if `m` is the child list of the current trie node, what is this pair used for? Feels like simply putting `String` here should be enough?

symbol-processing-aa-embeddable/build.gradle.kts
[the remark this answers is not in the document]
ting-yuan: zip's specification says that it must be a forward slash.

ting-yuan: When it's ending with a dot.

symbol-processing-aa-embeddable/build.gradle.kts
ting-yuan: The first element of the pair is parent. `m` is shared among all nodes and it maps (parent, key) to the child of the parent.

neetopia (approved):

PR #1737, opened 2024-02-16
include type arguments in reference elements's referencedName/toString

### LINKED ISSUE #1200: Rewrite KSTypeReferenceImpl.toString() with its inner arguments.
For type references like ```Map<Int, Int>```, its ```toString()``` is ```Map``` without inner arguments, which makes it logged a little unclearly.

evant:
Fixes #1200

This just feels like an oversight to me as the other impls do include them, but lmk!


jaschdoc:
This seems to be fixed on main

Issue #1759, opened 2024-02-28
[KSP2] IllegalStateException when getting the modifiers from properties in a Java annotation declaration


kuanyingchou:
Here's a repro: https://github.com/kuanyingchou/ksp/commit/b3da767b51f8c36474defe2638401bb90f2b5f17

Stacktrace:

```
Unexpected symbol type null
java.lang.IllegalStateException: Unexpected symbol type null
	at com.google.devtools.ksp.impl.symbol.kotlin.AbstractKSDeclarationImpl$modifiers$2.invoke(AbstractKSDeclarationImpl.kt:71)
	at com.google.devtools.ksp.impl.symbol.kotlin.AbstractKSDeclarationImpl$modifiers$2.invoke(AbstractKSDeclarationImpl.kt:58)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at com.google.devtools.ksp.impl.symbol.kotlin.AbstractKSDeclarationImpl.getModifiers(AbstractKSDeclarationImpl.kt:58)
	at com.google.devtools.ksp.processor.GetByNameProcessor$process$1$1$1$list$1.invoke(GetByNameProcessor.kt:22)
	at com.google.devtools.ksp.processor.GetByNameProcessor$process$1$1$1$list$1.invoke(GetByNameProcessor.kt:22)
	at kotlin.text.StringsKt__AppendableKt.appendElement(Appendable.kt:84)
	at kotlin.sequences.SequencesKt___SequencesKt.joinTo(_Sequences.kt:2891)
	at kotlin.sequences.SequencesKt___SequencesKt.joinToString(_Sequences.kt:2910)
	at kotlin.sequences.SequencesKt___SequencesKt.joinToString$default(_Sequences.kt:2909)
	at com.google.devtools.ksp.processor.GetByNameProcessor.process(GetByNameProcessor.kt:22)
	at com.google.devtools.ksp.impl.KotlinSymbolProcessing.execute(KotlinSymbolProcessing.kt:528)
	at com.google.devtools.ksp.test.AbstractKSPAATest.runTest(AbstractKSPAATest.kt:141)
	at com.google.devtools.ksp.test.AbstractKSPTest.runTest(AbstractKSPTest.kt:227)
	at com.google.devtools.ksp.test.KSPAATest.testGetByName(KSPAATest.kt:272)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #1769, opened 2024-02-29
fix modifier calculation for synthetic members

neetopia:
test result is only modified for KSP2 for the reason that KSP2 treats Java annotation values as properties while KSP1 treats it as functions, therefore the test processor does not look into Java annotation values in KSP1.

One change to shared logic: `isOpen()` should always check for `FINAL` modifier first, this hasn't been an issue for KSP1 since certain synthetic members of data class (namely `componentX`) is not generated in KSP1, but ran into issue with KSP2 where the `OVERRIDE` modifier present in `componentX` take over the `isOpen` check.


neetopia:
fixes #1759

ting-yuan (approved):

PR #1783, opened 2024-03-07
Customize KotlinCompilerVersion in symbol-processing-aa-embeddable

ting-yuan:
Also renamed META-INF/compiler.version to META-INF/ksp.compiler.version

Tested locally.


symbol-processing-aa-embeddable/build.gradle.kts
```diff
@@ -203,3 +203,32 @@ signing {
     useInMemoryPgpKeys(signingKey, signingPassword)
     sign(extensions.getByType<PublishingExtension>().publications)
 }
+
+abstract class WriteVersionSrcTask @Inject constructor(
+    @get:Input val kspVersion: String,
```
neetopia, line 208: seems that kspVersion is not used other than being annotated with `Input` to potentially trigger rebuild?

neetopia (approved):

symbol-processing-aa-embeddable/build.gradle.kts
ting-yuan: updated!

neetopia (approved):
LGTM

Issue #1791, opened 2024-03-13
Mutual dependencies with other code generating Gradle plugins


ting-yuan:
KSP gets inputs from `KotlinCompilation.allKotlinSourceSetsObservable` and registers its outputs to `KotlinCompilation.defaultSourceSet.kotlin`. This creates a circular dependency. So it excludes generated sources from inputs like this: `filter { !generatedSources.isParentOf(it) }`. It works until there are multiple plugins doing so and they are all broken by mutual dependencies.

While waiting for a solution from [KGP](https://youtrack.jetbrains.com/issue/KT-66407), we need a quick workaround.

ting-yuan:
A mid-term solution from @bitspittle in #1793 is merged. It requires other parties, which add generated code to the default Kotlin source set, to explicitly exclude their generated code from KSP via `ksp { excludedSources(...) }`.

PR #1792, opened 2024-03-13
Add `excludedSources` to the KSP extension

bitspittle:
This new property allows users to create a Gradle task that generates Kotlin code, which can
then get excluded by KSP's processing step.

This can help avoid circular Gradle dependency issues, such as in the case where KSP
generates an intermediate file (like a resource) which a Gradle plugin then consumes
and uses to generate its own source. Without this change, KSP would want to rerun
when the new source got generated, which upsets Gradle because it already ran.

The following is one example of using the `excludedSources` property:

```
val generateCodeTask = tasks.register("generateCodeTask") {
    val outputDir = project.layout.buildDirectory.dir("generated/src")
    outputs.dir(outputDir)

    doLast {
        /* ... generate code into outputDir ... */
    }
}

ksp.excludedSources.from(generateCodeTask)

kotlin {
    sourceSets.main {
        kotlin.srcDir(generateCodeTask)
    }
}
```

Bug #1791


bitspittle:
I'm going to reupload this PR with a new commit message that includes a co-author (I definitely did not do the work of this PR alone). Also, doing this should fix the CLA check error.

PR #1796, opened 2024-03-14
Implement Resolver.getSubpackagesOf()

### LINKED ISSUE #1795: Feature request: Resolver.getSubpackages(String)
We have a use case for scanning subpackages of a given package and referencing symbols found in those packages in an aggregating processor. Currently we can get a specific package via `Resolver.getDeclarationsFromPackage()`, but we can't resolve subpackages from that. If we add a `Resolver.getSubpackages(String): Sequence<String>`, we could load those and chain them into the former API.

ZacSweers:
Resolves #1795

Tests coming, just getting this up for early feedback

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
@@ -239,6 +239,15 @@ interface Resolver {
     fun getDeclarationsFromPackage(packageName: String): Sequence<KSDeclaration>
 
     /**
+     * Returns a sequence of subpackage names of the given [packageName].
+     * @param packageName the root package name to search from.
+     * @return A sequence of subpackage names, if any. Does not include the input package name.
+     * This will return declarations from both dependencies and source.
+     */
+    @KspExperimental
+    fun getSubpackagesOf(packageName: String): Sequence<String>
+
+    /**
      * Returns the corresponding Kotlin class with the given Java class.
      *
      * E.g.
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -95,6 +95,7 @@ import org.jetbrains.kotlin.resolve.constants.ConstantValue
 import org.jetbrains.kotlin.resolve.constants.KClassValue
 import org.jetbrains.kotlin.resolve.constants.evaluate.ConstantExpressionEvaluator
 import org.jetbrains.kotlin.resolve.descriptorUtil.classId
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
 import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperClassifiers
 import org.jetbrains.kotlin.resolve.descriptorUtil.module
 import org.jetbrains.kotlin.resolve.descriptorUtil.propertyIfAccessor
@@ -1011,6 +1012,22 @@ class ResolverImpl(
             .asSequence()
             .mapNotNull { (it as? MemberDescriptor)?.toKSDeclaration() }
             .plus(javaPackageToClassMap.getOrDefault(packageName, emptyList()).asSequence())
+    }
+
+    @KspExperimental
+    override fun getSubpackagesOf(packageName: String): Sequence<String> {
+        fun PackageViewDescriptor.subPackages(): Sequence<PackageViewDescriptor> = memberScope
+            .getContributedDescriptors(DescriptorKindFilter.PACKAGES)
+            .asSequence()
+            .filterIsInstance<PackageViewDescriptor>()
+
+        return generateSequence(listOf(module.getPackage(FqName(packageName)))) { subPackages ->
+            subPackages
+                .flatMap { it.subPackages() }
+                .ifEmpty { null }
+        }
+            .flatMap { it.asSequence() }
+            .mapNotNull { it.fqNameOrNull()?.asString() }
     }
 
     override fun getTypeArgument(typeRef: KSTypeReference, variance: Variance): KSTypeArgument {
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -1020,6 +1020,7 @@ class ResolverImpl(
             .getContributedDescriptors(DescriptorKindFilter.PACKAGES)
             .asSequence()
             .filterIsInstance<PackageViewDescriptor>()
+            .filterNot { it == this@subPackages }
 
         return generateSequence(listOf(module.getPackage(FqName(packageName)))) { subPackages ->
             subPackages
@@ -1027,6 +1028,8 @@ class ResolverImpl(
                 .ifEmpty { null }
         }
             .flatMap { it.asSequence() }
+            // Drop the input package
+            .drop(1)
             .mapNotNull { it.fqNameOrNull()?.asString() }
     }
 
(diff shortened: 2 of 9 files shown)


api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
```diff
@@ -238,6 +238,15 @@ interface Resolver {
     @KspExperimental
     fun getDeclarationsFromPackage(packageName: String): Sequence<KSDeclaration>
 
+    /**
+     * Returns a sequence of subpackage names of the given [packageName].
```
neetopia, line 242: The implementation is only looking at immediate subpackage, rather than recursively searching deeper sub packages, it is worth pointing out in the API documentation here.

ZacSweers: oh I intended for it to recurse, would rather it not?

ZacSweers:
Added tests. Interestingly, I get different results in the tests but I also don't fully understand the test package structure in these tests, and curious for feedback on if I'm doing something wrong in setup here. I'm not following how there's a `main.test.main.test` in the test sources 🤔

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
neetopia: Ah I guess I must have misunderstood the code, never mind.

neetopia:
> Added tests. Interestingly, I get different results in the tests but I also don't fully understand the test package structure in these tests, and curious for feedback on if I'm doing something wrong in setup here. I'm not following how there's a `main.test.main.test` in the test sources 🤔

It is likely limitation from the PSI handling of incorrectly placed files (file path & package mismatch), KSP2's behavior still looks more correct to me. We probably won't have a good solution for this behavior since it is inherited from PSI.

ZacSweers:
Gotcha, so do you think it's just an artifact of the test infra in this case?

ZacSweers:
`:test-utils:test` passes for me locally 🤔

neetopia:
it is passing for me on mac as well. failed only on linux, it seems that the problematic part is from the files where file path mismatches package names, it is a known issue for PSI as we discussed in this PR, at this moment there is probably no good solution for this behavior, maybe we can consider remove the misplaced files ?

ZacSweers:
> maybe we can consider remove the misplaced files

Can you expand on what this would entail?

neetopia:
```
expected: <subpackages of lib1
    subpackages of lib2
    subpackages of main
    main.test
    main.test.nested
    main.test.main
    main.test.main.test
    main.test.main.test.nested
    subpackages of non.exist> but was: <subpackages of lib1
    subpackages of lib2
    subpackages of main
    main.test
    main.test.nested
    subpackages of non.exist>

```

it looks that the diffs are     
```
    main.test.main
    main.test.main.test
    main.test.main.test.nested
```
which seems to be a result from the files being put in wrong directory with respect to the package names
```
// FILE: main/test/main/test/nested/M.java
package main.test.nested;
```
and if this behavior is inherited from PSI to cause the inconsistency, and there is no fix on KSP side, then we might have to fix the test case to avoid it, after all it is the getSubPackage functionality that we want to test in this case.

ZacSweers:
@neetopia @ting-yuan were either of you able to look further into the PSI discrepancies here?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1013,6 +1014,25 @@ class ResolverImpl(
             .plus(javaPackageToClassMap.getOrDefault(packageName, emptyList()).asSequence())
     }
 
+    @KspExperimental
+    override fun getSubpackagesOf(packageName: String): Sequence<String> {
+        fun PackageViewDescriptor.subPackages(): Sequence<PackageViewDescriptor> = memberScope
+            .getContributedDescriptors(DescriptorKindFilter.PACKAGES)
+            .asSequence()
+            .filterIsInstance<PackageViewDescriptor>()
+            .filterNot { it == this@subPackages }
```
ting-yuan, line 1023: May I know why `.filterIsInstance<PackageViewDescriptor>()` and `.filterNot { it == this@subPackages }` are needed?

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
ting-yuan: Maybe commenting that it is recursive then?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -346,6 +347,26 @@ class ResolverAAImpl(
         }
     }
 
+    @KspExperimental
+    override fun getSubpackagesOf(packageName: String): Sequence<String> {
+        return analyze {
+            fun KtPackageSymbol.subPackages(): Sequence<KtPackageSymbol> = getPackageScope()
+                .getPackageSymbols()
+                .distinct()
```
ting-yuan, line 355: Needing to `distinct()` may indicate some issues in the underlying Analysis API. Have you tried removing `distinct()`? If it works then it works. If not, feel free to leave it there and we'll investigate.

test-utils/src/test/kotlin/com/google/devtools/ksp/test/KSPAATest.kt
```diff
@@ -265,6 +265,12 @@ class KSPAATest : AbstractKSPAATest() {
         runTest("../kotlin-analysis-api/testData/getPackage.kt")
     }
 
+    @TestMetadata("getSubPackages.kt")
+    @Test
+    fun testGetSubPackages() {
+        runTest("../kotlin-analysis-api/testData/getSubPackages.kt")
```
ting-yuan, line 271: Please replace with: `runTest("../test-utils/testData/api/getSubPackages.kt")`. I.e., don't use the forked version.

kotlin-analysis-api/testData/getSubPackages.kt
```diff
@@ -0,0 +1,105 @@
+/*
```
ting-yuan, line 1: Tests under `kotlin-analysis-api` are forked version of those in `test-utils`. They are (ideally) only used when there is a behavior difference between KSP1 and KSP2. It doesn't seem to be necessary in this case.

test-utils/testData/api/getSubPackages.kt
```diff
@@ -0,0 +1,104 @@
...
+
+// TEST PROCESSOR: GetSubPackagesProcessor
+// EXPECTED:
+// subpackages of lib1
+// subpackages of lib2
+// subpackages of main
+// main.test
+// main.test.main
+// main.test.main.test
+// main.test.main.test.nested
```
ting-yuan, line 26: Please replace `main.test.main[\..]*` with `main.test.nested`

test-utils/testData/api/getSubPackages.kt
```diff
@@ -0,0 +1,104 @@
...
+package main;
+
+public class K {
+
+}
+
+class KK {}
+
+
+// FILE: main/test/main/test/L.java
```
ting-yuan, line 91: Replace with `// FILE: main/test/L.java`

test-utils/testData/api/getSubPackages.kt
```diff
@@ -0,0 +1,104 @@
...
+package main.test;
+
+public class L {
+
+}
+
+
+// FILE: main/test/main/test/nested/M.java
```
ting-yuan, line 99: Replace with `// FILE: main/test/nested/M.java`

ting-yuan (commented):
Let's fix the test case. IMHO the behavior on KSP1 and KSP2/Mac are wrong. Package names should always be consistent with the `package` directive.

ZacSweers:
We no longer need this API for our use case and I'm limited on bandwidth, so I'll need to relinquish this feature to whoever wants to implement it in the future :)

PR #1802, opened 2024-03-26
support type alias for getSymbolsWithAnnotation

neetopia:
Changes in test cases mainly due to ordering change, with one exception in the last test case where additional members are being returned (annotated property declaration in constructors now have 2 presence) which actually makes more sense.


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -107,6 +109,39 @@ class ResolverAAImpl(
...
+                }
+            }
+
+            override fun visitClassDeclaration(classDeclaration: KSClassDeclaration, data: Unit) {
+                val qualifiedName = classDeclaration.qualifiedName
+                if (qualifiedName != null) {
+                    nameToKSMap[qualifiedName] = classDeclaration
+                }
+                classDeclaration.declarations.forEach { it.accept(this, data) }
+            }
```
ting-yuan, line 131: This block is unnecessary. Type aliases are top-level.

ting-yuan (approved):

PR #1809, opened 2024-03-30
fix backing field checking for top level callables, 

neetopia:
downcast to FIR internal to get file class names

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
+
 package com.google.devtools.ksp.impl.symbol.kotlin
 
 import com.google.devtools.ksp.closestClassDeclaration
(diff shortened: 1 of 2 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
```diff
@@ -107,13 +113,29 @@ class KSPropertyDeclarationImpl private constructor(internal val ktPropertySymbo
...
+                ktPropertySymbol.modality == Modality.ABSTRACT -> false
+                else -> {
+                    val classId = when (
+                        val containerSource =
+                            (ktPropertySymbol as? KtFirKotlinPropertySymbol)?.firSymbol?.containerSource
+                    ) {
+                        is JvmPackagePartSource -> containerSource.classId
+                        is KotlinJvmBinarySourceElement -> containerSource.binaryClass.classId
+                        else -> null
+                    } ?: return@lazy ktPropertySymbol.hasBackingField
```
ting-yuan, line 129: Should this be fixed in the upstream?

ting-yuan (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
neetopia: unfortunately, the implementation in AA reads from metadata and this information is missing in metadata, an ultimate fix will require fix to metadata in klib which will take a long time.

PR #1834, opened 2024-04-10
Set Kotlin to 2.0.0-RC1

ting-yuan:
on the 1.0.21-release branch.


ting-yuan:
only for KSP1. KSP2 uses ToT.

PR #1848, opened 2024-04-18
#1232: Include type hint into KSErrorType.

### LINKED ISSUE #1232: Expose simple names on a missing types's synthetic declaration
In javac, an `ErrorType` is a `DeclaredType` and one can get an `Element` out of it. Its `simpleName` will be the name as it was defined in the sources.

But in KSP, the `KSDeclaration` of a an error `KSType` is a synthetic declaration (specifically that `KSErrorTypeClassDeclaration` singleton) and its simple name is always `<Error>`.

This would be very helpful for reporting to the user nice traces to the missing type:
```
package test

import bar.MissingType;

class TestClass {
  fun foo(missing: MissingType) = TODO()
}
```
```
Validation trace:
    => element (CLASS): test.TestClass
    => element (METHOD): foo(MissingType)
    => type (EXECUTABLE method): (MissingType)void
    => type (ERROR parameter type): MissingType
```
Note that if the type is qualified in the source, it will be similarly be reported as its 'simple name' in javac:
```
 fun foo(missing: bar.MissingType) = TODO()
```
```
    => type (ERROR parameter type): bar.MissingType
```

Jeffset:
This CL makes `KSErrorType` a class with a string
"hint", that by convention should be a "simpleName" of an unresolved type.

`KSErrorTypeClassDeclaration` is also no longer a singleton and references a corresponding `KSErrorType`.

The implementation consists of extracting the available info on the unresolved type on the best effort basis.

Resolves #1232

## Changes for the clients

- if `isError` is true for a `type`, then the `toString()` for the `type` will "likely" return a meaningful value in a format of `"<ERROR TYPE: SomeSimpleName>"`, of `<ERROR TYPE>` (as before) if no info was available.
- `KSErrorType` and their synthetic declarations will now be separate instances.

-------------------

This is the first iteration; if API changes are needed and/or other issues are uncovered, I'm ready to work on them. If my approach to solving the original issue is presumed to be OK, I may write more tests for error types in addition to the existing ones.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -1248,7 +1248,7 @@ class ResolverImpl(
     }
 
     // Convert type arguments for Java wildcard, recursively.
-    private fun KotlinType.toWildcard(mode: TypeMappingMode): KotlinType {
+    private fun KotlinType.toWildcard(mode: TypeMappingMode): Result<KotlinType> {
         val parameters = constructor.parameters
         val arguments = arguments
 
@@ -1258,10 +1258,11 @@ class ResolverImpl(
                 parameter.variance != org.jetbrains.kotlin.types.Variance.INVARIANT &&
                 argument.projectionKind != org.jetbrains.kotlin.types.Variance.INVARIANT
             ) {
-                // conflicting variances
-                throw IllegalArgumentException(
-                    "Conflicting variance: variance '${parameter.variance.label}' vs projection " +
-                        "'${argument.projectionKind.label}'"
+                return Result.failure(
+                    IllegalArgumentException(
+                        "Conflicting variance: variance '${parameter.variance.label}' vs projection " +
+                            "'${argument.projectionKind.label}'"
+                    )
                 )
             }
 
@@ -1270,10 +1271,10 @@ class ResolverImpl(
             val genericMode = argMode.toGenericArgumentMode(
                 getEffectiveVariance(parameter.variance, argument.projectionKind)
             )
-            TypeProjectionImpl(variance, argument.type.toWildcard(genericMode))
+            TypeProjectionImpl(variance, argument.type.toWildcard(genericMode).getOrElse { return Result.failure(it) })
         }
 
-        return replace(wildcardArguments)
+        return Result.success(replace(wildcardArguments))
     }
 
     private val JVM_SUPPRESS_WILDCARDS_NAME = KSNameImpl.getCached("kotlin.jvm.JvmSuppressWildcards")
@@ -1372,12 +1373,19 @@ class ResolverImpl(
             if (position == RefPosition.SUPER_TYPE &&
                 argument.projectionKind != org.jetbrains.kotlin.types.Variance.INVARIANT
             ) {
-                throw IllegalArgumentException("Type projection isn't allowed in immediate arguments to supertypes")
+                val errorType = KSErrorType(
+                    name = type.toString(),
+                    message = "Type projection isn't allowed in immediate arguments to supertypes"
+                )
+                return KSTypeReferenceSyntheticImpl.getCached(errorType, null)
             }
         }
 
         val wildcardType = kotlinType.toWildcard(typeMappingMode).let {
-            var candidate: KotlinType = it
+            var candidate: KotlinType = it.getOrElse { error ->
+                val errorType = KSErrorType(name = type.toString(), message = error.message)
+                return KSTypeReferenceSyntheticImpl.getCached(errorType, null)
+            }
             for (i in indexes.reversed()) {
                 candidate = candidate.arguments[i].type
             }
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -692,7 +692,8 @@ class ResolverImpl(
                 builtIns.arrayType.replace(typeArgs)
             }
             else -> {
-                getClassDeclarationByName(psiType.canonicalText)?.asStarProjectedType() ?: KSErrorType
+                getClassDeclarationByName(psiType.canonicalText)?.asStarProjectedType()
+                    ?: KSErrorType(psiType.canonicalText)
             }
         }
     }
@@ -1054,7 +1055,7 @@ class ResolverImpl(
             }
         }
         // if substitution fails, fallback to the type from the property
-        return KSErrorType
+        return KSErrorType.fromReferenceBestEffort(property.type)
     }
 
     internal fun asMemberOf(
@@ -1247,7 +1248,7 @@ class ResolverImpl(
     }
 
     // Convert type arguments for Java wildcard, recursively.
-    private fun KotlinType.toWildcard(mode: TypeMappingMode): KotlinType? {
+    private fun KotlinType.toWildcard(mode: TypeMappingMode): KotlinType {
         val parameters = constructor.parameters
         val arguments = arguments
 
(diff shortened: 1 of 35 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1257,16 +1258,18 @@ class ResolverImpl(
                 argument.projectionKind != org.jetbrains.kotlin.types.Variance.INVARIANT
             ) {
                 // conflicting variances
-                // TODO: error message
-                return null
+                throw IllegalArgumentException(
```
Jeffset, line 1261: I took liberty in making this an exception. My argument is that returning an error type in a situation seemingly close to an internal error is a bit weird, maybe it's better crash. If we are not ready to make such decision here, please tell me and I'll revert to returning `null` and later on an error type.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -125,7 +125,7 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
 
     override fun asType(typeArguments: List<KSTypeArgument>): KSType {
         if (typeArguments.isNotEmpty() && typeArguments.size != asStarProjectedType().arguments.size) {
-            return KSErrorType
+            return KSErrorType()
```
Jeffset, line 128: In this (and similar) cases, where an error type is used to signal the incorrect API usage rather than to represent a truly missing/unresolved type, I've decided to not provide any additional hints. If that's not okay or if I've misinterpreted something, let's discuss the matter.

Jeffset:
@neetopia @ting-yuan I wonder if we are ready to discuss the solution, if not, please let me know.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
```diff
@@ -110,12 +110,12 @@ class KSTypeImpl private constructor(internal val type: KtType) : KSType {
...
 
     override fun replace(arguments: List<KSTypeArgument>): KSType {
-        return type.replace(arguments.map { it.toKtTypeProjection() })?.let { getCached(it) } ?: KSErrorType
+        return type.replace(arguments.map { it.toKtTypeProjection() })?.let { getCached(it) } ?: KSErrorType()
     }
 
     override fun starProjection(): KSType {
         return type.replace(List(type.typeArguments().size) { KtStarTypeProjection(type.token) })
-            ?.let { getCached(it) } ?: KSErrorType
+            ?.let { getCached(it) } ?: KSErrorType()
```
neetopia, line 118: should we add something to replace type arguments use case? This is not helping with error types.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSTypeReferenceResolvedImpl.kt
```diff
@@ -66,7 +67,13 @@ class KSTypeReferenceResolvedImpl private constructor(
                 ktType is KtClassErrorType || (ktType.classifierSymbol() == null)
             }
         ) {
-            KSErrorType
+            KSErrorType(
+                when (ktType) {
+                    is KtClassErrorType -> ktType.getNameHint()
+                    is KtTypeErrorType -> null // No info available
```
neetopia, line 73: we can look at the parent for possible information, you can ignore this comment if it is too much work for you.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -691,7 +691,8 @@ class ResolverImpl(
                 builtIns.arrayType.replace(typeArgs)
             }
             else -> {
-                getClassDeclarationByName(psiType.canonicalText)?.asStarProjectedType() ?: KSErrorType
+                getClassDeclarationByName(psiType.canonicalText)?.asStarProjectedType()
+                    ?: KSErrorType(psiType.canonicalText)
```
neetopia, line 695: Any specific reason to use canonical text? I think you are using simple names for implementations in analysis api?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
neetopia: Conflicting variance might come from incorrectly defined classes and types applications which should be error types to make more sense as it is user input error rather than the compiler internal error. I think your error message below is good though, so maybe adding it to the hint since the hint is for easier error understanding anyways?

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -1368,19 +1371,17 @@ class ResolverImpl(
             if (position == RefPosition.SUPER_TYPE &&
                 argument.projectionKind != org.jetbrains.kotlin.types.Variance.INVARIANT
             ) {
-                // Type projection isn't allowed in immediate arguments to supertypes.
-                // TODO: error message
-                return KSTypeReferenceSyntheticImpl.getCached(KSErrorType, null)
+                throw IllegalArgumentException("Type projection isn't allowed in immediate arguments to supertypes")
```
neetopia, line 1374: same here.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSClassDeclarationDescriptorImpl.kt
```diff
@@ -153,7 +153,7 @@ class KSClassDeclarationDescriptorImpl private constructor(val descriptor: Class
     override fun asType(typeArguments: List<KSTypeArgument>): KSType =
         descriptor.defaultType.replaceTypeArguments(typeArguments)?.let {
             getKSTypeCached(it, typeArguments)
-        } ?: KSErrorType
+        } ?: KSErrorType()
```
neetopia, line 156: include the original text for the type?

neetopia (commented):
overall looks good, please take a look at the comments and let me know what you think. Also please resolve the conflicts.

There are a couple of `KSErrorType`s without hint filled in, can you please also address them? I did not comment all of them to avoid spamming comments.

Jeffset:
I resolved the conflicts in the first commit with minimal changes. All the additional fixes and requested review changes are in the second commit for more convenience.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
Jeffset: I've tried to generalize the error type creation in such cases and introduced an `errorTypeOnInconsistentArguments` utility function which is now used in the old and AA impls. This pretty much covers all unfilled error types, except some where there's really nothing meaningful to report.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSTypeReferenceResolvedImpl.kt
Jeffset: I've noticed that when there's some type information (at least an identifier, or even identifier matched with an import/qualified identifier) then we get a `KtClassErrorType` which is good. When the type is simply missing altogether - we get `KtTypeErrorType`. Nevertheless I've used `KtTypeErrorType.tryRenderAsNonErrorType` as a last resort, as some info may still be available though there. The place where we handle AA's error types is now `KtType.render()` function for consistency.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
Jeffset: Yes, the idea is to try including as much _type_ information as available. If additional qualifiers are resolved by the compiler - include them. I guess you're right about the things I did in AA, so I went and tried to use qualified names there as well. See `KtType.render() -> KtClassErrorType.qualifiers`

neetopia (approved):
thanks for the work!

PR #1881, opened 2024-05-10
Return KSClassDeclarationEnumEntryImpl for enum entry annotation values.

### LINKED ISSUE #1861: KSP2: Model enum entries in annotation values
Currently it is modeled as `KSType`. In KSP1, each enum entry has its own subtype, so users know which entry is passed.

In KSP2, all entries are of the same containing enum type and become indistinguishable. We either emulate the old behavior in KSP1, or model it differently.

neetopia:
fixes #1861

### DIFF
--- a/kotlin-analysis-api/testData/annotationValue/java.kt
+++ b/kotlin-analysis-api/testData/annotationValue/java.kt
@@ -26,8 +26,8 @@
 // Array
 // @Foo
 // @Suppress
-// RGB
-// JavaEnum
+// G
+// ONE
 // 31
 // [warning1, warning 2]
 // END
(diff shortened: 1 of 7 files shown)


kotlin-analysis-api/testData/annotationValue/java.kt
```diff
@@ -26,8 +26,8 @@
 // Array
 // @Foo
 // @Suppress
-// RGB
-// JavaEnum
+// G
+// ONE
```
ting-yuan, line 30: Aren't you going to change `toString` to include the parent enum?

ting-yuan (approved):

PR #1885, opened 2024-05-11
Add command line parser and entry points

ting-yuan:
So that it can be invoked like:
```
java -cp <KSP and Kotlin library classpath> \
      com.google.devtools.ksp.cmdline.KSPJvmMain \
      -module-name=MyModule \
      -source-roots /path/to/source/roots \
      -kotlin-output-dir=/path/to/output/Kotlin \
      -java-output-dir /path/to/output/Kotlin \
      ...
      /path/to/processor.jar
```

Help message can be printed with `-h` or `--help`:
```
Available options:

*   -module-name=String
*   -source-roots=List<File>
    -common-source-roots=List<File>
    -libraries=List<File>
    -processor-options=Map<String, String>
...
where:
 * is required
 List is colon separated. E.g., arg1:arg2:arg3
 Map is in the form key1=value1:key2=value2
```


cmdline-parser-gen/src/main/kotlin/CmdlineParserGenerator.kt
```diff
@@ -0,0 +1,132 @@
...
+                    val optionNameLen = optionName.length
+                    when (typeName) {
+                        "String", "Boolean", "File" -> {
+                            os.appendLine(
+                                "        arg == \"$optionName\" -> " +
+                                    "$propName = parse$typeName(getArg(args, i++))"
+                            )
+                            os.appendLine(
+                                "        arg.startsWith(\"$optionName=\") -> " +
+                                    "$propName = parse$typeName(arg.substring(${optionNameLen + 1}))"
```
neetopia, line 68: substringAfter?

neetopia (approved):

PR #1901, opened 2024-05-20
handle internal names and JvmName annotations for name mangling.

neetopia:
putting it in draft mode since this PR only fixes source symbols. Theoretically it should handle library symbols if psi stub creation for library symbols are in place (as per the document of the psi elements in analysis API symbols). The fix of this PR can be seen by enabling `mangledNames` test, JvmName annotations and internal names should be handled now.

Note that light class still won't handle inline names.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
@@ -63,10 +63,12 @@ import org.jetbrains.kotlin.analysis.api.types.KtType
 import org.jetbrains.kotlin.analysis.decompiler.stub.file.ClsKotlinBinaryClassCache
 import org.jetbrains.kotlin.analysis.low.level.api.fir.api.getFirResolveSession
 import org.jetbrains.kotlin.analysis.project.structure.KtSourceModule
+import org.jetbrains.kotlin.asJava.toLightClass
 import org.jetbrains.kotlin.builtins.jvm.JavaToKotlinClassMap
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCliJavaFileManagerImpl
 import org.jetbrains.kotlin.fir.types.isRaw
 import org.jetbrains.kotlin.fir.types.typeContext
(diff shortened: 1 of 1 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -480,8 +483,18 @@ class ResolverAAImpl(
...
 
-    // TODO: handle @JvmName annotations, mangled names
+    // TODO: handle library symbols
     override fun getJvmName(accessor: KSPropertyAccessor): String? {
+        (
+            (accessor.receiver.closestClassDeclaration() as? KSClassDeclarationImpl)
+                ?.ktClassOrObjectSymbol?.psi as? KtClassOrObject
+            )?.toLightClass()?.allMethods?.filterIsInstance<SymbolLightSimpleMethod>()
+            ?.singleOrNull {
+                (it.parameters.isNotEmpty() xor (accessor is KSPropertyGetter)) &&
```
neetopia, line 493: This line is very tricky, the idea is we don't have a way to tell if a light method is getter or setter, and both getter and setter shares a same property declaration as the `kotlinOrigin`, so we just check if an accessor has any parameters to determine whether it is a getter or setter.

ting-yuan:
> Theoretically it should handle library symbols if psi stub creation for library symbols are in place (as per the document of the psi elements in analysis API symbols)

Will this ever be available in standalone mode? Does it require indexing of binary dependencies?

neetopia:
> > Theoretically it should handle library symbols if psi stub creation for library symbols are in place (as per the document of the psi elements in analysis API symbols)
> 
> Will this ever be available in standalone mode? Does it require indexing of binary dependencies?



> > Theoretically it should handle library symbols if psi stub creation for library symbols are in place (as per the document of the psi elements in analysis API symbols)
> 
> Will this ever be available in standalone mode? Does it require indexing of binary dependencies?

That's something we need to investigate further. Currently there are already library symbols with a stub in KSP2, see https://github.com/google/ksp/pull/1873/

ting-yuan (approved):
Feel free to merge this first and address the remaining cases later.

PR #1925, opened 2024-05-24
Replace DirectoriesScope with IncrementalGlobalSearchScope

ting-yuan:


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/standalone/KspSourceModuleBuilder.kt
```diff
@@ -97,6 +95,22 @@ class KspModuleBuilder(
...
+class IncrementalGlobalSearchScope(
+    project: Project,
+    initialSet: Collection<VirtualFile> = emptyList(),
+) : GlobalSearchScope(project) {
+    // TODO: optimize space with trie.
+    val files = mutableSetOf<VirtualFile>().apply { addAll(initialSet) }
+
+    fun addAll(files: Collection<VirtualFile>) = this.files.addAll(files)
+
+    override fun contains(file: VirtualFile): Boolean = file in files
```
neetopia, line 107: does this work? the `virtualFiles` we have in this class during the initialization phase are directories, do you need to manually walk the directories and add all files into the `files` property of this class?

ting-yuan: `collectVirtualFilesByRoots` in the above walks the directories and returns only files.

neetopia (approved):

Issue #1911, opened 2024-05-22
`NoSuchMethodError` when running KSP2 with Dagger (KSP 2.0.0-1.0.21)


lwasyl:
Dagger fails with `java.lang.NoSuchMethodError: 'com.google.common.collect.ImmutableBiMap com.google.common.collect.ImmutableBiMap$Builder.buildOrThrow()'`. The method [was added in Guava 31](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/ImmutableBiMap.Builder.html#buildOrThrow())

I see in project dependencies that Dagger correctly depends on Guava>31 and nothing seems to be overriding that version:
```java
kspKotlinProcessorClasspath
+--- project :dependencies-platform
|    +--- org.junit:junit-bom:5.10.2
|    +--- org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.8.1
|    +--- com.google.firebase:firebase-bom:33.0.0
|    +--- com.squareup.okhttp3:okhttp-bom:4.12.0
|    +--- androidx.compose:compose-bom:2024.05.00
|    +--- org.jetbrains.kotlin:kotlin-bom:2.0.0
|    |    +--- org.jetbrains.kotlin:kotlin-stdlib:2.0.0 (c)
|    |    +--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:2.0.0 (c)
|    |    +--- org.jetbrains.kotlin:kotlin-reflect:2.0.0 (c)
|    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk7:2.0.0 (c)
|    +--- com.google.dagger:dagger-compiler:2.51.1 (c)
|    +--- com.google.dagger:dagger:2.51.1 (c)
|    \--- javax.inject:javax.inject:1 (c)
\--- com.google.dagger:dagger-compiler -> 2.51.1
     +--- com.google.dagger:dagger:2.51.1
     |    \--- javax.inject:javax.inject:1
     +--- com.google.dagger:dagger-spi:2.51.1
     |    +--- com.google.dagger:dagger:2.51.1 (*)
     |    +--- com.google.code.findbugs:jsr305:3.0.2
     |    +--- com.google.devtools.ksp:symbol-processing-api:1.9.20-1.0.14
     |    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.0 -> 2.0.0
[...]

PR #1931, opened 2024-05-28
Bump Guava version to 33.2.0.

ting-yuan:


ting-yuan:
Temporarily fixes #1911, where KSP's dependency of older Guava polluted the classpath. There are a few options:
1. Do not bundle Guava. However, because of the isolating classloader, KSP still needs to specify a version of Guava manually and cannot let Gradle decide it. There is no different from bundling Guava from processor's point of view.
2. Implement a special classloader that does some filtering. This is a bit complicated but may be the solution in the long run.
3. (this patch) Bump Guava version and assume that it is backward compatible.

neetopia (approved):

PR #1934, opened 2024-05-30
optimize annotation default value evaluation.

### LINKED ISSUE #1909: KSP2: default annotation arguments no longer work in 1.0.21
Example using auto service

```java
public @interface AutoService {
  Class<?>[] value();
}
```

The below sample

```kotlin
@AutoService // <-- no args, so default [] is used
class CustomCallable : Callable<String> {
    override fun call(): String = "Hello world!"
}
```

Results in a `KSAnnotation` that reports that both `annotation.arguments` and `annotation.defaultArguments` are empty.

Repro is the failing test in https://github.com/ZacSweers/auto-service-ksp/pull/89

image

image

neetopia:
fixes #1909


neetopia:
context for picking `null` as default value for non array values: https://github.com/google/ksp/issues/1909#issuecomment-2138675171

ting-yuan (approved):

PR #1973, opened 2024-06-18
recursively substitute functions for compute asMemberOf

### LINKED ISSUE #1919: [KSP2] KSFunctionDeclaration.asMemberOf does not return correct types
This is a case in XProcessing. Not sure if it's an intended change. With classes like:

```
interface Usage : Foo<Long, Integer> {
    fun foo(param: Foo<Double, Integer>): Foo<String, Integer>
}
interface Foo<V1, V2: Integer> : Bar<Baz<V1, Number>, V2> {}
interface Bar<U1, U2: Integer> : Baz<U1, U2> {}
interface Baz<T1, T2: Number> {
    fun method1(): T1
    fun method2(): T2
}
```

And getting the return type of `method1.asMemberOf(UsageType)` we get `U1` in KSP2 but `Baz<Long, Number>` in KSP1.

Repro: https://github.com/kuanyingchou/ksp/commit/e3414b9a34be49a84cfce9d775796c43dd2c71f9

neetopia:
fixes #1919

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
@@ -831,8 +831,15 @@ class ResolverAAImpl(
                         // recursively substitute to ensure transitive substitution works.
                         // should fix in upstream as well.
                         var result = resolved.type
+                        var cnt = 0
                         while (it.substitute(result) != result) {
+                            if (cnt > 100) {
+                                throw IllegalStateException(
+                                    "Potential infinite loop in type substitution for computeAsMemberOf"
+                                )
+                            }
                             result = it.substitute(result)
+                            cnt += 1
                         }
                         KSTypeImpl.getCached(result)
                     }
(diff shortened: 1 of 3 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -865,10 +865,25 @@ class ResolverAAImpl(
...
                     buildSubstitutor {
                         fillInDeepSubstitutor(containing.type, this@buildSubstitutor)
-                    }.let { (function as KSFunctionDeclarationImpl).ktFunctionSymbol.substitute(it) }.let {
+                    }.let {
+                        // recursively substitute to ensure transitive substitution works.
+                        // should fix in upstream as well.
+                        // for functions we need to test both returnType and value parameters converges.
+                        var funcToSub = (function as KSFunctionDeclarationImpl).ktFunctionSymbol.substitute(it)
+                        var next = funcToSub.substitute(it)
+                        while (
```
ting-yuan, line 876: The complexity of this logic and also its dependency on the substitutor makes me worried whether it always converges. Can you add a counter to throw in case it doesn't?

neetopia: Sounds fair to me, I thought about the converge issue before but can't think about a fair number to exit.

neetopia: tentatively put 100 as a safe value.

PR #1977, opened 2024-06-25
UPDATE_AA_VERSION: 2.0.20-dev-6501

ting-yuan:


neetopia (approved):

neetopia:
Question: what will happen when this gets merged? The 4 middle commits will be picked into release branch, have we verified these commits will work with old versions of compiler/IDE?

ting-yuan:
They are all under kotlin-analysis-api and will not participate functional tests in release branches, by design.

PR #1983, opened 2024-06-28
Fork common-util and test-utils into compiler-plugin and kotlin-analysis-api

ting-yuan:


kotlin-analysis-api/compiler/cli/cli-common/resources/META-INF/extensions/compiler.xml
ting-yuan, line 1: Previously, KSPAATest happened to work because this file sits with KotlinCoreEnvironment in the same kotlin-compiler jar, which overrides kotlin-compiler-common-for-ide and kotlin-compiler-fe10-for-ide. Now that the classpath order has changed, it this file become not found by KotlinCoreEnvironment, and therefore we need to provide it here explicitly.

See https://github.com/JetBrains/kotlin/blob/master/compiler/cli/cli-base/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreEnvironment.kt#L682 for details.

Note that the shadow jars work because it is packed into the same jar of KotlinCoreEnvironment.

neetopia (approved):

PR #1986, opened 2024-07-02
Clear caches before early exit

### LINKED ISSUE #1939: [KSP2] KSPropertyDeclaration.hasBackingField returns inconsistent results
We have 6 tests in XProcessing that hit this. `hasBackingField` returns different results when running the test individually vs running all tests in the test class.

Repro: https://github.com/kuanyingchou/ksp/commit/b7122a91afc4b24230c8c256b4e332c98452a34c

ting-yuan:
Fixes #1939

### DIFF
--- a/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPAATest.kt
+++ b/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPAATest.kt
@@ -146,6 +146,18 @@ class KSPAATest : AbstractKSPAATest() {
         runTest("../test-utils/testData/api/builtInTypes.kt")
     }
 
+    @TestMetadata("A.kt")
+    @Test
+    fun testA() {
+        runTest("../test-utils/testData/api/A.kt")
+    }
+
+    @TestMetadata("B.kt")
+    @Test
+    fun testB() {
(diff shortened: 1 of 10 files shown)


kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPAATest.kt
```diff
@@ -146,6 +146,18 @@ class KSPAATest : AbstractKSPAATest() {
         runTest("../test-utils/testData/api/builtInTypes.kt")
     }
 
+    @TestMetadata("ObjCacheA.kt")
+    @Test
+    fun testObjCacheA() {
```
neetopia, line 151: name suggests objCache, I assume it means the caches in psi manager as suggested by the code above rather than the KSP's own object cache?

neetopia (approved):

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPAATest.kt
ting-yuan: No, it is KSP's object cache. PSI and AA caches need to be cleared too.

neetopia (approved):

PR #2009, opened 2024-07-20
Fix `:kotlin-analysis-api:shadowJar` failing on Windows

lukellmann:
This is similar to https://github.com/google/ksp/pull/1436.


ting-yuan:
Thanks for the patch!

Issue #1963, opened 2024-06-10
[KSP2] NullPointerException when getting name of nested annotation


kuanyingchou:
Got NullPointerException when getting the name of the nested annotation when the origin of the annotation is JAVA_LIB.

Repro: https://github.com/kuanyingchou/ksp/commit/60fa7a08e756cc78f008dd99dad6fead8c82c50b

neetopia:
This should be the same issue as nested annotation value.

PR #2015, opened 2024-07-25
Nested annotation value: handle non-java-builtin types

ting-yuan:
The fix is in `kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt`. Other commits are various tests on nested type annotations, based on @kuanyingchou 's test cases.


ting-yuan:
Fixes #1963

PR #2032, opened 2024-08-15
Dispose root disposable of AA

ting-yuan:


ting-yuan:
Verified with continuous builds of examples/playground. Before the change it OOM after ~50 builds. After the change it pass 500 builds. There is also no noticeable heap growth according to VisualVM.

PR #2034, opened 2024-08-18
Replace usages of findProperty

### LINKED ISSUE #1752: Gradle Plugin: Support isolated projects
- Replace `findProperty` with `providers.gradleProperty`
- Ideally, KGP adds more provider (and DomainCollection) APIs instead Observable/Lazy, ConfigurationAwareProvider...

ansman:
`findProperty` is not compatible with project isolation. The new `providers.gradleProperty` is however.

This partially fixes #1752


ting-yuan (approved):

ZacSweers:
I'd like to request that this be either reverted or expose another way to enable/disable KSP2 per-project or per-task. With this change, it can only be enabled/disabled globally because `ProviderFactory.gradleProperty()` only reads the root project's gradle properties.

ansman:
@ZacSweers Fix in https://github.com/google/ksp/pull/2089

Monday1908 (changes requested):
API29

PR #2035, opened 2024-08-19
Use properties where possible in KspExtension

### LINKED ISSUE #1752: Gradle Plugin: Support isolated projects

ansman:
This PR converts all internal things in KspExtension to be properties instead of lists and maps. This will ensure that reads are deferred for as long as possible and will prevent forcing reads during configuration.

Unfortunately some properties cannot be changed without causing changes to the ABI. New properties could be introduced but the names would have to be different.

This partially fixes #1752

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspExtension.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspExtension.kt
@@ -17,12 +17,12 @@
 
 package com.google.devtools.ksp.gradle
 
-import javax.inject.Inject
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.provider.Provider
 import org.gradle.process.CommandLineArgumentProvider
+import javax.inject.Inject
 
 abstract class KspExtension @Inject constructor(project: Project) {
     internal val apOptions = project.objects.mapProperty(String::class.java, String::class.java)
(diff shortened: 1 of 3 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspExtension.kt
```diff
@@ -18,19 +18,24 @@
...
+import org.gradle.api.provider.Provider
 import org.gradle.process.CommandLineArgumentProvider
+import javax.inject.Inject
 
-abstract class KspExtension {
-    internal val apOptions = mutableMapOf<String, String>()
-    internal val commandLineArgumentProviders = mutableListOf<CommandLineArgumentProvider>()
-    internal val excludedProcessors = mutableSetOf<String>()
+abstract class KspExtension @Inject constructor(project: Project) {
+    internal val apOptions = project.objects.mapProperty(String::class.java, String::class.java)
```
ansman, line 28: Ideally this should have `finalizeValueOnRead` as well, but since it's exposed as `arguments`, it's not safe if consumers are reading the values.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -269,20 +296,18 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
...
-                        emptyList(),
-                    )
-                }
+                getSubpluginOptions(
+                    project = project,
+                    kspExtension = kspExtension,
+                    sourceSetName = sourceSetName,
+                    target = target,
+                    isIncremental = isIncremental,
+                    allWarningsAsErrors = project.provider { kspExtension.allWarningsAsErrors },
```
ansman, line 305: Ideally `allWarningsAsErrors` should be a a property already but for now this is better.

ting-yuan (approved):

ting-yuan:
Thank you so much for the PR!

PR #2068, opened 2024-08-29
Implement support for reading resource files in a symbol processor

### LINKED ISSUE #2008: Accessing resources by name
Right now you cannot access resources (in `src/main/resources` let's say), like you can access classes by name. You can obviously still read the file normally, but you won't be able to track it as a dependency. In my particular case this is not much of an issue since the resource file changes only very rarely, but for other use cases it wouldn't be great to have to manually invalidate the KSP cache anytime a resource changes.

ansman:
This would fix #2008


api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
```diff
@@ -37,6 +38,9 @@ interface Resolver {
      */
     fun getAllFiles(): Sequence<KSFile>
 
+    fun getResource(path: String): InputStream?
+    fun getAllResources(): Sequence<String>
```
ansman, line 42: Obviously these would have to be documented before merging.

I also don't know how they would behave on non JVM targets which would have to be tested.

common-util/src/main/kotlin/com/google/devtools/ksp/common/ResourceUtils.kt
```diff
@@ -0,0 +1,28 @@
+package com.google.devtools.ksp.common
+
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileNotFoundException
+import java.io.InputStream
+
+fun Iterable<File>.getResource(path: String): InputStream? = firstNotNullOfOrNull { root ->
```
ansman, line 8: TODO: Add a test that verifies that you cannot access things in parent directories.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -305,6 +320,12 @@ abstract class KspGradleConfig @Inject constructor() {
     @get:PathSensitive(PathSensitivity.RELATIVE)
     abstract val javaSourceRoots: ConfigurableFileCollection
 
+    @get:InputFiles
+    @get:SkipWhenEmpty
+    @get:IgnoreEmptyDirectories
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    abstract val resourceRoots: ListProperty<Directory>
```
ansman, line 327: I did not make resource handling incremental. Not sure if it's needed and how the API would look. Maybe there would be a separate `getNewResources` or something like that.

api/src/main/kotlin/com/google/devtools/ksp/processing/Resolver.kt
ansman, line 42: TODO: We'd need a way to associate an output file with a specific resource.

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/ResourcesIT.kt
```diff
@@ -0,0 +1,74 @@
...
+    fun testResources() {
+        val gradleRunner = GradleRunner.create().withProjectDir(project.root)
+        gradleRunner.withArguments(":workload:assemble").build().let { result ->
+            result.assertOutputContains("foo.txt: Hello")
+            result.assertOutputContains("foo/bar/baz/quux: World")
+        }
+    }
+
+    @Test
+    fun testUpToDate() {
```
ansman, line 29: FYI this test will fail until Gradle is updated due to issues with how older Gradle handles `SkipWhenEmpty` and `ListProperty<Directory>`

PR #2069, opened 2024-08-29
Cleanup the build scripts and update to Gradle 8.10

ansman:


ting-yuan (approved):
LGTM, thanks!

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -323,13 +323,14 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
...
 
+            @Suppress("DEPRECATION")
             if (kspExtension.allowSourcesFromOtherPlugins) {
                 val kotlinCompileTask = kotlinCompileProvider.get()
                 fun setSource(source: FileCollection) {
                     // kspTask.setSource(source) would create circular dependency.
                     // Therefore we need to manually extract input deps, filter them, and tell kspTask.
-                    kspTask.setSource(project.provider { source.files })
-                    kspTask.dependsOn(project.provider { source.nonSelfDeps(kspTaskName) })
+                    kspTask.setSource(project.provider { source.files.onEach { println(it) } })
```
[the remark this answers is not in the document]
ansman: Oops yes, that's from debugging the issue with the removed test

build.gradle.kts
```diff
@@ -25,7 +27,7 @@ plugins {
 
     // Adding plugins used in multiple places to the classpath for centralized version control
     id("com.github.johnrengelman.shadow") version "7.1.2" apply false
-    id("org.jetbrains.dokka") version "1.8.10" apply false
+    id("org.jetbrains.dokka") version "1.9.20" apply false
```
ansman, line 30: @ting-yuan I had to update to the latest dokka when updating AGP

gradle.properties
```diff
@@ -2,7 +2,7 @@
 org.gradle.jvmargs=-Duser.country=US -Dkotlin.daemon.jvm.options=-Xmx4096m -Dfile.encoding=UTF-8
 
 kotlinBaseVersion=2.1.0-dev-5441
-agpBaseVersion=7.3.1
+agpBaseVersion=8.0.2
```
ansman, line 5: @ting-yuan To use Gradle 8, AGP 8 needs to be used. If KSP still needs to support AGP 7 there is a way, let me know.

build.gradle.kts
```diff
@@ -94,12 +96,25 @@ subprojects {
...
+                languageVersion = compileJavaVersion
+            }
+        }
+    }
+
     tasks.withType<JavaCompile>().configureEach {
-        sourceCompatibility = JavaVersion.VERSION_1_8.toString()
-        targetCompatibility = JavaVersion.VERSION_1_8.toString()
+        sourceCompatibility = JavaVersion.VERSION_11.toString()
+        targetCompatibility = JavaVersion.VERSION_11.toString()
```
ansman, line 114: This might be a big deal, I don't know. I doubt many people are using Java 8 to run Gradle but it's worth noting in the changelog at least

ansman, line 114: @ting-yuan This might be a big deal, I don't know. I doubt many people are using Java 8 to run Gradle but it's worth noting in the changelog at least

hfhbd: Well, AGP 8.6 requires JVM 17, and Gradle 8.10 adds a warning, and Gradle 9 requires 17 too.

ansman: Yeah it's more about how old of a Gradle version that KSP wants to support

.github/workflows/main.yml
```diff
@@ -34,21 +34,10 @@ jobs:
...
       uses: actions/setup-java@v1.4.3
       with:
-        java-version: '9'
-        java-package: jdk
-        architecture: x64
-    - name: set JDK_9 environment variable for kotlin compiler
-      shell: bash
-      env:
-        ACTIONS_ALLOW_UNSECURE_COMMANDS: true
-      run: echo ::set-env name=JDK_9::$(echo $JAVA_HOME)
```
ansman, line 47: I'm not sure what the purpose of this was. I don't see any traces of Java 9 anywhere else

gradle.properties
ting-yuan: I'm a bit conservative on dropping support of older AGP versions. Can we stay with AGP 7 a bit longer?

.github/workflows/main.yml
ting-yuan: The test framework depended on JDK_9. I need to check whether that is still the case.

gradle.properties
ansman: AGP 7 only supports Gradle 7 while AGP 8 only supports AGP 8.

If we want to we can try and change the Gradle runner version for the tests that use Android. But supporting multiple major versions might be tricky.

ansman: So actually this isn't dropping support for older Android since no new APIs are being used. There is a test case for supporting AGP 7.4.1, I will add an additional one for 7.3.1

.github/workflows/main.yml
ansman: I don't see any usage of the `JDK_9` variable on any sources (including the test framework). The tests also pass locally at least.

gradle.properties
scott-pollom: +1, users can still use older AGP versions even if this gets upgraded.

scott-pollom:
Is this blocked by anything? I'm hoping it will be merged soon so I can rebase [this PR] on top of it.

ting-yuan:
Sorry I missed the latest push which added test for AGP compatibility. The PR looks perfect to me! Please rebase and I'll merge as soon as possible.

ansman:
@scott-pollom this is merged now

Issue #2078, opened 2024-09-02
[KSP2] KSAnnotation.arguments contains args with original names when @JvmName is applied


kuanyingchou:
When using a Kotlin annotation from JAVA or JAVA_LIB its arguments contain both the `@JvmName` version and the original version.

Repro: https://github.com/kuanyingchou/ksp/commit/26f11c82df2a024e62ebc9f3151bb561c48d0ff7

PR #2086, opened 2024-09-05
KSAnnotation.arguments: exclude non-explict defaults

ting-yuan:
This makes failing test cases in #2078 pass. However, it doesn't fix it. Tests with explicit default value still fail.

KSP1 also fails if the default value is explicit; It doesn't handle `@JvmName` as well.


kotlin-analysis-api/testData/jvmName.kt
```diff
@@ -2,10 +2,25 @@
 // EXPECTED:
 // (getX, setX), (getY, null)
 // (getX, setX), (getY, null)
+// stringParameter
```
kuanyingchou, line 5: Somehow this is still `stringParameter, stringParam` on my local checkout.

kotlin-analysis-api/testData/jvmName.kt
```diff
@@ -2,10 +2,25 @@
...
+// stringParameter
```
kuanyingchou, line 6: Same as above.

kotlin-analysis-api/testData/annotationValue/annotationValue_java.kt
```diff
@@ -24,8 +24,6 @@
 // MyClassInLib: stringParam = 2
 // MyClassInLib: stringParam2 = 1
 // MyClassInLib: stringArrayParam = [3, 5, 7]
-// []
```
kuanyingchou, line 27: Accidental removal?

kuanyingchou (approved):
Thanks for looking into this!

kotlin-analysis-api/testData/jvmName.kt
ting-yuan: Sorry, I pushed the debug version. Fixed it now.

kotlin-analysis-api/testData/annotationValue/annotationValue_java.kt
ting-yuan: This is intended. The annotation class has two parameters that have no explicit default value declared.

ting-yuan: To be clear,
1. `KSAnnotation.arguments` doesn't include default arg if the corresponding parameter doesn't have a explicit default value.
2. `KSAnnotation.defaultArguments` returns the type default value (0 for numerics, null for others) if no explicit default is specified.

This makes missing arguments detectable in `KSAnnotation.arguments`. `KSAnnotation.defaultArguments` should probably align with that, but I'm worrying that changing it would break a lot of processors.

PR #2089, opened 2024-09-06
Allow enabling KSP2 per module

ansman:
With the change in #2034 a side effect was that KSP2 had to be enabled per Gradle project, not per module.

With this change a DSL property has been added to handle this.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspExtension.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspExtension.kt
@@ -17,14 +17,24 @@
 
 package com.google.devtools.ksp.gradle
 
+import com.google.devtools.ksp.KspExperimental
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import org.gradle.process.CommandLineArgumentProvider
 import javax.inject.Inject
 
 abstract class KspExtension @Inject constructor(project: Project) {
+    /**
+     * Enables or disables KSP 2, defaults to the `ksp.useKsp2` gradle property or `false` if that's not set.
+     *
+     * This API is temporary and will be removed once KSP1 is removed.
+     */
+    @KspExperimental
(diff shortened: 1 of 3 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspExtension.kt
```diff
@@ -37,6 +48,16 @@ abstract class KspExtension @Inject constructor(project: Project) {
 
     open val arguments: Map<String, String> get() = apOptions.get()
 
+    init {
+        @Suppress("LeakingThis")
+        useKsp2.convention(
+            project.providers
```
ZacSweers, line 54: You can inject a providerfactory and Objectfactory so you can initialize the whole property inline

ansman: I typically prefer to use abstract properties. You do have to to suppress the lint error but I like it better anyway

ZacSweers: Then I think it's at least better to move the convention setting to the task registration block? This code also accesses `Task.getProject()`, which is a generally soft-deprecated API even if you're using it in the one safe place here.

ansman: This is an extension, not a task. The convention setting could be moved to the plugin, I have no strong opinion

ansman: Moved the convention to the plugin

ting-yuan (approved):

ting-yuan:
The test case needs to be updated:

```
    * What went wrong:
    Script compilation error:

      Line 7: ksp { useKsp2.set(true) }
                    ^ This API is experimental.It may be changed in the future without notice or might be removed.
```

Also lint failed in another place:
```
> Task :gradle-plugin:ktlint FAILED
/home/runner/work/ksp/ksp/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt:68:1: Unused import

```

PR #2096, opened 2024-09-10
Rollback custom constant value evaluator

### LINKED ISSUE #2077: [KSP2] default value is null for nested annotation
This is similar to https://github.com/google/ksp/issues/1833. Getting the default value returns null when the origin is KOTLIN_LIB or JAVA_LIB.

Repro: https://github.com/kuanyingchou/ksp/commit/28add760542587b4e1df3a857084afd54259e01f

ting-yuan:
Use FIR's implementation instead.

Test case is derived from https://github.com/kuanyingchou/ksp/commit/28add760542587b4e1df3a857084afd54259e01f

Fixes #2077

### DIFF
--- a/kotlin-analysis-api/testData/annotationValue/defaultKClassValue.kt
+++ b/kotlin-analysis-api/testData/annotationValue/defaultKClassValue.kt
@@ -1,8 +1,8 @@
 // TEST PROCESSOR: DefaultKClassValueProcessor
 // EXPECTED:
+// java.lang.String
 // kotlin.String
-// kotlin.String
-// kotlin.String
+// java.lang.String
 // kotlin.Int
 // END
 // MODULE: lib1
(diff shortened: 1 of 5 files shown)


kotlin-analysis-api/testData/annotationValue/defaultKClassValue.kt
```diff
@@ -1,8 +1,8 @@
 // TEST PROCESSOR: DefaultKClassValueProcessor
 // EXPECTED:
+// java.lang.String
 // kotlin.String
-// kotlin.String
-// kotlin.String
+// java.lang.String
```
ting-yuan, line 5: Given `annotation class ExampleAnnotation(val value: kotlin.reflect.KClass<*> = java.lang.String::class)`, I think the  old expected values are wrong.

kuanyingchou (approved):
Thanks!

PR #2101, opened 2024-09-12
Make setter's parameter name `value` by default

### LINKED ISSUE #1865: [KSP2] Property setter parameter name becomes `value`
In KSP2 the parameter name of property setters is `value` when the origin is `KOTLIN_LIB` but is `<set-?>` when the origin is `KOTLIN`, while it was always `<set-?>` in KSP1.

Repro: https://github.com/kuanyingchou/ksp/commit/49600e0fd5933bee4338eaee9b7b7dfe86636fa8

ting-yuan:
per https://kotlinlang.org/docs/properties.html#getters-and-setters

KSP1 is left unchanged.

Fixes #1865


kuanyingchou (approved):
Thanks!

PR #2106, opened 2024-09-17
Use `main` when module name of library is unavailable

### LINKED ISSUE #2105: JVM Module name is changed for libraries
Not sure if it's intended and this happens in both KSP1 and KSP2. The module name of libraries become "$Library for XXX" if it's not specified with "-module-name" with a name other than "main" while compiling. So internal function names in libraries become, say, "f$$Library for main" instead of "f$main".

This may come from here: https://github.com/google/ksp/blob/main/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt#L193

ting-yuan:
which follows kotlinc.

Fixes #2105


kuanyingchou (approved):
Ha, that's fast! Thanks!

ting-yuan:
It was on my todo list :)

PR #2114, opened 2024-09-23
Update utils.kt

GeorgCantor:
**Variable Naming:** Changed `parent` to `current` to better reflect its role in the loop. This makes it clearer that we are traversing up the hierarchy.

**Early Return:** Instead of casting `parent` to `KSFile?` at the end, we check if `current` is an instance of `KSFile` during each iteration. If it is, we return it immediately. This avoids unnecessary casting and enhances readability.

**Null Return:** The method now directly returns `null` when no containing file is found, which is more idiomatic in Kotlin.


jaschdoc:
Hi, thank you for the PR. I'm closing this PR for now since this is only a stylistic change and so there does not seem to be a benefit to changing it. Additionally, the latest version does not perform a cast in the return statement since the compiler smart casts it.

PR #2122, opened 2024-09-27
Support AGP's built-in Kotlin compilation

scott-pollom:
Previously, the sourceSet-specific KSP configurations (e.g., kspTest) weren't being created when AGP's built-in Kotlin compilation was enabled.

This change also increases the AGP version to 8.7.0-beta01 in order to test against a version of AGP with the built-in Kotlin support.

Bug: b/362279380
Test: GradleCompilationTest

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
@@ -35,11 +35,13 @@ class KspConfigurations(private val project: Project) {
         }.replaceFirstChar { it.lowercase() }
     }
 
+    /**
+     * Returns a new or existing [Configuration] with the given [name], with applied properties.
+     */
     private fun createConfiguration(
         name: String,
         readableSetName: String,
     ): Configuration {
-        // maybeCreate to be future-proof, but we should never have a duplicate with current logic
         return project.configurations.maybeCreate(name).apply {
             description = "KSP dependencies for the '$readableSetName' source set."
             isCanBeResolved = false // we'll resolve the processor classpath config
@@ -48,14 +50,19 @@ class KspConfigurations(private val project: Project) {
         }
     }
 
-    private fun getAndroidConfigurationName(target: KotlinTarget, sourceSet: String): String {
+    /**
+     * Returns the Android sourceSet-specific KSP configuration name given a [kotlinTarget] and [sourceSet].
+     *
+     * For single-platform, [kotlinTarget] can be null.
+     */
+    private fun getAndroidConfigurationName(kotlinTarget: KotlinTarget?, sourceSet: String): String {
         val isMain = sourceSet.endsWith("main", ignoreCase = true)
         val nameWithoutMain = when {
             isMain -> sourceSet.substring(0, sourceSet.length - 4)
             else -> sourceSet
         }
         // Note: on single-platform, target name is conveniently set to "".
-        return configurationNameOf(PREFIX, target.name, nameWithoutMain)
+        return configurationNameOf(PREFIX, kotlinTarget?.name ?: "", nameWithoutMain)
     }
 
     private fun getKotlinConfigurationName(compilation: KotlinCompilation<*>, sourceSet: KotlinSourceSet): String {
@@ -85,6 +92,13 @@ class KspConfigurations(private val project: Project) {
         project.plugins.withType(KotlinBasePluginWrapper::class.java).configureEach {
             // 1.6.0: decorateKotlinProject(project.kotlinExtension)?
             decorateKotlinProject(project.extensions.getByName("kotlin") as KotlinProjectExtension, project)
+        }
+        // Create sourceSet-specific KSP configurations for the case when the KotlinBaseApiPlugin is applied instead
+        // of the KotlinBasePluginWrapper (e.g., when AGP's built-in Kotlin support is enabled).
+        project.plugins.withType(KotlinBaseApiPlugin::class.java) {
+            // FIXME: After KT-70897 is fixed and AGP's built-in Kotlin support adds a `kotlin` extension, call
+            //  decorateKotlinProject here instead.
+            createAndroidSourceSetConfigurations(project, kotlinTarget = null)
         }
     }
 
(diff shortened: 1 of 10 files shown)


gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
```diff
@@ -399,4 +399,19 @@ class GradleCompilationTest {
...
+    @Test
+    fun androidGradlePluginBuiltInKotlin() {
+        testRule.setupAppAsAndroidApp(enableAgpBuiltInKotlinSupport = true)
+        testRule.appModule.dependencies.addAll(
+            listOf(
+                artifact(configuration = "ksp", "androidx.room:room-compiler:2.4.2"),
+                artifact(configuration = "kspTest", "androidx.room:room-compiler:2.4.2")
+            )
+        )
+        testRule.runner().withDebug(true).withArguments(":app:assembleDebug").build()
```
ting-yuan, line 415: Please remove `.withDebug(true)`.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
```diff
@@ -86,6 +86,16 @@ class KspConfigurations(private val project: Project) {
...
+        // Create sourceSet-specific KSP configurations for the case when the KotlinBaseApiPlugin is applied instead
+        // of the KotlinBasePluginWrapper (e.g., when AGP's built-in Kotlin support is enabled).
+        project.plugins.withType(KotlinBaseApiPlugin::class.java) {
+            AndroidPluginIntegration.forEachAndroidSourceSet(project) { sourceSet ->
+                createConfiguration(
+                    name = getAndroidConfigurationName(kotlinTargetName = "", sourceSet),
+                    readableSetName = "$sourceSet (Android)"
+                )
+            }
+        }
```
ting-yuan, line 98: It looks like the `kotlin` extension [can be obtained](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinBaseApiPlugin.kt#L64) from `KotlinBaseApiPlugin`. Have you tried to just call `decorateKotlinProject(...)`?
1. It will cover the case where `KotlinBaseApiPlugin` is applied in the context other than AGP.
2. also much easier to maintain.

If `kotlin` doesn't work, either it should be initialized properly by AGP, and/or the configurations should only be created when AGP's built-in Kotlin compilation is applied. From maintenance's point of view, I'd also hoist the common code of `AndroidPluginIntegration.forEachAndroidSourceSet...` into a private function.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
```diff
@@ -86,6 +86,16 @@ class KspConfigurations(private val project: Project) {
             // 1.6.0: decorateKotlinProject(project.kotlinExtension)?
             decorateKotlinProject(project.extensions.getByName("kotlin") as KotlinProjectExtension, project)
         }
+        // Create sourceSet-specific KSP configurations for the case when the KotlinBaseApiPlugin is applied instead
+        // of the KotlinBasePluginWrapper (e.g., when AGP's built-in Kotlin support is enabled).
```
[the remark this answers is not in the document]
scott-pollom: It's possible that they would both be applied, but there's currently no use case for it.

If they are both applied, this will still work because of the `maybeCreate` in `createConfigurations`.

gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
scott-pollom: Done

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
scott-pollom: > It looks like the kotlin extension can be obtained from KotlinBaseApiPlugin. Have you tried to just call decorateKotlinProject(...)?

I think that's the right long-term approach, but currently AGP doesn't add the `kotlin` extension (block by [KT-70897](https://youtrack.jetbrains.com/issue/KT-70897/Add-KotlinBaseApiPlugin.kotlinAndroidExtension)). After that is fixed and AGP adds the `kotlin` extension, I can modify this to instead call `decorateKotlinProject()`.

> From maintenance's point of view, I'd also hoist the common code of AndroidPluginIntegration.forEachAndroidSourceSet... into a private function.

Done

ting-yuan (approved):

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/DependencyResolutionCheckIT.kt
```diff
@@ -23,19 +23,16 @@ import org.junit.Rule
...
+class DependencyResolutionCheckIT(useKSP2: Boolean) {
     @Rule
     @JvmField
     val project: TemporaryTestProject = TemporaryTestProject("playground-android-multi", "playground", useKSP2)
 
     @Test
     fun testDependencyResolutionCheck() {
-        val gradleRunner = GradleRunner.create().withProjectDir(project.root).withGradleVersion("7.6.3")
-
-        File(project.root, "gradle.properties").appendText("\nagpVersion=7.4.1")
```
ting-yuan, line 38: Can we keep this? We need to cover users of older Gradle and AGP versions.

scott-pollom: I only changed this because the CI test was failing with the following error:

```
Could not determine the dependencies of task ':workload:compileDebugKotlin'.
    > Failed to install the following SDK components:
          tools Android SDK Tools
```

It was passing for me locally, so I wasn't sure how to troubleshoot.

This test isn't really specific to AGP 7.4.1... it will fail similarly with AGP 8.7.0 if resolving dependencies early.

FWIW, there is still AGP731IT which is a similar test using AGP 7.3.1 and the same Gradle version.

But the bottom line is that I'm happy to leave this test as-is, but I don't know how to troubleshoot the test failure :)

Issue #1075, opened 2022-08-20
Question: how to make getSymbolsWithAnnotation() method return annotations also from dependency modules (like getDeclarationsFromPackage method) and not only from sourceset?


Marcoral:
Hi
I'm working on a Kotlin project whose logic is heavily annotation-driven. KSP is helping me a lot with this.
Unfortunately, I've been facing a problem for quite a while now, which I don't know how to solve on my own - it's in the ticket title.
I noticed that a similar question has been asked before:

#527 

but I feel that the answer contained in it doesn't suit my case.

I have prepared a small example of what I am trying to achieve:
[https://github.com/Marcoral/KSP-resolving-problem](https://github.com/Marcoral/KSP-resolving-problem)

Perhaps the getSymbolsWithAnnotation() method should contain some kind of flag to control whether symbols from sourceset only or also from dependencies should be visible? I noticed that the getDeclarationsFromPackage() method can detect the latter as well, but in my example w I can't use it (the annotation processor can't be changed with the creation of new libraries providing the annotations it should detect).

I would be very grateful for the help provided

neetopia:
It is not supported in KSP, and potential implementation to support this will be very slow as it needs to scan the whole classpath, therefore we do not have a plan to support this use case.

`getDeclarationsFromPackage` is a different story as getting symbols from class path given package name is efficient enough as the path to look up symbols are determined.

neetopia:
Closed as answered.

eugene-krivobokov:
> getDeclarationsFromPackage is a different story as getting symbols from class path given package name is efficient enough as the path to look up symbols are determined.

Does it mean that `getDeclarationsFromPackage` should see classes from the whole classpath?

PR #2136, opened 2024-10-08
Kdoc: more details about what classpath is searched by Resolver

martinbonnin:
Kdoc only change. 

See https://github.com/google/ksp/issues/1075


ting-yuan (approved):
Thanks for the patch!

PR #2141, opened 2024-10-09
UPDATE_KOTLIN_VERSION: 2.1.0-dev-7621

ting-yuan:
Only K1 and KGP is updated here. KSP2 is awaiting for the fix of [this regression](https://youtrack.jetbrains.com/issue/KT-72002/Analysis-API-psi-KaTypeParameterSymbol-for-default-Java-constructor-is-null) to come in the next Kotlin dev build.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
@@ -50,8 +50,10 @@ import com.intellij.openapi.vfs.StandardFileSystems
 import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.psi.PsiJavaFile
 import com.intellij.psi.PsiManager
+import com.intellij.psi.impl.file.impl.JavaFileManager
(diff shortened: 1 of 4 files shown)


compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
```diff
@@ -509,3 +513,13 @@ fun findTargetInfos(languageVersionSettings: LanguageVersionSettings, module: Mo
             else -> UnknownPlatformInfoImpl(platform.toString())
         }
     } ?: emptyList()
+
+// FIXME: remove as soon as possible.
```
dx404, line 517: What is the plan (on what condition) to remove this wordkaround

dx404 (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/KotlinSymbolProcessingExtension.kt
ting-yuan: The fix just landed in the Kotlin compiler and should be available in the next bootstrap release and Kotlin 2.1.0-RC.

PR #2143, opened 2024-10-10
Ensure all opened input streams are closed

ansman:


ZacSweers (approved):

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/common/PersistentMap.kt
```diff
@@ -53,7 +54,7 @@ abstract class PersistentMap<K, V>(
         @OptIn(ExperimentalSerializationApi::class)
         protected fun <K, V> deserialize(serializer: KSerializer<Map<K, V>>, storage: File): MutableMap<K, V> {
             return if (storage.exists()) {
-                storage.inputStream().use {
+                BufferedInputStream(storage.inputStream()).use {
```
ansman, line 57: From what I can tell, `Json.decodeFromStream` doesn't buffer internally

ting-yuan (approved):

PR #2156, opened 2024-10-18
Ignore default annotation values for classes without found files

Nillerr:
Since the release of Kotlin 2, I have experienced issues with KSP in [mockative/mockative](https://github.com/mockative/mockative). I finally took the time to investigate if there's a way to resolve it, and it turns out the only issue remaining is this:

```
java.lang.NullPointerException
	at com.google.devtools.ksp.impl.symbol.kotlin.UtilKt.getDefaultValue(util.kt:692)
	at com.google.devtools.ksp.impl.symbol.kotlin.UtilKt.getDefaultValue(util.kt:657)
	at com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSAnnotationResolvedImpl$defaultArguments$2.invoke(KSAnnotationResolvedImpl.kt:92)
	at com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSAnnotationResolvedImpl$defaultArguments$2.invoke(KSAnnotationResolvedImpl.kt:66)
    ...
```

Caused by the following code in **utils.kt**:

```kotlin
val file = analyze {
    (fileManager.findClass(classId, analysisScope) as JavaClassImpl).virtualFile!!.contentsToByteArray()
}
```

Specifically, `findClass` returns `null` for the annotation `kotlin.annotation.Target`.

I haven't been able to identify what in Mockative causes this, as I receive the `NullPointerException` in the call to:

```kotlin
resolver.getSymbolsWithAnnotation(...)
```

By changing the call in **utils.kt** to return `null` if the class (and thus file) cannot be found, Mockative works once again. 

I see a similar change is being implemented in `1.0.26`, however I would it seems like that's targeting Kotlin 2.1, so I would love if we could include this change in a `2.0.21-1.0.26` release so users of Mockative can finally upgrade to Kotlin 2.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
@@ -632,8 +632,9 @@ internal fun KaValueParameterSymbol.getDefaultValue(): KaAnnotationValue? {
                 val parentClass = this.getContainingKSSymbol()!!.findParentOfType<KSClassDeclaration>()
                 val classId = (parentClass as KSClassDeclarationImpl).ktClassOrObjectSymbol.classIdIfNonLocal!!
                 val file = analyze {
-                    (fileManager.findClass(classId, analysisScope) as JavaClassImpl).virtualFile!!.contentsToByteArray()
-                }
+                    fileManager.findClass(classId, analysisScope)
+                        ?.let { javaClass -> (javaClass as JavaClassImpl).virtualFile!!.contentsToByteArray() }
+                } ?: return null
                 var defaultValue: JavaAnnotationArgument? = null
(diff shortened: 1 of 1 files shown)


ting-yuan:
Hi @Nillerr, thanks for the patch! This was [fixed](https://github.com/google/ksp/pull/2103) recently. Would you mind to give the [nightly builds](https://github.com/google/ksp?tab=readme-ov-file#nightly-builds) a try? The stable release 1.0.26 is also coming soon.

Nillerr:
Hi @ting-yuan. Thank you for the reply, and thank you for directing my attention to the nightly builds.

 I wasn't sure if that fix was going into Kotlin 2.0.21, or if it was targeting Kotlin 2.1, which is why I decided to submit this PR.

The changes applied in `2.0.21-1.0.26-SNAPSHOT` work, so I will close this PR and look out for the release of KSP 1.0.26. Thank you for your work on this project! 👏🏻

Issue #2174, opened 2024-10-25
Fail to apply ksp before android


ting-yuan:
Discussion on Slack: https://kotlinlang.slack.com/archives/C013BA8EQSE/p1729789842772099

People reporting that after updating to ksp 1.0.26 they got the following error in all the modules that have the ksp line before the android one. Was not the case for as far as this project existed and the first versions of ksp.

```
> Failed to apply plugin class 'org.jetbrains.kotlin.gradle.plugin.KotlinBaseApiPlugin'.
   > Extension with name 'android' does not exist. Currently registered extension names: [ext, sonar, sonarqube, libs, versionCatalogs, projects, ksp, base, defaultArtifacts, spotless, reporting, detekt]
```

scott-pollom:
This is caused by [this Gradle bug](https://github.com/gradle/gradle/issues/31092).

There is a workaround that should work, so I'll work on a implementing that.

PR #2183, opened 2024-10-31
Fix error when KSP plugin applied before AGP

scott-pollom:
Previously, when the KSP plugin was applied before AGP and AGP's built-in Kotlin support for test fixtures or screenshot tests was enabled, there would be an error when building.

The underlying cause for this error is
https://github.com/gradle/gradle/issues/31092, which caused the KSP plugin to try to access the android extension before AGP's apply method had completed.

This change fixes the problem by using the "com.android.base" plugin as the indicator that a module is an Android module. The "com.android.base" plugin is applied after the android extension has been added, so it works around the Gradle bug.

Bug: https://github.com/google/ksp/issues/2174
Test: GradleCompilationTest


ting-yuan (approved):
Thank you so much for the patch! Would you mind to fix the lint error?

PR #2185, opened 2024-10-31
Fix JVM names for properties starting with 'is'

### LINKED ISSUE #2181: [KSP2] JVM names for property getters/setters are incorrect if they start with "is"
## Summary
JVM names for Kotlin property getters & setters are incorrect when the Kotlin property name starts with "is". Specifically, `Resolver.getJvmName` is returning a name that does not match what the Kotlin compiler generated.

### Environment
KSP: 2.1.0-Beta2-1.0.26 with KSP2
Kotlin: 2.1.0-Beta2
AGP: 8.7.1
Gradle: 8.10

## Context
Consider the following code:
```
package dev.jonamireh

interface TestInterface {
    var isOpen: String
}
```

* For the property getter of `isOpen`, `Resolver.getJvmName()`  is returning `getIsOpen` even though the Kotlin compiler generated `isOpen`. 
* For the property setter of `isOpen`, `Resolver.getJvmName()`  is returning `setIsOpen` even though the Kotlin compiler generated `setOpen`. 

This permuation is documented in the Kotlin documentation for [Calling Kotlin from Java](https://kotlinlang.org/docs/java-to-kotlin-interop.html#properties):
> If the name of the property starts with `is`, a different name mapping rule is used: the name of the getter will be the same as the property name, and the name of the setter will be obtained by replacing `is` with `set`. For example, for a property `isOpen`, the getter will be called `isOpen()` and the setter will be called `setOpen()`. This rule applies for properties of any type, not just Boolean.

## Sample project
[Here's a sample project](https://github.com/user-attachments/files/17579057/is_getters.zip) where `:processor` will print the JVM name according to `getJvmName` for each property getter & setter. You can test out what KSP thinks the getter/setter should be with:
```
./gradlew :jvm:assemble --info --rerun-tasks | grep "\[ksp\]"
```

jonamireh:
Fixes #2181

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
@@ -458,16 +458,18 @@ class ResolverAAImpl(
             return accessor.receiver.simpleName.asString()
         }
 
-        val prefix = if (accessor is KSPropertyGetter) {
-            "get"
-        } else {
-            "set"
+        val name = accessor.receiver.simpleName.asString()
+        val uppercasedName = name.replaceFirstChar(Char::uppercaseChar)
+        // https://kotlinlang.org/docs/java-to-kotlin-interop.html#properties
+        val prefixedName = when(accessor) {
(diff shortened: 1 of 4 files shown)


ting-yuan (approved):

ting-yuan:
Thanks for the fix!

PR #2200, opened 2024-11-11
Enable both KSP1 and KSP2 in gradle plugin tests

ting-yuan:


Ppeepost4489 (commented):
gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt

Issue #1678, opened 2024-01-10
Android Studio with KSP: After changing kotlin code, the first assembleDebug task fail with compilation error, 2nd attempt always works fine


MaratTursynbek:
After upgrading to Android Studio Hedgehog | 2023.1, AGP to 8.2.0, kotlin to 1.9.21 and ksp to 1.9.21-1.0.16 I started to experience following behaviour of IDE:
1. Change kotlin code
2. Run the app
3. Compilation fails
4. Click run again (no code changes)
5. Compiles and runs the app but takes a long time to build the project (feels like clean rebuild although only assembleDebug command was given.

As a result, what would take a couple of seconds to compile and run has become apprx 2 minutes between each code changes.

Changing XML code works fine. Incremental build is fast and no problems there. After AS Patch 1 was released with AGP 8.2.1 and kotlin 1.9.22 with corresponding ksp version, the problem is still there.

I would appreciate any help in resolving this problem. Developer experience is now very low and it is getting hard to be productive at work. Is there any settings I'm missing or should I reinstall the macOS and AS all together?

Compilation error stack trace shows different things but this one below is the most frequent among them.

My system info:
```
Android Studio Hedgehog | 2023.1.1 Patch 1
Build #AI-231.9392.1.2311.11255304, built on December 26, 2023
Runtime version: 17.0.7+0-17.0.7b1000.6-10550314 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 14.2
GC: G1 Young Generation, G1 Old Generation
Memory: 4096M
Cores: 12
Metal Rendering is ON
Registry:
    external.system.auto.import.disabled=true
    debugger.new.tool.window.layout=true
    ide.text.editor.with.preview.show.floating.toolbar=false
    ide.instant.shutdown=false
    ide.experimental.ui=true
```

```
    id("com.android.application") version "8.2.1" apply false
    id("com.android.library") version "8.2.1" apply false
    id("com.android.test") version "8.2.1" apply false
    id("org.jetbrains.kotlin.android") version "1.9.22" apply false
    id("com.google.devtools.ksp") version "1.9.22-1.0.16" apply false
```

```
[ksp] java.io.IOException: java.lang.ArrayIndexOutOfBoundsException: Array index out of range: 0
	at org.jetbrains.kotlin.com.intellij.util.io.CompressedAppendableFile.loadChunk(CompressedAppendableFile.java:185)
[...]

PR #2211, opened 2024-11-15
#1678 - Set overwrite flag on copyWithTimestamp based on if a file or not

jsoberg:
This should fix the `FileAlreadyExistsException` from #1678, as called out by @jonamireh [here](https://github.com/google/ksp/issues/1678#issuecomment-2332157791). Ultimately this appears to be a symptom of a larger issue; If this fixes the repeated `FileAlreadyExistsException` thrown during rebuilds in the meantime, however, it would be a good bridge for maintaining developer velocity since this issue causes almost every build to require a manual rebuild (or, disabling incremental KSP builds) for the team that I work on.


jsoberg:
Added further explanation [here](https://github.com/google/ksp/issues/1678#issuecomment-2479696403)

jaschdoc:
Closing this PR since it's no longer relevant. Thank you! :)

PR #2244, opened 2024-11-29
Adapt to removal of old JVM backend

udalov:
We're in the process of removing old JVM backend in the Kotlin repo, which leads to removing some internal compiler API that was only used from the old backend. See [KT-71197](https://youtrack.jetbrains.com/issue/KT-71197) for all the changes.

Apparently some of the utilities, specifically those in `KotlinTypeMapper`, were used in KSP. In this change, I've adapted KSP sources to make it compilable against Kotlin master. Mainly I copy-pasted functions from `KotlinTypeMapper` which were removed [here](https://github.com/JetBrains/kotlin/compare/4f5c5a3a462e...cd3ec31ee46c) and simplified them.

I have some questions regarding the semantics of, for example, the newly introduced `mapPropertySignature`, which either returns a generic signature or an erased JVM type descriptor, depending on whether the field is generic, but for now I just restored the old behavior without (hopefully) any changes.

I'm marking this PR as a draft to make sure that it's not merged blindly, it should only be incorporated during the next update of Kotlin in KSP.

### DIFF
--- a/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
+++ b/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
@@ -57,7 +57,7 @@ import com.intellij.psi.impl.source.PsiClassReferenceType
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.builtins.jvm.JavaToKotlinClassMap
 import org.jetbrains.kotlin.codegen.ClassBuilderMode
-import org.jetbrains.kotlin.codegen.OwnerKind
+import org.jetbrains.kotlin.codegen.signature.BothSignatureWriter
 import org.jetbrains.kotlin.codegen.state.KotlinTypeMapper
 import org.jetbrains.kotlin.config.LanguageVersionSettingsImpl
 import org.jetbrains.kotlin.container.ComponentProvider
@@ -66,9 +66,11 @@ import org.jetbrains.kotlin.container.tryGetService
 import org.jetbrains.kotlin.descriptors.*
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
 import org.jetbrains.kotlin.descriptors.annotations.Annotations
+import org.jetbrains.kotlin.fileClasses.JvmFileClassUtil
 import org.jetbrains.kotlin.incremental.components.NoLookupLocation
 import org.jetbrains.kotlin.load.java.components.TypeUsage
 import org.jetbrains.kotlin.load.java.descriptors.JavaForKotlinOverridePropertyDescriptor
+import org.jetbrains.kotlin.load.java.descriptors.getImplClassNameForDeserialized
 import org.jetbrains.kotlin.load.java.lazy.*
 import org.jetbrains.kotlin.load.java.lazy.descriptors.LazyJavaTypeParameterDescriptor
 import org.jetbrains.kotlin.load.java.lazy.types.JavaTypeResolver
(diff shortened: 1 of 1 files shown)


jsjeon:
cc @ting-yuan

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -356,20 +360,24 @@ class ResolverImpl(
     override fun mapToJvmSignature(declaration: KSDeclaration): String? = mapToJvmSignatureInternal(declaration)
 
     internal fun mapToJvmSignatureInternal(declaration: KSDeclaration): String? = when (declaration) {
-        is KSClassDeclaration -> resolveClassDeclaration(declaration)?.let { typeMapper.mapType(it).descriptor }
+        is KSClassDeclaration -> resolveClassDeclaration(declaration)?.let { typeMapper.mapType(it.defaultType).descriptor }
```
jsjeon, line 363: This line is long, causing linter issue:
```
/home/runner/work/ksp/ksp/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt:363:1: Exceeded max line length (120) (cannot be auto-corrected)
```
Please add line wrap around braces.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -356,20 +360,24 @@ class ResolverImpl(
...
                 is FunctionDescriptor -> typeMapper.mapAsmMethod(it).descriptor
-                is PropertyDescriptor -> typeMapper.mapFieldSignature(it.type, it) ?: typeMapper.mapType(it).descriptor
+                is PropertyDescriptor -> typeMapper.mapPropertySignature(it)
                 else -> throw IllegalStateException("Unexpected descriptor type for declaration: $declaration")
             }
         }
-        is KSPropertyDeclaration -> resolvePropertyDeclaration(declaration)?.let {
-            typeMapper.mapFieldSignature(it.type, it) ?: typeMapper.mapType(it).descriptor
-        }
+        is KSPropertyDeclaration -> resolvePropertyDeclaration(declaration)?.let { typeMapper.mapPropertySignature(it) }
```
jsjeon, line 371: Nit: it is exactly 120 line length, but keeping the same structure (and just rewrite `mapFieldSignature` with `mapPropertySignature`) would make the overall change look minimal.

compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt
```diff
@@ -880,22 +888,13 @@ class ResolverImpl(
 
     @KspExperimental
     override fun getJvmName(accessor: KSPropertyAccessor): String? {
-        val descriptor = resolvePropertyAccessorDeclaration(accessor)
-
-        return descriptor?.let {
-            // KotlinTypeMapper.mapSignature always uses OwnerKind.IMPLEMENTATION
-            typeMapper.mapFunctionName(descriptor, OwnerKind.IMPLEMENTATION)
-        }
+        return resolvePropertyAccessorDeclaration(accessor)?.let(typeMapper::mapFunctionName)
```
jsjeon, line 891: This (and below) currently failed to build:
```
> Task :compiler-plugin:compileKotlin FAILED
e: file://.../ksp/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt:891:66 Type mismatch: inferred type is KFunction2<FunctionDescriptor, OwnerKind?, String> but (TypeVariable(T)) -> TypeVariable(R) was expected
e: file://.../ksp/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/processing/impl/ResolverImpl.kt:897:86 Type mismatch: inferred type is KFunction2<FunctionDescriptor, OwnerKind?, String> but (TypeVariable(T)) -> TypeVariable(R) was expected
```
but that's because you didn't upgrade kotlin version: https://github.com/google/ksp/blob/main/gradle.properties#L4

Why don't you change that line to, e.g., `2.1.20-dev-5634`?

I found that upgrading AA (`aaKotlinBaseVersion`) causes other problems, but those should be addressed separately.

jsjeon (changes requested):

jsjeon:
FYI, I created https://github.com/google/ksp/pull/2255 instead where every check is passing now.

PR #2295, opened 2025-01-16
Add visibility check in ResolverAAImpl.overrides

### LINKED ISSUE #2254: [KSP2] Resolver#overrides() gives incorrect result when method isn't accessible.
Consider the following two classes:

```java
package a;
public class A {
  void method() { ... }
} 

package c;
public class C extends A {
  void method() { ... }
}
```

When using `Resolver#overrides(methodC, methodA)` we get different results in KSP1 and KSP2:

  * KSP1: `Resolver#overrides(methodC, methodA) == false`
  * KSP2: `Resolver#overrides(methodC, methodA) == true`

I think KSP2's result is wrong in this case because `A#method()` is not visible from class `B`, so rather than an override it's more like shadowing.

ting-yuan:
override cannot happen without visibility.

fixes #2254

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
@@ -659,6 +659,8 @@ class ResolverAAImpl(
         }
         recordLookupForPropertyOrMethod(overrider)
         recordLookupForPropertyOrMethod(overridee)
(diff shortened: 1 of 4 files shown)


jsjeon:
Those `PackagePrivate(Base)` classes are `public`, not really package private. Just to double-check if those don't matter (i.e., naming issue).

dx404 (approved):

ting-yuan:
> Those `PackagePrivate(Base)` classes are `public`, not really package private. Just to double-check if those don't matter (i.e., naming issue).

Updated naming of those classes for clarity. Thanks!

PR #2301, opened 2025-01-21
Try to find KaFirPsiJavaClassSymbol if possible

### LINKED ISSUE #2246: Resolver returns non-equal instances of the same declaration.
In KSP1, all instance of a `KSDeclaration` for the same declaration were equal (this is a similar to the expectation in javac for `Element` instances).

However, in KSP2 we can get two different (non-equal) instances for the same declaration depending on if we get the instance using `Resolver#getClassDeclarationByName()` or `Resolver#getSymbolsWithAnnotation()`, e.g. suppose we have:

```kotlin
@MyAnnotation class MyClass
```

And the following processor:

```kotlin
  override fun process(resolver: Resolver): List<KSAnnotated> {
    val declaration1 = resolver.getSymbolsWithAnnotation("MyAnnotation").single()
    val declaration2 = resolver.getClassDeclarationByName( resolver.getKSNameFromString("MyClass"))!!
    println("declaration1.equals(declaration2): ${declaration1.equals(declaration2)}")
    return listOf()
  }
```

This results in:

```
// KSP1:
declaration1.equals(declaration2): true

// KSP2:
declaration1.equals(declaration2): false
```

ting-yuan:
This works around an issue that equals not working for KaFirNamedClassSymbol and KaFirPsiJavaClassSymbol. See KT-74598 for details.

Fixes #2246

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
@@ -29,6 +29,7 @@ import com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSClassifierParameter
 import com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSClassifierReferenceResolvedImpl
 import com.google.devtools.ksp.impl.symbol.util.getDocString
 import com.google.devtools.ksp.symbol.*
+import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiJavaFile
 import com.intellij.psi.impl.compiled.ClsMemberImpl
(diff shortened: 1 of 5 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
```diff
@@ -404,7 +405,10 @@ internal fun ClassId.toKtClassSymbol(): KaClassSymbol? {
                 it.asString() == this@toKtClassSymbol.shortClassName.asString()
             }?.singleOrNull() as? KaClassSymbol
         } else {
-            findClass(this@toKtClassSymbol)
+            // Try to find KaFirPsiJavaClassSymbol. See KT-74598 for details.
+            findClass(this@toKtClassSymbol)?.let {
+                (it.psi as? PsiClass)?.namedClassSymbol ?: it
```
jsjeon, line 410: Instead of this unconditional trial, I wonder if you can check `origin` first, like `KaSymbolOrigin.JAVA_*`?

ting-yuan: Good point. Check of source is actually needed because psi can come from libraries 🤷

jsjeon (approved):

dx404 (approved):

Issue #2250, opened 2024-12-05
Migrate KSP off AGP's legacy Variant API


scott-pollom:
`AndroidPluginIntegration` is using the deprecated AGP Variant API and should migrate to the new AGP Variant API.

E.g., `KotlinJvmAndroidCompilation.androidVariant` returns an instance of the deprecated `com.android.build.gradle.api.BaseVariant`.

scott-pollom:
Related bug, blocking the migration for the usage in `AndroidPluginIntegration::getCompilationSourceSets`: https://issuetracker.google.com/issues/317215060

scott-pollom:
> Related bug, blocking the migration for the usage in AndroidPluginIntegration::getCompilationSourceSets: https://issuetracker.google.com/issues/317215060

The new API was added in AGP 8.9.0-alpha09, so KSP must update its AGP dependency to at least that version to migrate.

Jolanrensen:
After this change (since version 1.0.30), KSP no longer works in our java-8 targeting library.

We're trying to target Kotlin 2.1.20, so the lowest available version is 1.0.31, which is past this change. I couldn't find a mention of a minimum JDK version for the gradle plugin anywhere, so I'd assume while Kotlin supports java 8, KSP would too.

hungvietnguyen:
cc @ting-yuan

In AGP 9.0 (canaries coming soon), access to the old variant API will be disabled by default (https://developer.android.com/build/releases/gradle-plugin-roadmap), so it would be great to finish the migration away from the legacy variant API.

Also related bug in KGP: https://youtrack.jetbrains.com/issue/KT-77300/

liutikas:
Any updates here?

hfmehmed:
I am trying to work out a plan for this.
What will probably end up happening is:
- KSP2 will not work with AGP 9.0 + kotlin-android (applied separately) without an explicit opt-in flag for the usage of old variant api
- KSP2 will work with AGP 9.0 with embedded Kotlin (in other words without kotlin-android plugin)
- KSP1 is deprecated and we will probably just state that it doesnt work with AGP 9.0 at all (or it may be broken)

The focus will be more on AGP with embedded kotlin rather than fixing kotlin-android itself to not rely on old variant api (which is possible too but will take much more effort to coordinate with JB)
How does that sound from AndroidX point of view @liutikas ?

mkmuir0:
This has been fixed in the latest 2.2.20-2.0.3 KSP release

[...]

PR #2312, opened 2025-01-30
Partially migrate KSP off AGP's legacy Variant API

scott-pollom:
Use AGP's new addKspConfigurations() API when possible.

Bug: https://github.com/google/ksp/issues/2250
Test: AGP741IT and existing ProcessorClasspathConfigurationsTest

### DIFF
--- a/gradle-plugin/build.gradle.kts
+++ b/gradle-plugin/build.gradle.kts
@@ -1,3 +1,4 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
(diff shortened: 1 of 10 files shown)


scott-pollom:
There is a test on the AGP side here: https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/integration-test/application/src/test/java/com/android/build/gradle/integration/application/ConfigurationPerSourceSetTest.kt;l=162

And there is a test on the KSP side in ProcessorClasspathConfigurationsTest.kt

AGP's API creates the same configurations previously created by KSP. The only difference is that it creates the `...KotlinProcessorClasspath` configurations eagerly.

ting-yuan (approved):

PR #2319, opened 2025-02-03
KSP2: reimplement Resolver.getDeclarationsFromPackage

### LINKED ISSUE #2309: Resolver#getDeclarationsFromPackage() does not find generated files
This can be reproduced with the following:

```kotlin
var round = 1

override fun process(resolver: Resolver): List<KSAnnotated> {
  if (round == 0) {
    createJavaFile("foo.bar", "MyGeneratedJavaClass")
  }
  println("Processing round $round")
  println("resolver.getClassDeclarationByName: ${
    resolver.getClassDeclarationByName(
      resolver.getKSNameFromString("foo.bar.MyGeneratedJavaClass")
    )?.qualifiedName?.asString()
  }")
  println("resolver.getDeclarationsFromPackage: ${
    resolver.getDeclarationsFromPackage("foo.bar").toList().map { it.qualifiedName?.asString() }
  }")
  round++
  return listOf()
}

private fun createJavaFile(packageName: String, className: String) {
  val dependencies = Dependencies(aggregating = false, sources = arrayOf())
  codeGenerator.createNewFile(dependencies, packageName, className, "java").use {
    it.bufferedWriter(Charsets.UTF_8).use {
       JavaFile.builder(packageName, TypeSpec.classBuilder(className).build())
           build()
          .writeTo(it)
    }
  }
}
```

Results in:

```
Processing round 1
resolver.getClassDeclarationByName: null
resolver.getDeclarationsFromPackage("foo.bar"): []

Processing round 2
resolver.getClassDeclarationByName: foo.bar.MyGeneratedJavaClass
resolver.getDeclarationsFromPackage("foo.bar"): []
```

Note that in round 2, after `foo.bar.MyGeneratedJavaClass` has been generated, it can be found by `resolver.getClassDeclarationByName()` but not by `resolver.getDeclarationsFromPackage()`

ting-yuan:
fixes #2309

### DIFF
--- a/test-utils/testData/api/packageProviderForGenerated.kt
+++ b/test-utils/testData/api/packageProviderForGenerated.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2025 Google LLC
+ * Copyright 2010-2025 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TEST PROCESSOR: PackageProviderForGeneratedProcessor
+// EXPECTED:
+// Processing round 0
+// resolver.getClassDeclarationByName: null
+// resolver.getDeclarationsFromPackage: []
+// Processing round 1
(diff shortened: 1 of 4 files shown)


test-utils/testData/api/packageProviderForGenerated.kt
```diff
@@ -0,0 +1,29 @@
...
+// Processing round 0
+// resolver.getClassDeclarationByName: null
+// resolver.getDeclarationsFromPackage: []
+// Processing round 1
+// resolver.getClassDeclarationByName: foo.bar.MyGeneratedJavaClass
+// resolver.getDeclarationsFromPackage: [foo.bar.MyGeneratedJavaClass]
+// END
+
+// FILE: Dummy.kt
+class Dummy
```
dx404, line 29: Does it matter whether `Dummy` is from a default package (this test class) or `foo.bar`?

dx404 (approved):

test-utils/testData/api/packageProviderForGenerated.kt
ting-yuan: It doesn't matter. It is a non-empty file that makes sure the processor runs.

PR #2326, opened 2025-02-08
fix typo exmample -> example

robstoll:


ting-yuan (approved):

ting-yuan:
thanks for the correction!

PR #2331, opened 2025-02-11
fix `KSTypeNotPresentException` for an existing nesting class

RishatShamsutdinov:
### Source
```
object Root {
    class Nested
}

annotation class MyAnnotation(val clazz: KClass<*>)

@MyAnnotation(Root.Nested::class)
fun main() = Unit
```

### Processor
`getAnnotationsByType(MyAnnotation::class).first().clazz`

### Why
`Map.Entry.class.getName()` prints `java.util.Map$Entry`
`Map.Entry.class.getCanonicalName()` prints `java.util.Map.Entry`

ksp's `qualifiedName` for `Map.Entry` is `java.util.Map.Entry`

ksp's `qualifiedName` is equivalent to Java's `Class.getCanonicalName` but `Class.forName` requires fully qualified name (`Class.getName`).


ting-yuan (approved):
Thanks for the fix!

Issue #2325, opened 2025-02-07
[KSP2] KSType.arguments return arguments on aliased type.


bcorso:
This seems like another case of https://github.com/google/ksp/issues/2088.

Suppose we have something like this:

```
val field: MyAlias<String>
typealias MyAlias<T> = Foo<Bar<T>, Baz<T>>
```

The `KSType` for the field gives a different result in KSP1 v KSP2.

```
// KSP1
fieldType.arguments // [String]

// KSP2
fieldType.arguments // [Bar<String>, Baz<String>]
```

I think KSP1 behavior is more correct in this case.

In particular, the `KSType` should represent the alias, `MyAlias<String>`, rather than the aliased type `Foo<Bar<String>, Baz<String>>`.

Hamoo456:
nice

bcorso:
Note: This has been fixed in 1.0.30 but adding more details since details in the original post were not enough to reproduce the issue.

Consider the following source:

```kotlin
class Subject(val param: MyAlias<String>)
typealias MyAlias<T> = Foo<Bar<T>, Baz<T>>
class Foo<T1, T2>
class Bar<T>
class Baz<T>
```

With the following processor logic:

```kotlin
val subjectName = resolver.getKSNameFromString("test.Subject")
val subject = resolver.getClassDeclarationByName(subjectName)!!
val constructor = subject.getConstructors().single()
val type1 = constructor.parameters.single().type.resolve()
val type2 = constructor.asMemberOf(subject.asType(emptyList())).parameterTypes.single()!!

println("type1: $type1")
println("type2: $type2")
```

Results in:

```
BACKEND: KSP1
type1: [typealias MyAlias]<String>
type2: [typealias MyAlias]<String>

BACKEND: KSP2
type1: [typealias MyAlias]
type2: Foo<Bar<String>, Baz<String>>
```

The issue happens when calling `KSFunctionDeclaration#asMemberOf(KSType)`. In particular, KSP2 seems to automatically resolve the type alias whereas KSP1 does not.

PR #2335, opened 2025-02-13
KSFunction: return type aliases if possible

ting-yuan:
AA expands types by default. Luckily, the abbreviations are handily available. This change makes KSFunction behave the same as other parts that return KSType in the API.


ting-yuan:
Fixes #2325. Tests provided and reviewed by @bcorso

Issue #2340, opened 2025-02-14
KSP2 Causes Cache Misses due to `java.home` Input


cdsap:
#### Description
When using `ksp.useKSP2=true`, the `KspAATask` uses `java.home` as an input, causing cache misses when the value differs across environments or the same environments but with different minor or patch versions of the JDK.
![Image](https://github.com/user-attachments/assets/4db29875-dbda-43c0-9b8f-fc78236e10d8)

#### Steps to Reproduce:
1. Clone the `nowinandroid` project
2. Add `ksp.useKSP2=true` in the gradle.properties file.
3. Update the Room dependency to 2.7.0-beta01 in libs.toml
4. Execute the KSP task `./gradlew clean :core:database:kspProdDebugKotlin`
5. Change the minor or patch version of the JDK used (e.g., switch between JDK 21.0.6 and 21.0.7).
6. Execute the task again: `./gradlew clean :core:database:kspProdDebugKotlin`

**Expected Result:**
The cache entry is reused because no relevant input has changed.

 **Current Result:**
A cache miss occurs due to a different value in kspConfig.jdkHome `Cache miss: Input property 'kspConfig.jdkHome' has different value`

#### Resources
Build scan comparison https://ge.solutions-team.gradle.com/c/hjzl54ql2dgum/jgbbvfetqfljm/task-inputs

PR #2341, opened 2025-02-14
jdkHome as internal and creating new jdkVersion input

cdsap:
### Proposal to Resolve Issue #2340

This PR is a quick proposal to address issue #2340.  

#### Problem  
The current implementation tracks `java.home` as an input, causing cache misses when the value differs. This not only affects CI vs. local builds but also causes issues in local environments if the user updates the minor or patch versions of Java, as the input value changes.  

#### Proposed Solution  
This PR updates the JDK home input to be marked as `@Internal` and introduces a new input, `jdkVersion`, to track the major Java version.  

The implementation follows the current style by retrieving the system property `java.version`. The purpose of using `jdkVersion` is to ensure that the task is invalidated if the major Java version changes, similar to how Java handles compatibility with the Java compiler.  

#### Concerns  
I am not entirely convinced by this approach because I don't see the current `KspAATask` implementation supporting `KotlinJavaToolChains` as `KspTaskJvm` does. Supporting `KotlinJavaToolChains` could help detect changes in the major version. However, I am unsure of the implementation details regarding why `java.home` is required in `KspAATask` and whether the symbol processor behavior should differ for minor or patch version changes.  

Feedback and suggestions are welcome!

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
@@ -270,6 +270,13 @@ abstract class KspAATask @Inject constructor(
                         // TODO: set proper jdk home
                         cfg.jdkHome.value(File(System.getProperty("java.home")))
 
+                        val javaVersion = System.getProperty("java.version")?.split(".")?.let {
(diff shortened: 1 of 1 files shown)


ting-yuan (approved):

ting-yuan:
LGTM and thanks for the patch!

`java.home` is not required by KSP's Gradle task. The implementation should be reworked.

PR #2351, opened 2025-02-26
KSP2: return type aliases if possible.

ting-yuan:


jsjeon (approved):
I guess you can squash these two commits into one? (In other words, not sure why those split is needed, as the overall things seem like the same idea of using `abbreviationOrSelf`)

dx404 (approved):

PR #2368, opened 2025-03-07
Fix validity check of annotations on properties

ting-yuan:
Annotations on properties may come from corresponding parameters if they are declared in constructors. They may not be valid on properties and need to be filtered out. The previous implementation incorrectly excludes all annotations with @Target(AnnotationTarget.VALUE_PARAMETER), while what should be excluded are the ones with @Target and without @Target(AnnotationTarget.PROPERTY).

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
@@ -64,21 +64,13 @@ class KSPropertyDeclarationImpl private constructor(internal val ktPropertySymbo
             .map { KSAnnotationResolvedImpl.getCached(it, this) }
             .plus(
                 if (ktPropertySymbol.isFromPrimaryConstructor) {
-                    (parentDeclaration as? KSClassDeclaration)?.primaryConstructor?.parameters
-                        ?.singleOrNull { it.name == simpleName }?.annotations ?: emptySequence()
+                    (parentDeclaration as? KSClassDeclaration)?.primaryConstructor?.parameters?.singleOrNull {
+                        it.name == simpleName
+                    }?.annotations?.filter { it.isValidOnProperty() } ?: emptySequence()
                 } else {
                     emptySequence()
                 }
-            ).filterNot { valueParameterAnnotation ->
-                valueParameterAnnotation.annotationType.resolve().declaration.annotations.any { metaAnnotation ->
-                    metaAnnotation.annotationType.resolve().declaration.qualifiedName
-                        ?.asString() == "kotlin.annotation.Target" &&
-                        (metaAnnotation.arguments.singleOrNull()?.value as? ArrayList<*>)?.any {
-                        (it as? KSClassDeclaration)?.qualifiedName
-                            ?.asString() == "kotlin.annotation.AnnotationTarget.VALUE_PARAMETER"
-                    } ?: false
-                }
-            }.plus(
+            ).plus(
                 // TODO: optimize for psi
                 ktPropertySymbol.backingFieldSymbol?.annotations
                     ?.map { KSAnnotationResolvedImpl.getCached(it, this@KSPropertyDeclarationImpl) } ?: emptyList()
@@ -203,6 +195,7 @@ internal fun KaAnnotation.isUseSiteTargetAnnotation(): Boolean {
             it == AnnotationUseSiteTarget.CONSTRUCTOR_PARAMETER
     } ?: false
 }
+
 internal fun KtAnnotationEntry.isUseSiteTargetAnnotation(): Boolean {
     return this.useSiteTarget?.getAnnotationUseSiteTarget()?.let {
         it == AnnotationUseSiteTarget.PROPERTY_GETTER ||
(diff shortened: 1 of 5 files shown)


dx404 (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
```diff
@@ -239,3 +233,11 @@ internal fun KaPropertySymbol.toModifiers(): Set<Modifier> {
     }
     return result
 }
+
+internal fun KSAnnotation.isValidOnProperty(): Boolean =
+    annotationType.resolve().declaration.annotations.none { metaAnnotation ->
+        metaAnnotation.annotationType.resolve().declaration.qualifiedName?.asString() == "kotlin.annotation.Target" &&
```
jsjeon, line 239: Nit: maybe it's just me, but I prefer to avoid using constant as much as possible (in case of, e.g., typo or some other silly mistakes) :)

That said, this can be `Target::class.qualifiedName`? Same for `AnnotationTarget.PROPERTY` below. (But not necessarily in this change.)

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/AnnotationTargetsProcessor.kt
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2020 Google LLC
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
```
jsjeon, line 3: Just to double-check, do these intentionally stay in 2020, or is it a good time to update the template somewhere to make this copyright comment up-to-date?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
```diff
@@ -239,3 +233,11 @@ internal fun KaPropertySymbol.toModifiers(): Set<Modifier> {
...
+            (metaAnnotation.arguments.singleOrNull()?.value as? ArrayList<*>)?.none {
```
jsjeon, line 240: It was previously `any`, but now `none`. Is that expected?

Also, because of the inversion, now the util name sounds counter-intuitive. It is read like, if none of the annotation has non-Property target, it is valid on property?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
```diff
@@ -239,3 +233,11 @@ internal fun KaPropertySymbol.toModifiers(): Set<Modifier> {
...
+            (it as? KSClassDeclaration)?.qualifiedName?.asString() == "kotlin.annotation.AnnotationTarget.PROPERTY"
+        } ?: false
```
jsjeon, line 242: In the code review, it's hard to understand where this elvis operator belongs...

I wonder if the aligned indentation would help? IIUC, this is:
```kt
  (metaAnno...)?.none {
    (it as? ...)
  } ?: false
```
but it's currently:
```kt
  (metaAnno...)?.none {
  (it as? ...)
} ?: false
```
?

test-utils/testData/api/annotationTargets.kt
```diff
@@ -0,0 +1,40 @@
...
+package com.example
+
+@Target(AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
+public annotation class MyAnnotationInLib
+
+// MODULE: main(Anno)
+// FILE: com/exampel/MyClass.kt
+package com.example
+
+@Target(AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
```
jsjeon, line 34: My confusion on `isValidOnProperty` made me wonder what happens if the annotation target doesn't have `PROPERTY`...

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
ting-yuan: I agree. This is a bug of ktlint.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
ting-yuan: This is expected. The default behavior, i.e. without `@Target` on it, allows an annotation to be attached to a property. So an annotation is invalid on a property if and only if it has `@Target` that has no `PROPERTY`. This is equivalent to: an annotation is valid iff none of its meta annotations is "`@Target` without `PROPERTY`".

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationImpl.kt
ting-yuan: Will fix in later refactoring / polishing commits.

test-utils/testData/api/annotationTargets.kt
ting-yuan: If the target doesn't have `PROPERTY`, it is invalid and should not be included in `KSPropertyDeclaration.annotations`.

ting-yuan: btw, we have this case covered in https://github.com/google/ksp/blob/main/kotlin-analysis-api/testData/annotationInDependencies.kt#L115

PR #2369, opened 2025-03-07
Store collected symbols by annotations

ting-yuan:
Previous implementation needs to scan through all symbols with annotations each time getSymbolsWithAnnotation is called. Let's just store the symbols in a map and index by annotations.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
@@ -544,15 +544,10 @@ class ResolverAAImpl(
         }
         val realAnnotationName = expandedIfAlias ?: annotationName
 
-        fun checkAnnotated(annotated: KSAnnotated): Boolean {
-            return annotated.annotations.any {
-                val kaType = (it.annotationType.resolve() as? KSTypeImpl)?.type ?: return@any false
-                kaType.fullyExpand().symbol?.classId?.asFqNameString() == realAnnotationName
-            }
-        }
-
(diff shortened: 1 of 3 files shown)


dx404 (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -565,16 +560,30 @@ class ResolverAAImpl(
...
+        mapAnnotatedSymbols(true)
     }
 
-    private val newAnnotatedSymbols: Collection<KSAnnotated> by lazy {
-        collectAnnotatedSymbols(false)
+    private fun mapAnnotatedSymbols(inDepth: Boolean): Map<String, Collection<KSAnnotated>> {
+        val newSymbols = collectAnnotatedSymbols(inDepth)
+        val withDeferred = newSymbols + deferredSymbolsRestored
+        return mutableMapOf<String, MutableCollection<KSAnnotated>>().apply {
+            withDeferred.forEach { annotated ->
```
jsjeon, line 575: Nit: http://go/kos-rule/DiscourageForEach (internal link)

Instead of `...forEach { ... return@forEach }`, just regular `for` loop and `continue` is easy to read (at least to me...)

jsjeon (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
ting-yuan: Done.

Issue #2348, opened 2025-02-24
[KSP2] KSValueParameterImpl changes field isVal if class defined in library


oscarthecat:
kotlin version 2.1.0
ksp version 2.1.0-1.0.28

we have a `module1/ConsoleLens.kt`, in which we have a 
```kotlin
@LensForType(type = Console::class, intoPackage = "com.foobar.blens.test")
interface ConsoleLens
```
we try to handle `@LensForType` this annotation in out ksp. It refer another Console class which is defined in `module2/Console.kt` like
```
data class Console(
    val brand: String,
    val type: String,
) {
    companion object
}
```
when our ksp runs we check Console class should be data class and all params in primary constructor should be val, but it failed when using ksp2, while it does work when using ksp1.

In ksp2 if `Console.kt` is also defined in module1(as `-source-roots` like `ConsoleLens.kt`), it also works. It failed to our check when `Console.kt` is defined in `-libraries` as a klib, this happen both via gradle build and command line build.

After debug `isVal` we find in KSValueParameterImpl.kt
```
    override val isVal: Boolean
        get() = (ktValueParameterSymbol.psi as? KtParameter)?.let { it.hasValOrVar() && !it.isMutable } ?: false
```
ktValueParameterSymbol (came from `-libraries` klib) it self `isVal` is true, but `ktValueParameterSymbol.psi` is null so finally it returns false.

Any way to get `ktValueParameterSymbol.psi` non null? or any change should we apply to out ksp2 command? Thanks a lot.

ps out ksp2 command is like
```shell
-cp
symbol-processing-common-deps.jar:symbol-processing-aa.jar:symbol-processing-api.jar:artifacts/org.jetbrains.kotlin/kotlin-stdlib/2.1.0/processed_kotlin-stdlib-2.1.0.jar:artifacts/org.jetbrains.kotlinx/kotlinx-coroutines-core-jvm/1.6.4/processed_kotlinx-coroutines-core-jvm-1.6.4.jar
com.google.devtools.ksp.cmdline.KSPCommonMain
-module-name=lens_test_library_ksp2_common
-project-base-dir
bazel-out/darwin_arm64-fastbuild/bin/srcs/base/blens/sample/module/lens_test_library_ksp2_common_generated
-source-roots
srcs/base/blens/sample/module/src/commonMain/kotlin/com/foobar/blens/ConsoleLens.kt (//Use Console class)
[...]

PR #2377, opened 2025-03-14
KSP2: fix KSValueParameter.isVal and isVar

### LINKED ISSUE #1192: Incorrect KSValueParameter#isVal/isVar() when inspecting precompiled class.
I'm running into an issue where `KSValueParameter#isVal()` and `KSValueParameter#isVar()` seem incorrect for a precompiled class like:

```kotlin
class MyClass(
    field: String,
    val valField: String,
    var varField: String,
)
```

In particular, I'm getting the following:

```
 fieldName | isPrecompiled | isVal() | isVar() 
-----------+---------------+---------+----------
  field    |     false     |  false  | false      // Correct
  valField |     false     |  true   | false      // Correct
  varField |     false     |  false  | true       // Correct
-----------+---------------+---------+----------
  field    |     true      |  true   | false      // Incorrect
  valField |     true      |  true   | false      // Correct
  varField |     true      |  true   | false      // Incorrect
```

Note that when `isPrecompiled=true` then `isVal()` always returns `true` and `isVar()` always returns `false`.

ting-yuan:
for symbols from libraries.

Fixes #1192 and #2348

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSValueParameterImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSValueParameterImpl.kt
@@ -24,6 +24,10 @@ import com.google.devtools.ksp.common.lazyMemoizedSequence
 import com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSTypeReferenceResolvedImpl
 import com.google.devtools.ksp.symbol.*
 import org.jetbrains.kotlin.analysis.api.fir.symbols.KaFirValueParameterSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaClassSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaConstructorSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaPropertySymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolOrigin
 import org.jetbrains.kotlin.analysis.api.symbols.KaValueParameterSymbol
 import org.jetbrains.kotlin.analysis.api.types.abbreviationOrSelf
 import org.jetbrains.kotlin.fir.java.JavaTypeParameterStack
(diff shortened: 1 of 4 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSValueParameterImpl.kt
```diff
@@ -77,11 +81,26 @@ class KSValueParameterImpl private constructor(
...
 
+    private val KaValueParameterSymbol.primaryConstructorProperty: KaPropertySymbol? by lazy {
+        when (ktValueParameterSymbol.origin) {
+            // ktValueParameterSymbol.generatedPrimaryConstructorProperty is always null in libraries.
+            // TODO: fix in AA
+            KaSymbolOrigin.LIBRARY, KaSymbolOrigin.JAVA_LIBRARY -> analyze {
+                val cstr = ktValueParameterSymbol.containingDeclaration as? KaConstructorSymbol
+                val cls = cstr?.containingDeclaration as? KaClassSymbol
+                cls?.declaredMemberScope?.declarations?.filterIsInstance<KaPropertySymbol>()
+                    ?.firstOrNull { it.name == ktValueParameterSymbol.name }
```
jsjeon, line 92: In addition to `name` check, you can use `isFromPrimaryConstructor`: https://github.com/JetBrains/kotlin/blob/master/analysis/analysis-api/src/org/jetbrains/kotlin/analysis/api/symbols/KaVariableSymbol.kt#L245 ?

ting-yuan: Unfortunately `isFromPrimaryConstructor` is also PSI based and doesn't work on libraries.

But that reminds me of a case that this patch doesn't cover:
```
class MyClass(prop: Int = 1) {
    val prop: Int = 0
}
```

IMHO it is impossible to distinguish without adding metadata in the libraries.

Anyway, this patch covers most of the use cases and doesn't introduce new issues. If there are no other concerns, I'm going to merge it.

PR #2409, opened 2025-04-11
Mention both ksp1/2 test suite in CONTRIBUTING.md

kevinguitar:
Noticed the test suite path in `CONTRIBUTING.md` is outdated, took a chance to mention KSP2 test suite as well.


ting-yuan (approved):
Thanks for the patch!

PR #2411, opened 2025-04-11
Return static function kind if modifier has static modifier

### LINKED ISSUE #2410: [KSP2] functionKind is MEMBER for static method in interface in Java
For the following Java code:
```java
public interface MyModule {
    static int provides() {
        return 1;
    }
}
```

`KSClassDeclaration.functionKind` returns `FunctionKind.MEMBER`, but `modifiers` contains `Modifier.JAVA_STATIC`

With KSP1 `KSClassDeclarationJavaImpl` returns `FunctionKint.STATIC`

bacecek:
Fixes #2410

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
@@ -44,7 +44,13 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
 
     override val functionKind: FunctionKind by lazy {
         when (ktFunctionSymbol.location) {
-            KaSymbolLocation.CLASS -> FunctionKind.MEMBER
+            KaSymbolLocation.CLASS -> {
+                if (ktFunctionSymbol is KaNamedFunctionSymbol && ktFunctionSymbol.isStatic) {
+                    FunctionKind.STATIC
+                } else {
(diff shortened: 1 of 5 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -43,6 +43,9 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
     }
 
     override val functionKind: FunctionKind by lazy {
+        if (modifiers.contains(Modifier.JAVA_STATIC)) {
+            FunctionKind.STATIC
+        }
```
ting-yuan, line 48: Thanks for the fix! Unfortunately getting modifiers can be slow in some cases. Can we patch the `when` case `KaSymbolLocation.CLASS` below, and make it consider `ktFunctionSymbol.isStatic` when it is a `KaNamedFunction`? I.e., something like
```
            KaSymbolLocation.CLASS -> {
                if (ktFunctionSymbol is KaNamedFunctionSymbol && ktFunctionSymbol.isStatic)
                    FunctionKind.STATIC
                else
                    FunctionKind.MEMBER
            }
```

bacecek: Yeah, looks way better, thank you!

ting-yuan (approved):

ting-yuan:
Thank you so much for the fix!

PR #2416, opened 2025-04-12
Use moduleName from compilerOptions if it exists

### LINKED ISSUE #2415: [KSP2] Wrong internal method name with custom moduleName compiler option
Steps to reproduce:
Specify `moduleName` in Android module `:app`:
```
kotlin {
    compilerOptions {
        moduleName = "foobar"
    }
}
```

In that case KSP2 will use default `app_debug` for internal methods instead of `foobar_debug` and Java compilation will fail.
KSP1 works correctly.

bacecek:
Fixes #2415


ting-yuan (approved):

ting-yuan:
Thanks for the fix!

PR #2419, opened 2025-04-16
Replace urls to JetBrains Maven repositories 

snpefk:
At JetBrains, we are moving our Maven repositories to a new hosting URL. Instead of using a direct link to the repositories, we’ve created a redirector that points to the current artifact location. This way, we won’t need to update the URL in projects using our Maven repositories in the future.


ting-yuan (approved):

ting-yuan:
Thanks for the patch!

PR #2430, opened 2025-04-29
Don't disable KSP2 native tasks if cross compilation is enabled

### LINKED ISSUE #2243: Support klib cross compilation
Kotlin 2.1.0 introduced [klib cross compilation](https://kotlinlang.org/docs/whatsnew21.html#ability-to-publish-kotlin-libraries-from-any-host) that allows building a klib for any target on any host.

KSP2 recently started [disabling tasks if they aren't on their native host](https://github.com/google/ksp/pull/2186). Can that change be rolled back to allow for klib cross compilation to work with KSP?

eygraber:
Fixes #2243

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
@@ -317,8 +317,11 @@ abstract class KspAATask @Inject constructor(
                         val konanTargetName = kotlinCompilation.target.konanTarget.name
                         cfg.konanTargetName.value(konanTargetName)
                         cfg.konanHome.value((kotlinCompileProvider.get() as KotlinNativeCompile).konanHome)
+
+                        val isKlibCrossCompilationEnabled =
+                            project.findProperty("kotlin.native.enableKlibsCrossCompilation") == "true"
                         kspAATask.onlyIf {
-                            HostManager().enabled.any {
+                            isKlibCrossCompilationEnabled || HostManager().enabled.any {
                                 it.name == konanTargetName
                             }
                         }


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -317,8 +317,11 @@ abstract class KspAATask @Inject constructor(
                         val konanTargetName = kotlinCompilation.target.konanTarget.name
                         cfg.konanTargetName.value(konanTargetName)
                         cfg.konanHome.value((kotlinCompileProvider.get() as KotlinNativeCompile).konanHome)
+
+                        val isKlibCrossCompilationEnabled =
+                            project.findProperty("kotlin.native.enableKlibsCrossCompilation") == "true"
                         kspAATask.onlyIf {
-                            HostManager().enabled.any {
+                            isKlibCrossCompilationEnabled || HostManager().enabled.any {
```
ting-yuan, line 324: I wonder if we should replace the entire logic in the `onlyIf` by `kotlinCompilation.target.publishable`. 

CC'ing @Tapchicoma,

ting-yuan (approved):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -317,8 +317,11 @@ abstract class KspAATask @Inject constructor(
                         val konanTargetName = kotlinCompilation.target.konanTarget.name
                         cfg.konanTargetName.value(konanTargetName)
                         cfg.konanHome.value((kotlinCompileProvider.get() as KotlinNativeCompile).konanHome)
+
+                        val isKlibCrossCompilationEnabled =
+                            project.findProperty("kotlin.native.enableKlibsCrossCompilation") == "true"
```
Tapchicoma, line 322: This is not the best approach. In the presence of custom CInterops cross compilation still be disabled even if this Gradle property value is `true`.

@ting-yuan could you file a new Kotlin issue to expose some API to get the cross-compilation state?

ting-yuan: Created here: https://youtrack.jetbrains.com/issue/KT-77258/Query-status-of-cross-compilation-or-compilation-task

Thank you!

PR #2459, opened 2025-05-15
Fix a variety of KSP gradle-plugin issues

liutikas:


ting-yuan (approved):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinFactories.kt
```diff
@@ -199,7 +199,7 @@ abstract class KspTaskJvm @Inject constructor(
     ),
     KspTask {
     @get:OutputDirectory
-    abstract val destination: Property<File>
+    abstract val destination: DirectoryProperty
```
Goooler, line 202: This is a breaking change, intended?  

I did a similar change in #2678 for migrating usages to `DirectoryProperty`.

Goooler: And I sent a PR for catching these API changes in #2679.

PR #2465, opened 2025-05-19
Fix origin of arguments of synthetic annotations

ting-yuan:
Origin of arguments of synthetic / default anntations should be where they were mentioned.

TODO: unify the the implementations by getting rid of Kotlin psi and Java psi.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
@@ -33,7 +33,7 @@ import org.jetbrains.kotlin.psi.KtFile
 class KSAnnotationResolvedImpl private constructor(
     private val annotationApplication: KaAnnotation,
     override val parent: KSNode?,
-    override val origin: Origin
+    override val origin: Origin,
 ) : KSAnnotation {
     companion object :
         KSObjectCache<IdKeyPair<KaAnnotation, KSNode?>, KSAnnotationResolvedImpl>() {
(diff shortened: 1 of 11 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
```diff
@@ -29,16 +28,18 @@ import org.jetbrains.kotlin.analysis.api.annotations.KaAnnotation
 import org.jetbrains.kotlin.analysis.api.impl.base.annotations.KaBaseNamedAnnotationValue
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolOrigin
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationUseSiteTarget.*
+import org.jetbrains.kotlin.psi.KtFile
 
 class KSAnnotationResolvedImpl private constructor(
     private val annotationApplication: KaAnnotation,
-    override val parent: KSNode?
+    override val parent: KSNode?,
+    override val origin: Origin
```
jsjeon, line 36: Nit: tailing comma will make next change look a bit cleaner...

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
```diff
@@ -51,7 +52,12 @@ class KSAnnotationResolvedImpl private constructor(
     }
     override val arguments: List<KSValueArgument> by lazy {
         val presentArgs = annotationApplication.arguments.map {
-            KSValueArgumentImpl.getCached(it, this, origin)
+            val argOrigin = if (origin == Origin.SYNTHETIC) {
```
jsjeon, line 55: So... the comment below should belong here?
```
// Annotations on deeply synthesized members like getter of Java annotation arguments can still be real.
```
It looks like, even though the `origin` is synthetic for annotation arguments, we're checking its containing file one more time as it could be real?

jsjeon (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
ting-yuan: Yes. The name and comment are updated to avoid the confusion.

PR #2500, opened 2025-06-20
New gradle property: `ksp.ksp2.profiling.mode`

ting-yuan:
Some profilers resolve names too late and the resolution may fail because the classloaders and processors are already garbage collected. This mode simply keeps references to processor classloaders. Therefore, it is disabled by default and should only be used when profiling.


jsjeon (approved):
Out of scope, but acronym like `cfg` and `gradleCfg` sound to me like control-flow graph, not configuration. :) I guess people usually use `config` as a short name?

PR #2503, opened 2025-06-23
Use excludedSources in AndroidPluginIntegration

hensonm-amzn:
We're seeing an issue in our internal use with KSP that's leading to a circular dependency. Similar to the integration test I've written to go with this change, we have a task that generates code but relies on the output of the KSP task before it runs. We do not need KSP to perform and work on the code generated from our task, so we used the `excludedSources` API.

The issue is that in the `AndroidPluginIntegration` class, Android Sources are being added to the source and dependencies of the KSP task without filtering against the `excludedSources` set. This means that a task added to the Android Sources via an API like [`SourceDirectories.addGeneratedSourceDirectory()`](https://developer.android.com/reference/tools/gradle-api/8.6/com/android/build/api/variant/SourceDirectories#addGeneratedSourceDirectory(org.gradle.api.tasks.TaskProvider,kotlin.Function1)) cannot currently be excluded from KSP.

This PR adds logic to that filter in `AndroidPluginIntegration` that has it respect sources added to `excludedSources`, allowing us to break the circular dependency.

---

excludedSources prevents KSP tasks from taking dependencies on any task that is added to the list in an effort to prevent circular dependencies. However, if an output is added via Android Gradle Plugin's APIs, KSP can still end up with a dependency on the task that generates the output, potentially leading to a circular dependency.

This change modifies the filter logic for adding outputs of Android projects to respect outputs added to `excludedSources`, which can be used to break the circular dependency at the cost of KSP ignoring the added source.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
@@ -76,13 +76,17 @@ object AndroidPluginIntegration {
         val kaptProvider: TaskProvider<Task>? =
             project.locateTask(kotlinCompilation.compileTaskProvider.kaptTaskName)
 
+        val kspExtension = project.extensions.getByType(KspExtension::class.java)
         val sources = kotlinCompilation.androidVariant.getSourceFolders(SourceKind.JAVA)
         kspTaskProvider.configure { task ->
             // this is workaround for KAPT generator that prevents circular dependency
             val filteredSources = Callable {
                 val destinationProperty = (kaptProvider?.get() as? KaptTask)?.destinationDir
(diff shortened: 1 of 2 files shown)


hensonm-amzn:
Hi folks, is there more process to getting this reviewed and/or incorporated? If so, please let me know and I'll get those balls rolling.

nxtstep:
Hi @hfmehmed , any chance of merging this PR in the near future? Since, this PR also fixes our issue: https://github.com/google/ksp/issues/2582 with circular dependencies.
Is there anything we can do to help expedite the process?

Thanks!

ps @hensonm-amzn nice work! :muscle: :pray:

nxtstep:
Apparently this PR has been superseded by https://github.com/google/ksp/pull/2662 - and can therefore be closed 🙏

hfmehmed:
Hi, yes, I updated the integration test to be more extensive and we merged the change. Should be released in the following ksp version

PR #2512, opened 2025-06-27
KspGradleLogger: print location in error()

ting-yuan:
Update: location is printed in front of the message, just like KSP1.

### DIFF
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/OnErrorIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/OnErrorIT.kt
@@ -20,8 +20,8 @@ class OnErrorIT(useKSP2: Boolean) {
 
         gradleRunner.withArguments("clean", "assemble").buildAndFail().let { result ->
             val errors = result.output.lines().filter { it.startsWith("e: [ksp]") }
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/OnErrorIT.kt
```diff
@@ -20,8 +20,8 @@ class OnErrorIT(useKSP2: Boolean) {
 
         gradleRunner.withArguments("clean", "assemble").buildAndFail().let { result ->
             val errors = result.output.lines().filter { it.startsWith("e: [ksp]") }
-            Assert.assertEquals("e: [ksp] Error processor: errored at 2", errors.first())
-            Assert.assertEquals("e: [ksp] NormalProcessor called error on 2", errors.last())
+            Assert.assertTrue(errors[0].endsWith("Error processor: errored at 2"))
+            Assert.assertTrue(errors[1].endsWith("NormalProcessor called error on 2"))
```
jsjeon: Out of scope, but why is it sometimes `at` or `on` a location?

ting-yuan: Done.

PR #2528, opened 2025-07-18
Kotlin version: 2.2.10-RC

mkmuir0:


gradle.properties
```diff
@@ -1,7 +1,7 @@
 # Copied from kotlinc
 org.gradle.jvmargs=-Duser.country=US -Dkotlin.daemon.jvm.options=-Xmx4096m -Dfile.encoding=UTF-8
 
-kotlinBaseVersion=2.2.20-Beta1
+kotlinBaseVersion=2.2.10-RC
```
ting-yuan: This likely needs to revert some patches that bump Kotlin 2.2.0 to 2.2.20-Beta1, because x.y.10 is generally a bug fix version of and is closer to x.y.0 in JetBrains' convention.

PR #2531, opened 2025-07-23
Update kotlin version to 2.2.10-RC

hfmehmed:


ting-yuan (approved):

ting-yuan:
Thanks for the patch! Just in case, please use **rebase and merge** rather than **create a merge commit**.

cc'ing @mkmuir0

PR #2537, opened 2025-07-30
UPDATE_KOTLIN_VERSION: 2.2.20-Beta2

ting-yuan:


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinFactories.kt
```diff
@@ -171,6 +170,7 @@ class KotlinFactories {
                             it.name == konanTargetName
                         }
                     }
+                    kspTask.separateKmpCompilation.value(false)
```
hfmehmed, line 173: If i remember correctly, this field was removed in 2.2.10-RC 
Is it added back in 2.2.20-Beta2 ?

hfmehmed (approved):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KotlinFactories.kt
ting-yuan: Yes. Unfortunately, it has to be edited back and forth because 2.2.10-* and 2.2.20-* releases are interleaved and we didn't branch out.

PR #2538, opened 2025-08-01
Add runtime warning when using KSP1

hfmehmed:
- When using AGP between 8.0.0 and 9.0.0 we will log a warning for users to move off KSP1
- After AGP 9.0.0 this will become an exception


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/agpUtils.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.gradle.api.Project
+
+fun Project.getAgpVersion(): AndroidPluginVersion? = try {
+    this.extensions
+        .findByType(com.android.build.api.variant.AndroidComponentsExtension::class.java)
+        ?.pluginVersion
+} catch (e: Exception) {
+    // Perhaps a version of AGP before pluginVersion API was added.
+    null
+} catch (e: NoClassDefFoundError) {
```
ting-yuan, line 10: This statement is unreachable per https://kotlinlang.org/docs/exceptions.html#handle-exceptions-using-try-catch-blocks. Please reorder the catches so that more specific ones come before less specific ones.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -233,6 +233,16 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
         val kspVersion = ApiVersion.parse(KSP_KOTLIN_BASE_VERSION)!!
         val kotlinVersion = ApiVersion.parse(project.getKotlinPluginVersion())!!
 
+        val useKsp2 = project.extensions.getByType(KspExtension::class.java).useKsp2.get()
+
+        if (useKsp2.not()) {
+            project.logger.warn(
+                "We noticed you are using KSP1 which is deprecated and support for it will be removed soon - " +
+                    "please migrate to KSP2 as soon as possible. " +
+                    "KSP1 will no longer be compatible with AGP 9.0.0 (and above) and KGP 2.3.0 (and above)"
```
ting-yuan: nit: replace KGP with Kotlin for developers who may not know much about KGP.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/agpUtils.kt
hfmehmed: Done

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
hfmehmed: Thanks, done

Issue #2553, opened 2025-08-14
Consider not printing the version warning if KSP2 is used?


martinbonnin:
My understanding is that KSP2 depends on its own version of the Kotlin compiler and therefore doesn't need to be tightly coupled to the KGP version?

If that's the case, can we remove the `Please upgrade ksp or downgrade kotlin-gradle-plugin` warning?

PR #2554, opened 2025-08-14
Remove version warning if KSP2 is used

martinbonnin:
Closes https://github.com/google/ksp/issues/2553


hfmehmed:
I think there are a couple of test (VersionIT) - could you update those in the cases where ksp2 is in use

martinbonnin:
@hfmehmed right, good catch. Tests updated.

PR #2574, opened 2025-08-28
Add integration test for AGP 9.0

hfmehmed:
Gradle 9.0 requires all task outputs to explicitly define which taskProvider it was created
Hence the need for `project.fileTree(resourcesOutputDir).builtBy(kspTaskProvider)`


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -172,6 +172,7 @@ object AndroidPluginIntegration {
             val kspJavaOutput = project.fileTree(javaOutputDir).builtBy(kspTaskProvider)
             val kspKotlinOutput = project.fileTree(kotlinOutputDir).builtBy(kspTaskProvider)
             val kspClassOutput = project.fileTree(classOutputDir).builtBy(kspTaskProvider)
+            project.fileTree(resourcesOutputDir).builtBy(kspTaskProvider)
```
ting-yuan: I may be wrong, but isn't this a no-op, if the returned object of `builtBy` isn't consumed by anything?

hfmehmed: Without that line, i am seeing the following Gradle 9.0 error:

org.gradle.internal.execution.WorkValidationException: A problem was found with the configuration of task ':workload:kspDebugKotlin' (type 'KspAATask').
  - Gradle detected a problem with the following location: '/tmp/junit1389268230539330646/workload/build/generated/ksp/debug/resources'.
    
    Reason: Task ':workload:processDebugJavaRes' uses this output of task ':workload:kspDebugKotlin' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':workload:kspDebugKotlin' as an input of ':workload:processDebugJavaRes'.
      2. Declare an explicit dependency on ':workload:kspDebugKotlin' from ':workload:processDebugJavaRes' using Task#dependsOn.
      3. Declare an explicit dependency on ':workload:kspDebugKotlin' from ':workload:processDebugJavaRes' using Task#mustRunAfter.

hfmehmed: Removed this for now while i investigate a failure after correcting this

ting-yuan (approved):

PR #2575, opened 2025-08-29
[KSP1] Fix missing KSP resources output in RuntimeLibraryClassesJar by adding ksp task dependency

larryxiao625:
# Issue
On KSP1, KSP resource output are registered to postJavacGeneratedBytecode outputs, these files will be placed in the MultipleArtifact.ALL_CLASSES_DIRS artifact in AGP 742 version, and it will have a filter by ArtifactTypeUtil
image
image
This filter will filter by isDirectory, so if the directory is not exists when clean build, it will return false and filter the KSP resource output dir.

Meanwhile, the ALL_CLASSES_DIRS artifact is consumed by the bundleLibRuntimeJar* task during library bundling.
If there is no explicit dependency between bundleLibRuntime and the KSP resource generation tasks, this can lead to a race condition where the generated outputs are not yet produced when bundleLibRuntimeJar* task runs, resulting in missing artifacts in the final runtime_library_classesjar/*/classes.jar AAR.

# Root Cause
- postJavacGeneratedBytecode output → contributes to ALL_CLASSES_DIRS artifact.

- bundleLibRuntimeJar* → uses ALL_CLASSES_DIRS artifact without depending on the KSP resource tasks.

- Without this dependency, bundleLibRuntimeJar* may execute before KSP tasks finish, causing missing ksp resources in the final classes.jar.

# Fix

Add an explicit task dependency for resource output dir so that bundleLibRuntimeJar*. This guarantees resource output artifacts are available before packaging.


hfmehmed:
Thanks for your contribution.

PR #2579, opened 2025-09-01
Fix Android integration for Gradle 9.0 and AGP with Built in Kotlin

hfmehmed:
Cherry-picked from main branch

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
@@ -27,6 +27,7 @@ import org.gradle.api.file.FileCollection
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.TaskProvider
 import org.jetbrains.kotlin.gradle.internal.KaptTask
+import org.jetbrains.kotlin.gradle.plugin.KotlinBaseApiPlugin
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinJvmAndroidCompilation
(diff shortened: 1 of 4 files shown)


liutikas:
What's missing to land this?

hfmehmed:
Nothing, this is old PR. Everything here landed and it released just now as part of 2.2.20-2.0.3 release

pablobaxter:
Will this be released for the previous versions of Kotlin as well?

hfmehmed:
We didnt plan to do that. Is there anything blocking from upgrading to KSP 2.0.3?

pablobaxter:
AGP 9 has a base Kotlin vresion of 2.2.10, and I'm actually trying to prepare for an easy swap to AGP 9 by upgrading all the necessary dependencies (as they are released), so I'm actually on Kotlin 2.1.20 still.

liutikas:
It would be nice to backport it to at least a few past releases of Kotlin

hfmehmed:
We will discuss releasing KSP 2.0.3 version with older Kotlin versions. But theoretically nothing stops you from updating to say 2.2.20-2.0.3 (even if you are using different base version of Kotlin) Or is there any concerns with that? 

Note that there are some regressions that went into 2.2.20-2.0.3 so if you see problems please try 2.2.20-2.0.4-SNAPSHOT from maven("https://central.sonatype.com/repository/maven-snapshots/") and report back.

pablobaxter:
This mainly stops us from being able to test AGP 9 and KSP 2.0.3 in our build without also updating to Kotlin 2.2.20 before we have completed validating that version. So theoretically, nothing stops us from updating, but it'll be a longer process for validating all the new libraries, and doesn't allow us easily to test them independently.

PR #2586, opened 2025-09-05
Configurable logging + docs + contributing task name fix

vaslabs:
## Motivation

The ksp cmdline tools like `KspJvmMain` are a great tool for integrating with KSP as a proof of concept. During such experimentation, I've found myself changing the ksp sources and rebuilding, first because of the logging level that is hardcoded. 

The idea behind this PR is to make logging configurable (while still having the warn as a default.

## Implementation

The implementation simply accepts a jvm parameter `ksp.logging` (which of course you may suggest more fitting names) in order to configure the logging.

Passing the logging level in the configuration instead does not play well with the current wiring, as `KSPJvmConfig`, `KSPJsConfig`  and `KSPCommonConfig` which are subclasses of `KSPConfig` would either result in `KspConfig` having a redundant logging (as `KotlinSymbolProcessing` would end up getting both the logging configuration and the `KspConfig` which would create confusion) or an additional indirection, like a KSPCliConfig abstract class extending from `KspConfig` and it didn't seem worth doing so.

So in the end I decided passing a jvm parameter is simpler. 

## Other fixes

I've noticed the `./gradlew :api:apiCheck` should be `./gradlew :api:checkApi` instead

I hope you find this useful and I'm open to suggestions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ This project follows
 
 ## API verification
 
-For changes that involves API changes(new API, API signature change), please also update [api.base](./api/api.base) file. You can monitor api change with `./gradlew :api:apiCheck`, and`./gradlew :api:updateApi` to generate new api signature.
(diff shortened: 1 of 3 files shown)


hfmehmed:
Thanks for your contribution :)

PR #2593, opened 2025-09-11
Upgrade to Kotlin 2.2.20

mkmuir0:


mkmuir0:
> To be pedantic, this is actually an upgrade 😃

ah you're right, thanks for catching!

PR #2602, opened 2025-09-12
Fallback to AA for unexpected Java psi type

### LINKED ISSUE #2598: ClassCastException when building with 2.2.20-2.0.3
Works with 2.2.20-2.0.2 so recent regression.

```
> A failure occurred while executing com.google.devtools.ksp.gradle.KspAAWorkerAction
   > class ksp.com.intellij.psi.impl.source.PsiRecordComponentImpl cannot be cast to class ksp.com.intellij.psi.PsiJvmModifiersOwner (ksp.com.intellij.psi.impl.source.PsiRecordComponentImpl and ksp.com.intellij.psi.PsiJvmModifiersOwner are in unnamed module of loader java.net.URLClassLoader @2f0fdacb)

```

[full-stacktrace.txt](https://github.com/user-attachments/files/22294554/full-stacktrace.txt)

ting-yuan:
when getting annotations. Avoid the dangerous cast.

Still trying to reproduce for test case, but the fix straightforward enough from the stack trace. Let's unblock 2.0.4 first.

Fixes #2598


sureshg:
@ting-yuan  Is `2.0.4` planned to release soon with this `ClassCastException` fix ?

hfmehmed:
There are a few more bugs found in 2.0.3 so we will try to fix those before releasing 2.0.4 but yes, it should come soon

PR #2609, opened 2025-09-20
Revert the way sources are queried for ksp task

hfmehmed:
The new Variant API was causing various issues. Due to its lazy nature it was not possible to filter out kapt task.

This change also adds some more tests for future resilience.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
@@ -63,9 +63,7 @@ object AndroidPluginIntegration {
     }
 
     fun getCompilationSourceSets(kotlinCompilation: KotlinJvmAndroidCompilation): List<String> {
-        return kotlinCompilation.androidVariant
-            .sourceSets
-            .map { it.name }
+        return kotlinCompilation.androidVariant?.sourceSets?.map { it.name } ?: emptyList()
     }
 
     /**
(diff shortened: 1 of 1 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -81,36 +81,32 @@ object AndroidPluginIntegration {
...
-                }
-                when (task) {
-                    is KspTaskJvm -> { task.source(filteredSources) }
-                    is KspAATask -> { task.kspConfig.javaSourceRoots.from(filteredSources) }
-                    else -> Unit
-                }
+        val androidVariant = kotlinCompilation.androidVariant
+        if (androidVariant == null) {
+            throw RuntimeException(
+                "KSP is not compatible with built-in Kotlin. " +
```
ting-yuan: I guess what you mean by "built-in Kotlin" is AGP's built-in Kotlin. If so, should this be "KSP is not compatible with Android Gradle Plugin's built-in Kotlin. Please disable it by ..."?

Also, does a developer need to apply the kotlin plugin when `android.builtInKotlin=false` is set?

hfmehmed: Yes, if they turn off builtInKotlin, they neeed to apply kotlin-android plugin.

hfmehmed: Updated the code to reflect that

PR #2624, opened 2025-10-02
Re-enable support for AGP + built in kotlin in KSP

hfmehmed:
Experimenting with reading AGP static and generated sources using sources.kotlin but regirested KSP generated sources to java sources to avoid circular dependency

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
@@ -76,36 +76,47 @@ object AndroidPluginIntegration {
         kspTaskProvider: TaskProvider<*>,
         androidComponent: Component?
     ) {
-        val kaptProvider: TaskProvider<Task>? =
-            project.locateTask(kotlinCompilation.compileTaskProvider.kaptTaskName)
+        if (project.isAgpBuiltInKotlinUsed() && androidComponent != null) {
+            val sources = androidComponent.sources.kotlin?.all
 
-        val androidVariant = kotlinCompilation.androidVariant
-        if (androidVariant == null) {
-            throw RuntimeException(
-                "KSP is not compatible with Android Gradle Plugin's built-in Kotlin. " +
-                    "Please disable by adding android.builtInKotlin=false to gradle.properties " +
-                    "and apply kotlin(\"android\") plugin"
-            )
-        }
-        val sources = androidVariant.getSourceFolders(SourceKind.JAVA)
+            kspTaskProvider.configure { task ->
+                when (task) {
+                    is KspTaskJvm -> {
+                        task.source(Callable { sources })
+                    }
 
-        kspTaskProvider.configure { task ->
-            // this is workaround for KAPT generator that prevents circular dependency
-            val filteredSources = Callable {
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -76,36 +76,47 @@ object AndroidPluginIntegration {
...
-            val filteredSources = Callable {
-                val destinationProperty = (kaptProvider?.get() as? KaptTask)?.destinationDir
-                val dir = destinationProperty?.get()?.asFile
-                sources.filter { dir?.isParentOf(it.dir) != true }
+        if (project.isAgpBuiltInKotlinUsed() && androidComponent != null) {
+            val sources = androidComponent.sources.kotlin?.all
+
+            kspTaskProvider.configure { task ->
+                when (task) {
+                    is KspTaskJvm -> {
```
ting-yuan, line 84: Is / should this ever reachable?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -76,36 +76,47 @@ object AndroidPluginIntegration {
...
-        val sources = androidVariant.getSourceFolders(SourceKind.JAVA)
-
-        kspTaskProvider.configure { task ->
-            // this is workaround for KAPT generator that prevents circular dependency
-            val filteredSources = Callable {
-                val destinationProperty = (kaptProvider?.get() as? KaptTask)?.destinationDir
-                val dir = destinationProperty?.get()?.asFile
-                sources.filter { dir?.isParentOf(it.dir) != true }
+        if (project.isAgpBuiltInKotlinUsed() && androidComponent != null) {
+            val sources = androidComponent.sources.kotlin?.all
```
ting-yuan, line 80: IIRC the first attempt was `androidComponent.sources.java.static`. Are they the same set of files? In other words, would this approach miss any desired files?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
hfmehmed: I suppose technically you can use builtin kotlin and ksp1 
but maybe we should restrict that it never reaches it since ksp1 is deprecated and we dont thoroughly test it

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
hfmehmed: `sources.kotlin` contains everything that java.static has and more stuff that are generated by AGP such as databinding and build config. 

The benefit of using `sources.kotlin.all` is also that it doesnt contain the ksp output (hence not leading to circular dependency)

ting-yuan (approved):

hfmehmed:
I found another use where this does not work. So holding off of merging this until we untangle that

Goooler:
Looks handled by #2674.

Goooler:
Looks this could be closed for now.

PR #2626, opened 2025-10-06
Replace KaFirStopWorldCacheCleaner with no-op implementation

fsladkey:
Replace KaFirStopWorldCacheCleaner with no-op implementation

This cache cleaner leads to flakey build failures in AndroidX. KSP has already updated to a Kotlin AA version which no longer registered the cache cleaner by default for standalone but it is still registered for all other usages. This CL replaces it with a no-op implementation.


fsladkey:
Not sure how testable this change is, but happy to add a test if someone can point me in the right direction.

hfmehmed:
LGTM. @ting-yuan can you review too

jsjeon:
Here is another way: https://cs.android.com/android-studio/platform/tools/base/+/1c13ba54892389f0706e7308f6e942039684680d

fsladkey:
Thanks jsjeon!

When I changed the LoadingState so that the correct registry value would be picked up it caused some other failures which I wasn't familiar with the internals of enough to confidently fix. Happy to work with someone more familiar with the codebase to get that version working if it's preferable.

ting-yuan:
Both of the ways work. Let's merge this for 2.0.4 and revisit it later if necessary.

PR #2639, opened 2025-10-10
UPDATE_KOTLIN_VERSION: 2.3.0-Beta1

hfmehmed:
Drops KSP1 testing


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspExtension.kt
```diff
@@ -17,23 +17,14 @@
...
 import javax.inject.Inject
 
 abstract class KspExtension @Inject constructor(project: Project) {
-    /**
-     * Enables or disables KSP 2, defaults to the `ksp.useKSP2` gradle property or `true` if that's not set.
-     *
-     * This API is temporary and will be removed once KSP1 is removed.
-     */
-    @KspExperimental
-    abstract val useKsp2: Property<Boolean>
```
ting-yuan, line 36: Removing this would break projects using it, even if they used `useKsp2=true`. Can we leave it here and give a runtime error if it is set to false?

and also the ksp.useKSP2 Gradle property.

hfmehmed: added it back

PR #2674, opened 2025-10-28
Add support for AGP 9.0 and Built in kotlin

hfmehmed:
* Updates AGP version against which we compile so that we can use new APIs added especially for ksp
* Updates some tests to opt out of built in kotlin and new dsl - because they may apply plugins that are still not compatible with AGP 9.0

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
@@ -79,19 +79,20 @@ object AndroidPluginIntegration {
         kspTaskProvider: TaskProvider<KspAATask>,
         androidComponent: Component?
     ) {
-        if (project.isAgpBuiltInKotlinUsed()) {
-            if (androidComponent != null && project.canUseInternalKspApis()) {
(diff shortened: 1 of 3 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -76,32 +79,42 @@ object AndroidPluginIntegration {
...
+            if (androidComponent != null && project.canUseInternalKspApis()) {
+                val sources =
+                    (androidComponent.sources.java!! as FlatSourceDirectoriesForJavaImpl).allButKspAndKaptGenerators()
+
+                kspTaskProvider.configure { task ->
+                    task.kspConfig.javaSourceRoots.from(sources)
                 }
+            } else {
+                throw RuntimeException(
+                    "KSP is not compatible with Android Gradle Plugin's built-in Kotlin. Please disable " +
```
cmwarrington: Maybe

KSP is not compatible with Android Gradle Plugin's built-in Kotlin prior to AGP 9.0.0-alpha12 ?

hfmehmed: Updated the messaging

PR #2689, opened 2025-11-03
Avoid renaming Nullable/NonNull/NotNull annotations

ting-yuan:
They are checked explicitly in FIR, when determining the nullability of types.

### DIFF
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/NullabilityAnnotationsIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/NullabilityAnnotationsIT.kt
@@ -1,0 +1,49 @@
+package com.google.devtools.ksp.test
+
+import com.google.devtools.ksp.test.fixtures.TemporaryTestProject
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.Assert
+import org.junit.Assume
+import org.junit.Rule
+import org.junit.Test
+
+class NullabilityAnnotationsIT() {
+    @Rule
+    @JvmField
+    val project: TemporaryTestProject = TemporaryTestProject("nullability-annotations")
+
+    private fun GradleRunner.buildAndCheck(vararg args: String, extraCheck: (BuildResult) -> Unit = {}) =
+        buildAndCheckOutcome(*args, outcome = TaskOutcome.SUCCESS, extraCheck = extraCheck)
+
+    private fun GradleRunner.buildAndCheckOutcome(
+        vararg args: String,
+        outcome: TaskOutcome,
+        extraCheck: (BuildResult) -> Unit = {}
+    ) {
+        val result = this.withArguments(*args).build()
+
+        Assert.assertEquals(outcome, result.task(":workload:build")?.outcome)
+
+        extraCheck(result)
+    }
+
+    @Test
+    fun testNullableAnnotations() {
+        // FIXME: `clean` fails to delete files on windows.
+        Assume.assumeFalse(System.getProperty("os.name").startsWith("Windows", ignoreCase = true))
+        val gradleRunner = GradleRunner.create().withProjectDir(project.root)
+        gradleRunner.buildAndCheck("clean", "build") { result ->
+            Assert.assertTrue(result.output.contains("w: [ksp] [Nullability check] s0: (String..String?)\n"))
+            Assert.assertTrue(result.output.contains("w: [ksp] [Nullability check] s1: String?\n"))
+            Assert.assertTrue(result.output.contains("w: [ksp] [Nullability check] s2: String\n"))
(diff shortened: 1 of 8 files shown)


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/PlaygroundIT.kt
```diff
@@ -49,6 +49,24 @@ class PlaygroundIT() {
...
 
+    @Test
+    fun testNullableAnnotations() {
+        // FIXME: `clean` fails to delete files on windows.
+        Assume.assumeFalse(System.getProperty("os.name").startsWith("Windows", ignoreCase = true))
+        val gradleRunner = GradleRunner.create().withProjectDir(project.root)
+        gradleRunner.buildAndCheck("clean", "build") { result ->
+            Assert.assertTrue(result.output.contains("w: [ksp] [Nullability check] s0: (String..String?)\n"))
+            Assert.assertTrue(result.output.contains("w: [ksp] [Nullability check] s1: String?\n"))
+            Assert.assertTrue(result.output.contains("w: [ksp] [Nullability check] s2: String\n"))
```
jsjeon: Whoa, so `@javax.annotation.*` helps determining the nullity, but not annotated at the end?

integration-tests/src/test/resources/playground/workload/build.gradle.kts
```diff
@@ -18,6 +18,10 @@ repositories {
 dependencies {
     implementation(kotlin("stdlib"))
     implementation(project(":test-processor"))
+    implementation("javax.annotation:javax.annotation-api:1.3.2")
+    implementation("com.google.code.findbugs:jsr305:3.0.2")
+    implementation("org.checkerframework:checker-qual:1.8.1")
+    implementation("org.jetbrains:annotations:26.0.2-1")
```
Goooler:
```suggestion
    compileOnly("javax.annotation:javax.annotation-api:1.3.2")
    compileOnly("com.google.code.findbugs:jsr305:3.0.2")
    compileOnly("org.checkerframework:checker-qual:1.8.1")
    compileOnly("org.jetbrains:annotations:26.0.2-1")
```

I noticed you excluded classes from these dependencies, seems we can replace the exclusion step with `compileOnly`.

build.gradle.kts
```diff
@@ -26,9 +26,9 @@ plugins {
     kotlin("jvm")
     id("io.github.gradle-nexus.publish-plugin") version "2.0.0"
     // Adding plugins used in multiple places to the classpath for centralized version control
-    id("com.gradleup.shadow") version "8.3.9" apply false
+    id("com.gradleup.shadow") version "9.2.2" apply false
```
Goooler: I suggest addressing the plugin update in a separate PR.

symbol-processing-aa-embeddable/build.gradle.kts
```diff
@@ -132,15 +129,31 @@ tasks.withType(ShadowJar::class.java).configureEach {
...
+            excludeEfficiently("org.checkerframework.checker.nullness.qual.Nullable")
         }
     }
     mergeServiceFiles()
     exclude("META-INF/compiler.version")
     exclude("META-INF/*.kotlin_module")
+    exclude("javax/**")
 
-    this.transform(AAServiceTransformer())
+    transform<AAServiceTransformer>()
```
Goooler: Seems missing `duplicatesStrategy`. I migrated these configurations, but haven't tested them. I can send a PR later. https://github.com/Goooler/ksp/commit/a85f89e77922ced0b1633f81e13ed42e41de137b

integration-tests/src/test/resources/playground/workload/src/main/java/com/example/BClass.java
```diff
@@ -11,4 +11,38 @@ public HELLO helloFun(){
...
+  String s3();
+
+  @org.jetbrains.annotations.NotNull
+  String s4();
+
+  @org.checkerframework.checker.nullness.qual.Nullable
+  String s5();
+
+  @org.checkerframework.checker.nullness.qual.NonNull
+  String s6();
```
Goooler: The full list could be refed to https://github.com/JetBrains/kotlin/blob/d0763373d5da3173095a86ede9e9ec8ef64a7310/core/compiler.common.jvm/src/org/jetbrains/kotlin/load/java/JvmAnnotationNames.kt#L64-L122.

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/PlaygroundIT.kt
ting-yuan: That is another issue: either KaType.toString() didn't do it consistently, or KaType.annotations() has a bug.

integration-tests/src/test/resources/playground/workload/src/main/java/com/example/BClass.java
liutikas: It looks like we missed org.jspecify.annotations.NonNull and caused a regression.

PR #2694, opened 2025-11-05
Fix regression related to AGP 8.7.0 and below

hfmehmed:
This fixes a regression introduced in #2656 where we check if target is of a specific type which is available since AGP 8.8.0 

This leads to runtime failure for project using 8.8.0 and below.
Also adding more tests with older AGP versions.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
@@ -1,12 +1,12 @@
 package com.google.devtools.ksp.gradle
 
-import com.android.build.api.dsl.KotlinMultiplatformAndroidLibraryTarget
 import com.google.devtools.ksp.gradle.AndroidPluginIntegration.useLegacyVariantApi
 import org.gradle.api.InvalidUserCodeException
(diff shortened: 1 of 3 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspConfigurations.kt
```diff
@@ -166,6 +164,22 @@ class KspConfigurations(private val project: Project) {
         }
     }
 
+    // check if this target is the old implementation of android kmp target (using com.android.library)
+    // org.jetbrains.kotlin.gradle.plugin.mpp.KotlinAndroidTarget
+    private fun isOldKmpAndroidTarget(target: KotlinTarget): Boolean {
+        val isKotlinAndroidTargetClass = try {
+            target is KotlinAndroidTarget
```
ting-yuan, line 171: Instead of `try`, couldn we check the KGP version explicitly?

Also, `KotlinAndroidTarget` seems be part of `kotlin-gradle-plugin`, not `kotlin-gradle-plugin-api`. Would that be vulnerable to compatibility to future KGP versions?

Would checking AGP version explicitly before referencing `KotlinMultiplatformAndroidLibraryTarget` a better alternative?

PR #2713, opened 2025-11-24
Register ksp generated sources using new KGP APIs

hfmehmed:
Support Gradle project isolation


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/kgpUtils.kt
cmwarrington, line 1: projectIsolationEnabled - this name sounds like it's the gradle level concept, and not our own flag - maybe enableProjectIsolationCompatibleCodepath?

hfmehmed: updated the naming

consp1racy:
Does this support generated Java sources?

PR #2718, opened 2025-11-29
Fix ClassNotFoundException in case there are no SymbolProcessorProviders.

kefirfromperm:
In case when there are no any SymbolProcessorProvider implementations in the ksp classpath, I catch an exception.

`java.lang.ClassNotFoundException: com.google.devtools.ksp.impl.KotlinSymbolProcessing$ExitCode`

There are no reasons to use this class to build an error message. So I fixed it, replaced with the class `com.google.devtools.ksp.processing.ExitCode`.

Also, I improved the error message to clarify what to do in case there a no symbol processor providers. And removed an unnecessary dependency `compileOnly(project(":kotlin-analysis-api"))`. It is unnecessary anymore.


kefirfromperm:
I've signed CLA, please rerun the workflow.

PR #2731, opened 2025-12-15
[AGP Integration] Pass kotlin sources from variant api 

hfmehmed:
They are no longer being passed to ksp task via allKotlinSourceSetsObservable when built in kotlin is enabled and android.disallowKotlinSourceSets is set to true (default in AGP 9.0.0-beta05)

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
@@ -81,11 +81,14 @@ object AndroidPluginIntegration {
     ) {
         if (androidComponent != null && project.isAgpBuiltInKotlinUsed()) {
             if (project.canUseInternalKspApis()) {
-                val sources =
+                val javaSources =
                     (androidComponent.sources.java as? FlatSourceDirectoriesForJavaImpl)?.allButKspAndKaptGenerators()
 
+                val kotlinSources = androidComponent.sources.kotlin?.static
+
                 kspTaskProvider.configure { task ->
-                    task.kspConfig.javaSourceRoots.from(sources)
+                    task.kspConfig.javaSourceRoots.from(javaSources)
+                    task.kspConfig.javaSourceRoots.from(kotlinSources)
                 }
             } else {
                 throw RuntimeException(
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/kgpUtils.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/kgpUtils.kt
@@ -8,6 +8,7 @@ import org.jetbrains.kotlin.gradle.plugin.getKotlinPluginVersion
 import org.jetbrains.kotlin.gradle.utils.ObservableSet
 import org.jetbrains.kotlin.tooling.core.KotlinToolingVersion
 
+// NOTE: for AGP with built in kotlin enabled and android.disallowKotlinSourceSets=true, this returns empty set
 internal val KotlinCompilation<*>.allKotlinSourceSetsObservable
     get() = this.allKotlinSourceSets as ObservableSet<KotlinSourceSet>
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@ org.gradle.jvmargs=-Duser.country=US -Dkotlin.daemon.jvm.options=-Xmx4096m -Dfil
 org.gradle.parallel=true
 
 kotlinBaseVersion=2.3.0-RC
-agpBaseVersion=9.0.0-beta01
+agpBaseVersion=9.0.0-beta05
 intellijVersion=241.19416.19
 junitVersion=4.13.1
 junit5Version=5.8.2
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/AndroidBuiltInKotlinIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/AndroidBuiltInKotlinIT.kt
@@ -75,7 +75,7 @@ class AndroidBuiltInKotlinIT {
     fun testPlaygroundAndroidWithBuiltInKotlinAGP90AboveAlpha14() {
         val gradleRunner = GradleRunner.create().withProjectDir(project.root).withGradleVersion("9.1.0")
 
-        File(project.root, "gradle.properties").appendText("\nagpVersion=9.0.0-alpha14")
+        File(project.root, "gradle.properties").appendText("\nagpVersion=9.0.0-beta05")
 
         gradleRunner.withArguments(
             "clean", "build", "minifyReleaseWithR8", "--configuration-cache", "--info", "--stacktrace"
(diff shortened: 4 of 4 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -81,11 +81,14 @@ object AndroidPluginIntegration {
...
-                val sources =
+                val javaSources =
                     (androidComponent.sources.java as? FlatSourceDirectoriesForJavaImpl)?.allButKspAndKaptGenerators()
 
+                val kotlinSources = androidComponent.sources.kotlin?.static
+
                 kspTaskProvider.configure { task ->
-                    task.kspConfig.javaSourceRoots.from(sources)
+                    task.kspConfig.javaSourceRoots.from(javaSources)
+                    task.kspConfig.javaSourceRoots.from(kotlinSources)
```
hungvietnguyen, line 91: nit: Should we rename it to `javaAndKotlinSourceRoots`?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -210,7 +211,8 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kotlinCompileProvider.configure { it.source(*generatedSources) }
         } else {
             val skipAddingSources = kotlinCompilation is KotlinJvmAndroidCompilation &&
-                project.isAgpBuiltInKotlinUsed() && project.canUseInternalKspApis()
+                project.isAgpBuiltInKotlinUsed() && project.canUseInternalKspApis() &&
+                project.allowKotlinSourceSets().not()
```
hungvietnguyen: `allowKotlinSourceSets` is a temporary AGP flag, it will be true by default from now on, and even if it's false it shouldn't change how KSP configures things. So perhaps remove this check here?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -81,11 +81,14 @@ object AndroidPluginIntegration {
     ) {
         if (androidComponent != null && project.isAgpBuiltInKotlinUsed()) {
             if (project.canUseInternalKspApis()) {
-                val sources =
+                val javaSources =
                     (androidComponent.sources.java as? FlatSourceDirectoriesForJavaImpl)?.allButKspAndKaptGenerators()
 
+                val kotlinSources = androidComponent.sources.kotlin?.static
```
hungvietnguyen: should we use `kotlin?.allButKspAndKaptGenerators`?

hfmehmed: kotlin doesnt have allButKspAndKaptGenerators
we probably would need a new API for that

hfmehmed: we could `.all` however

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
hfmehmed: i would rather avoid renaming this in case it leads to some third party breakages. we can. consider it in a major version update

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
hfmehmed: i had it as a safeguard for when the property it set to "false" explicitely by the user. but removed it now to keep things simple

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -81,11 +81,14 @@ object AndroidPluginIntegration {
     ) {
         if (androidComponent != null && project.isAgpBuiltInKotlinUsed()) {
             if (project.canUseInternalKspApis()) {
-                val sources =
+                val javaSources =
                     (androidComponent.sources.java as? FlatSourceDirectoriesForJavaImpl)?.allButKspAndKaptGenerators()
 
+                val kotlinSources = androidComponent.sources.kotlin?.all
```
hungvietnguyen: would `kotlin.all` cause cyclic task dependencies? (the same issue that we tried to avoid by using `java.allButKspAndKaptGenerators`)

hfmehmed: we dont register the output of ksp into the kotlin sourcset so it wont cause circular dependency

hfmehmed: reverted to static, as `all` may be a risky choice at the moment. We can always re-evaluate this in future

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
hfmehmed: decided to keep it static

ting-yuan (approved):
Although I share the same concerns around circular dependencies with @hungvietnguyen, `.static` seems to be safe and better than nothing.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
hungvietnguyen, line 87: nit: Add a TODO(bug) to consider `kotlin.all` or `kotlin.allExceptKspOutputs` later so we don't forget?

(I believe we need `kotlin.allExceptKspOutputs`, but if you think `kotlin.all` is safe then perhaps we can confirm that with a test?)

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -81,11 +81,14 @@ object AndroidPluginIntegration {
     ) {
         if (androidComponent != null && project.isAgpBuiltInKotlinUsed()) {
             if (project.canUseInternalKspApis()) {
-                val sources =
+                val javaSources =
                     (androidComponent.sources.java as? FlatSourceDirectoriesForJavaImpl)?.allButKspAndKaptGenerators()
```
hungvietnguyen, line 85: just curious: KSP doesn't generate Java sources, so I'm not sure why this API is called `FlatSourceDirectoriesForJava.allButKspAndKaptGenerators` rather than just `allButKaptGenerators`

hfmehmed: ksp does generate java sources actually

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
hungvietnguyen: Btw, it seems that Kapt can generate Kotlin sources too, so perhaps it should be `kotlin.allExceptKspAndKaptOutputs`.

PR #2734, opened 2025-12-18
Bump AA to 2.3.20-dev-7064

ting-yuan:


ting-yuan:
AA now depends on `org.jetbrains.intellij.deps.kotlinx:kotlinx-coroutines-core-jvm:1.8.0-intellij-14` instead of `org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.6.0`. We may need to bundle it.

PR #2740, opened 2025-12-24
Add minimum supported AGP version (8.3.0)

hfmehmed:
This is a new policy of support. This target will keep moving as AGP releases new versions.

It will simplify the KSP code that strictly depends on the runtime version of AGP.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/agpUtils.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/agpUtils.kt
@@ -17,6 +17,15 @@ fun Project.getAgpVersion(): AndroidPluginVersion? = try {
 
 fun Project.isAgpBuiltInKotlinUsed() = isKotlinBaseApiPluginApplied() && isKotlinAndroidPluginApplied().not()
 
+fun Project.checkMinimumAgpVersion() {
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/agpUtils.kt
```diff
@@ -42,3 +42,11 @@ fun Project.canUseInternalKspApis(): Boolean {
...
     return agpVersion >= AndroidPluginVersion(9, 0, 0).alpha(14)
 }
+
+/**
+ * Defines the minimum supported Android Gradle Plugin (AGP) version.
+ *
+ * KSP aims to support AGP versions released approximately within the last year
+ * from the current KSP release date.
+ */
+val MINIMUM_SUPPORTED_AGP_VERSION = AndroidPluginVersion(8, 3, 0)
```
ting-yuan, line 61: If this is the reason for the CI error, where it is referenced and evaluated too early, we may wrap it in a function.

hfmehmed: I moved the code around a bit. Hopefully now jvm shouldnt try to load it prematurely before we know AGP is applied

ting-yuan (approved):

PR #2747, opened 2026-01-08
Fix detection of android base plugin usage

hfmehmed:
When applying settings plugins (that come from composite gradle build), some classloader change leads to "com.android.base" plugin not being detected as applied by Gradle, leading to problems. 

When using the plugins.withType(Class) API this works, however now we face a Class not found error if the AGP plugin is not applied. So everything is wrapped in a try catch block.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
@@ -24,6 +24,7 @@ import com.android.build.api.variant.impl.DirectoryEntry
 import com.android.build.api.variant.impl.FlatSourceDirectoriesForJavaImpl
 import com.android.build.api.variant.impl.FlatSourceDirectoriesImpl
 import com.android.build.gradle.BaseExtension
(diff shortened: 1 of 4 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -48,9 +49,13 @@ import java.util.concurrent.Callable
...
-        project.pluginManager.withPlugin("com.android.base") {
-            // for android modules, we need a configuration per source set
-            decorateAndroidExtension(project, onSourceSet)
+        try {
+            project.plugins.withType(AndroidBasePlugin::class.java).configureEach {
+                // for android modules, we need a configuration per source set
+                decorateAndroidExtension(project, onSourceSet)
+            }
+        } catch (e: Throwable) {
+            // Android plugin not found, ignore
```
ting-yuan, line 58: It is too broad and may bite us in the future. Is it feasible to catch the specific exception type and make sure it is because of AndroidBasePlugin?

hfmehmed:
Actually i am still unconvinced we should do this. Lets wait until i get more clarity what is happening here

PR #2748, opened 2026-01-09
Remove dependency on KotlinCompile.libraries

### LINKED ISSUE #2743: Circular Dependency Between KAPT and KSP with AGP 9 and Legacy KAPT
Using the [built-in Kotlin](https://developer.android.com/build/migrate-to-built-in-kotlin) and `legacy-kapt` in AGP 9 emit this error:

```
* Exception is:
org.gradle.api.CircularReferenceException: Circular dependency between the following tasks:
:midiroll-screen:kaptDebugKotlin
+--- :midiroll-screen:kaptGenerateStubsDebugKotlin
|  \--- :midiroll-screen:kspDebugKotlin
|     \--- :midiroll-screen:kaptDebugKotlin (*)
\--- :midiroll-screen:kspDebugKotlin (*)

(*) - details omitted (listed previously)
```

Might be related with https://github.com/google/ksp/issues/1772

## Tested Versions

- 2.2.10-2.0.2 -> No issue
- 2.3.4 -> Circula ref issue

Here's a reproducer: https://github.com/esafirm/circular-ref-kapt-ksp

hfmehmed:
Fixes a circular dependency in KSP when in use with legacy-kapt plugin

KotlinCompile.libraries contains the output classes dir of kapt. Since this is a lazy gradle api, filtering does not remove the task dependency so ksp ends up depending on kapt. KAPT itself depends on KSP so this leads to a circular dependency.

Instead use KotlinCompilation.compileDependencyFiles which only contains the compilation jars. On top we need to add the android bootclasspath (containing android.jar and build tools jar)

Fixes #2743

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
@@ -17,9 +17,11 @@
 
 package com.google.devtools.ksp.gradle
 
+import com.android.build.api.variant.AndroidComponentsExtension
 import com.google.devtools.ksp.gradle.utils.allKotlinSourceSetsObservable
 import com.google.devtools.ksp.gradle.utils.canUseGeneratedKotlinApi
 import com.google.devtools.ksp.gradle.utils.enableProjectIsolationCompatibleCodepath
+import com.google.devtools.ksp.gradle.utils.isAgpBuiltInKotlinUsed
 import com.google.devtools.ksp.impl.KotlinSymbolProcessing
 import com.google.devtools.ksp.processing.ExitCode
 import com.google.devtools.ksp.processing.KSPCommonConfig
@@ -31,6 +33,7 @@ import com.google.devtools.ksp.processing.KspGradleLogger
 import org.gradle.api.DefaultTask
 import org.gradle.api.JavaVersion
 import org.gradle.api.artifacts.Configuration
+import org.gradle.api.attributes.Attribute
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.logging.LogLevel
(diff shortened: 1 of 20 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -223,21 +226,38 @@ abstract class KspAATask @Inject constructor(
...
                                 }
                             )
-                        )
+                        } else {
+                            val kaptGeneratedClassesDir = getKaptGeneratedClassesDir(project, sourceSetName)
+                            val kspOutputDir = KspGradleSubplugin.getKspOutputDir(project, sourceSetName, target)
+                            cfg.libraries.from(
+                                project.files(
+                                    Callable {
+                                        kotlinCompileProvider.get().libraries.filter {
```
ting-yuan, line 252: If `KotlinCompile.libraries` is going to be removed, do you have a schedule of removing the else part?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -223,21 +226,38 @@ abstract class KspAATask @Inject constructor(
...
+                            val androidComponents = project.extensions.getByType(AndroidComponentsExtension::class.java)
+                            cfg.libraries.from(androidComponents.sdkComponents.bootClasspath)
+                            cfg.libraries.from(
+                                project.provider {
+                                    val configuration = project.configurations.getByName(
+                                        kotlinCompilation.compileDependencyConfigurationName
+                                    )
+                                    configuration.incoming.artifactView { config ->
+                                        config.attributes.attribute(
+                                            Attribute.of("artifactType", String::class.java), "android-classes-jar"
```
ting-yuan, line 241: Looks like a bug in AGP or KGP to me. Why isn't `artifactType` properly set in `configuration`?

hfmehmed: As far as i understand the consumer of a specific configuration needs to define what artifactType they want to consume. So here we have KSP saying please give me the incoming dependencies from this configuration, and i only want to see classes jar even if the dependency contains much more. So i dont think this is a bug on the producer side of this configuration (which is KGP)

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
hfmehmed: I havent had a chance to think about it yet. There is 2 things that need to be answered:
- Is the alternative (the if case) a satisfactory replacement for this?
- Will KGP provide separate API similar to KotlinCompile.libraries but one that will exclude kapt

If KGP implements the new API we can move to it as soon as it is released. Otherwise I believe we need some time (so users can test this) to determine if the alternative really covers all edge cases out there

PR #2749, opened 2026-01-13
Bump Kotlin Analsys API to 2.3.20-dev-7064

ting-yuan:
with quite some cleanups.

### DIFF
--- a/kotlin-analysis-api/build.gradle.kts
+++ b/kotlin-analysis-api/build.gradle.kts
@@ -132,6 +132,8 @@ dependencies {
     libsForTestingCommon(kotlin("stdlib-common", aaKotlinBaseVersion))
 
     depJarsForCheck("org.jetbrains.kotlin", "kotlin-stdlib", kotlinBaseVersion)
+    depJarsForCheck("org.jetbrains.intellij.deps.kotlinx:kotlinx-coroutines-core-jvm:$aaCoroutinesVersion")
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -180,7 +180,7 @@ abstract class KspAATask @Inject constructor(
                 ),
                 project.dependencies.create("org.jetbrains.kotlin:kotlin-stdlib:$KSP_KOTLIN_BASE_VERSION"),
                 project.dependencies.create(
-                    "org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:$KSP_COROUTINES_VERSION"
+                    "org.jetbrains.intellij.deps.kotlinx:kotlinx-coroutines-core-jvm:$KSP_COROUTINES_VERSION"
```
hfmehmed, line 183: what was the reason for this change?

ting-yuan: intellij 251 depends on this forked version of coroutines. As a temporary workaround, I made KSP depend on it, too. Hopefully we can revert back to standard coroutines library after AA / compiler upgrades to intellij 253 and later. See https://youtrack.jetbrains.com/issue/KT-81457/Update-IntelliJ-SDK-dependency-to-253 for details.

hfmehmed: Thanks for the details

PR #2757, opened 2026-01-28
Remove KSP1 deprecation warning logic

hfmehmed:


ting-yuan (commented):
We'd have to explain this very explicitly in the release note and communicate in all the release announcements. Can we, instead of removing the warnings, make it toggle-able through an environment variable?

PR #2759, opened 2026-01-29
Bump Kotlin Analysis API to 2.4.0-dev-169

jaschdoc:


jaschdoc:
I'm getting a build error on `./gradlew :kotlin-analysis-api:validateShadowJar` but there is no difference between the baseline expected `shadow-validation-baseline.txt` and the actual generated file. What could be the reason for this? Note: I also get this on `main` so maybe it's not my branch that's causing the error, but it definitely seems to be something that calls for investigation.

EDIT: I have submitted a fix here https://github.com/google/ksp/pull/2762

ting-yuan:
Ah, I see what you mean. Please never mind the comments on multiple commits. Thanks for that fix too!

Issue #2651, opened 2025-10-16
symbol-processing-aa-embeddable is now packing CVE-2022-23305 vulnerability (log4j files)


achmyr:
**Context:** 
With the change: https://github.com/google/ksp/commit/046b5398bbb3dc8439b4f16c5a9f119a03c9ff7c, all of log4j-v1.jar files went into the jar file, including JDBCAppender class flagged by [CVE-2022-23305](https://nvd.nist.gov/vuln/detail/CVE-2022-23305)

**Recommendation from CVE:** Users should upgrade to Log4j 2, as it addresses numerous other issues from the previous versions.

**Consequence:** This blocks KSP upgrades for all the projects protected by Security Scanner tools (like NexusIQ)

cc: @ting-yuan

PR #2761, opened 2026-01-30
Exclude `jdbc` package from `log4j`

jaschdoc:
CVE-2022-23305 is caused by misconfiguration of `JDBCAppender` which is now removed.

I have checked that the added validation logic fails if `log4j` is not filtered.

Fixes https://github.com/google/ksp/issues/2651

### DIFF
--- a/kotlin-analysis-api/build.gradle.kts
+++ b/kotlin-analysis-api/build.gradle.kts
@@ -2,6 +2,7 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import com.google.devtools.ksp.RelativizingInternalPathProvider
 import com.google.devtools.ksp.RelativizingPathProvider
 import java.io.ByteArrayOutputStream
+import java.util.jar.JarFile
 
 description = "Kotlin Symbol Processing implementation using Kotlin Analysis API"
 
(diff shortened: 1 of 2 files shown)


jaschdoc:
Marking as draft for now. I will extend the build files to validate that the `jdbc` package is gone from the built jars.

jaschdoc:
> My I know why not simply bumping log4j version, or switching to jetbrains' fork if still used, or completely removing it if not used?
> 
> Excluding the contents from the uber jar seems a bit tricky.

I attempted to remove it entirely, but Kotlin still requires it (and still depends on version `1.2.17.2` that we also have). We could probably update to version `1.2.17.3` since tests are green but it seems like a bad idea to introduce such a mismatch between KSP and Kotlin. For now this seems like the best *local* fix that we can apply.

ting-yuan:
I see. But isn't the Kotlin compiler also suffering from the same vulnerability? How do they handle it?

jaschdoc:
> I see. But isn't the Kotlin compiler also suffering from the same vulnerability? How do they handle it?

I'm not sure they really handle it at all, but I also do not think they are affected by it. The CVE requires that you misconfigure `org.apache.log4j.jdb.JDBCAppender` to allow for an attack in the first place, and Kotlin does not mention this class at all: https://github.com/search?q=repo%3AJetBrains%2Fkotlin+%22jdbc%22&type=code

Here is the important sentence from description of the CVE:

> **Note this issue only affects Log4j 1.x when specifically configured to use the JDBCAppender, which is not the default.**

Here is the description of the CVE (I have highlighted the same sentence as above):

> By design, the JDBCAppender in Log4j 1.2.x accepts an SQL statement as a configuration parameter where the values to be inserted are converters from PatternLayout. The message converter, %m, is likely to always be included. This allows attackers to manipulate the SQL by entering crafted strings into input fields or headers of an application that are logged allowing unintended SQL queries to be executed. **Note this issue only affects Log4j 1.x when specifically configured to use the JDBCAppender, which is not the default.** Beginning in version 2.0-beta8, the JDBCAppender was re-introduced with proper support for parameterized SQL queries and further customization over the columns written to in logs. Apache Log4j 1.2 reached end of life in August 2015. Users should upgrade to Log4j 2 as it addresses numerous other issues from the previous versions.

https://nvd.nist.gov/vuln/detail/cve-2022-23305

Of course, long-term it is better to suggest to them that they either also remove the class entirely or upgrade to log4j2.

PR #2766, opened 2026-02-02
Fix deprecation warnings of `.nullability` in `KSTypeImpl`

jaschdoc:
I have attempted to the best of my knowledge to remove use of `.nullability` since the field is deprecated. I have replaced it with `isMarkedNullable` or `isNullable` where appropriate. Any feedback or corrections is much appreciated as I'm a little unsure if it's entirely correct everywhere.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
@@ -70,8 +70,10 @@ class KSTypeImpl private constructor(internal val type: KaType) : KSType {
 
     override val nullability: Nullability by lazy {
         when {
-            type is KaFlexibleType && type.lowerBound.nullability != type.upperBound.nullability -> Nullability.PLATFORM
-            analyze { type.canBeNull } -> Nullability.NULLABLE
+            type is KaFlexibleType && analyze {
+                !type.lowerBound.isMarkedNullable && type.upperBound.isMarkedNullable
+            } -> Nullability.PLATFORM
+            analyze { type.isNullable } -> Nullability.NULLABLE
             else -> Nullability.NOT_NULL
         }
     }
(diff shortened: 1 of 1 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
```diff
@@ -70,8 +70,9 @@ class KSTypeImpl private constructor(internal val type: KaType) : KSType {
 
     override val nullability: Nullability by lazy {
         when {
-            type is KaFlexibleType && type.lowerBound.nullability != type.upperBound.nullability -> Nullability.PLATFORM
-            analyze { type.canBeNull } -> Nullability.NULLABLE
+            type is KaFlexibleType &&
+                analyze { type.lowerBound.isMarkedNullable != type.upperBound.isMarkedNullable } -> Nullability.PLATFORM
```
jaschdoc: Maybe this should be `type.lowerBound.isNullable` and `type.upperBound.isNullable`?

ting-yuan: Did you mean `!type.lowerBound.isNullable && type.upperBound.isNullable`?

Strictly speaking, the original implementation may be a bit over-optimized. [Platform types](https://kotlinlang.org/docs/java-interop.html#null-safety-and-platform-types) **is implemented** by flexible types that lower bound and upper bound are based on the same type with lower bound non-nullable and upper bound nullable.

jaschdoc: Aha, that makes sense. I updated it to `!type.lowerBound.isNullable && type.upperBound.isNullable`.

EDIT: Doing so fails in `KSPAATest.testAsMemberOf`, so either `isMarkedNullable` is the correct property to use or there is a defect somewhere.

PR #2773, opened 2026-02-05
Bump kotlin analysis API to 2.4.0-dev-2124

hfmehmed:
Bumping in order to take in a fix in Kotlin AA (https://youtrack.jetbrains.com/issue/KT-83801/Nested-typealiases-are-not-correctly-indexed-in-standalone-mode) that will unblock us supporting nested typealiases

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/standalone/IncrementalKotlinPackageProviderFactory.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/standalone/IncrementalKotlinPackageProviderFactory.kt
@@ -23,7 +23,11 @@ class IncrementalKotlinPackageProviderFactory(
     }
 
     fun update(files: Collection<KtFile>) {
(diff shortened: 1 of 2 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/standalone/IncrementalKotlinPackageProviderFactory.kt
```diff
@@ -21,7 +23,7 @@ class IncrementalKotlinPackageProviderFactory(
     }
 
     fun update(files: Collection<KtFile>) {
-        val staticFactory = KotlinStandalonePackageProviderFactory(project, files)
+        val staticFactory = KotlinStandalonePackageProviderFactory(project, files, emptyList())
```
jaschdoc: Should we name this `emptyList()` parameter so it's not just a magic value? E.g., `libraryRoots = emptyList()`.

hfmehmed: Good call. Updated

PR #2777, opened 2026-02-08
Register generated ksp output for Android KMP plugin

### LINKED ISSUE #2744: Android KMP + KSP on IP-compatible path ignores generated java sources
`compileAndroidMainJavaWithJavac` task expects generated sources at `[...]\build\generated\ksp\android\androidMain\classes`. The actual path is `[...]\build\generated\ksp\android\androidMain\java`

* KSP 2.3.5-SNAPSHOT
* AGP 9.0.0-rc02
* Kotlin 2.3.0

See [AGP9DaggerKSP.zip](https://github.com/user-attachments/files/24434538/AGP9DaggerKSP.zip). Run `.\gradlew build`.

#### Workaround

```kotlin
androidComponents {
    onVariants { variant ->
        val ksp = tasks.named<KspAATask>("ksp" + variant.name.capitalize())
        variant.sources.java?.addGeneratedSourceDirectory(ksp) { it.kspConfig.javaOutputDir }
    }
}
```

hfmehmed:
Fixes #2744 
And errors such as 
```
Reason: Task ':extra:settings:api:extractAndroidMainAnnotations' uses this output of task ':extra:settings:api:kspAndroidMain' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.

Possible solutions:
1. Declare task ':extra:settings:api:kspAndroidMain' as an input of ':extra:settings:api:extractAndroidMainAnnotations'.
2. Declare an explicit dependency on ':extra:settings:api:kspAndroidMain' from ':extra:settings:api:extractAndroidMainAnnotations' using Task#dependsOn.
3. Declare an explicit dependency on ':extra:settings:api:kspAndroidMain' from ':extra:settings:api:extractAndroidMainAnnotations' using Task#mustRunAfter.

For more information, please refer to https://docs.gradle.org/9.2.1/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.
```
when using the new Android-kmp plugin, which previously required manual workaround like 
```
tasks.named { it == "kspAndroidMain" }.configureEach {
    dependsOn(tasks.named("extractAndroidMainAnnotations"))
}
```

`afterEvaluate` is needed to ensure that the Variant API callback is executed when we retrieve the Component object from the cache.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -25,6 +25,7 @@ import com.google.devtools.ksp.gradle.utils.canUseInternalKspApis
 import com.google.devtools.ksp.gradle.utils.checkMinimumAgpVersion
 import com.google.devtools.ksp.gradle.utils.enableProjectIsolationCompatibleCodepath
 import com.google.devtools.ksp.gradle.utils.isAgpBuiltInKotlinUsed
+import com.google.devtools.ksp.gradle.utils.minimumAndroidKotlinMultiplatformVersion
 import com.google.devtools.ksp.gradle.utils.useLegacyVariantApi
 import org.gradle.api.GradleException
 import org.gradle.api.Project
@@ -261,17 +262,19 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             )
         }
 
+        // The variant API in KMP runs after KotlinCompilerPluginSupportPlugin.applyToCompilation()
+        // afterEvaluate is needed here to ensure the Variant API onVariants callback has run and populated the
+        // androidComponentCache, before we attempt to query for the Component object
         project.afterEvaluate {
-            try {
-                if (kotlinCompilation is KotlinMultiplatformAndroidCompilation) {
-                    val component = androidComponentCache.get(kotlinCompilation.componentName)
-                    AndroidPluginIntegration.registerGeneratedSourcesKmp(
-                        kspTaskProvider = kspTaskProvider,
-                        androidComponent = component,
-                    )
-                }
-            } catch (e: NoClassDefFoundError) {
-                // ignore
+            if (project.isAndroidKmpProject() &&
+                project.minimumAndroidKotlinMultiplatformVersion() &&
+                kotlinCompilation is KotlinMultiplatformAndroidCompilation
+            ) {
+                val component = androidComponentCache.get(kotlinCompilation.componentName)
+                AndroidPluginIntegration.registerGeneratedSourcesKmp(
+                    kspTaskProvider = kspTaskProvider,
+                    androidComponent = component,
+                )
             }
         }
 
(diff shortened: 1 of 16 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -248,6 +248,7 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             }
         }
         if (kotlinCompilation is KotlinJvmAndroidCompilation) {
+            val component = androidComponentCache.get(kotlinCompilation.name)
```
cmwarrington, line 252: unused? or is it that you need to do this for a side-effect? If so worth a comment.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -260,6 +261,20 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             )
         }
 
+        project.afterEvaluate {
```
cmwarrington, line 268: Also worth a comment for why afterEvaluate makes sense here

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -260,6 +261,20 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
...
+            try {
+                if (kotlinCompilation is KotlinMultiplatformAndroidCompilation) {
+                    val component = androidComponentCache.get(kotlinCompilation.componentName)
+                    AndroidPluginIntegration.registerGeneratedSourcesKmp(
+                        kspTaskProvider = kspTaskProvider,
+                        androidComponent = component,
+                    )
+                }
+            } catch (e: NoClassDefFoundError) {
```
cmwarrington: Is this for KotlinMultiplatformAndroidCompilation ? I wonder if you could avoid wrapping  the registerGeneratedSourcesKmp within the catch to make it clearer which bit you expect to throw.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
hfmehmed: It is used. Is it passed to AndroidPluginIntegration.syncSourceSets. The reason for moving it down here is it is only required in this flow. For kmp we query differently

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
hfmehmed: Done

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
hfmehmed: removed the try-catch, instead added checks for pluing application and agp version

PR #2780, opened 2026-02-09
Fix build warnings in `TestModule`

jaschdoc:
This PR removes calls the deprecated `StringBuilder.appendln` method and replaces it with `StringBuilder.appendLine`. It also asserts that `File.getName` is a non-null string and throws an error otherwise. Maybe the explicit throw can be omitted if it is handled by Kotlin's implicit non-null assertions? Although `File.getName` is annotated with `NotNull` it still seems safer to throw at the binding site.

### DIFF
--- a/gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/testing/TestModule.kt
+++ b/gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/testing/TestModule.kt
@@ -35,7 +35,7 @@ class TestModule(
     val plugins = LinkedHashSet(plugins)
     val dependencies = LinkedHashSet(dependencies)
     val buildFileAdditions = LinkedHashSet<String>()
-    val name = moduleRoot.name
+    val name: String = moduleRoot.name
 
     init {
         moduleRoot.mkdirs()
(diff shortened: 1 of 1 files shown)


gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/testing/TestModule.kt
```diff
@@ -35,7 +35,7 @@ class TestModule(
     val plugins = LinkedHashSet(plugins)
     val dependencies = LinkedHashSet(dependencies)
     val buildFileAdditions = LinkedHashSet<String>()
-    val name = moduleRoot.name
+    val name: String = moduleRoot.name ?: throw Exception("unexpected empty string: 'moduleRoot.name'")
```
jaschdoc: Is it better to just annotate the binding without throwing?

ting-yuan: The compiler should automatically insert a null check and inference a non-null `String` type. Why is this needed?

jaschdoc: Okay then it's as I expected. I was unsure about that. I will remove the explicit exception.

PR #2787, opened 2026-02-11
Cherry-pick optimization to MockFileIndexFacade

ting-yuan:
Original commit id and message from
https://github.com/JetBrains/intellij-community

4c6eef049c6088759467837ef392af2233adc427
IDEA-382356 [java] Optimize MockFileIndexFacade#isInLibraryClases

See also: https://youtrack.jetbrains.com/issue/IDEA-382356/

### DIFF
--- a/kotlin-analysis-api/src/main/java/com/intellij/mock/MockFileIndexFacade.java
+++ b/kotlin-analysis-api/src/main/java/com/intellij/mock/MockFileIndexFacade.java
@@ -1,0 +1,97 @@
+// Copyright 2000-2023 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.intellij.mock;
+
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.UnloadedModuleDescription;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.roots.FileIndexFacade;
+import com.intellij.openapi.util.ModificationTracker;
+import com.intellij.openapi.vfs.VfsUtilCore;
+import com.intellij.openapi.vfs.VirtualFile;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+
+public class MockFileIndexFacade extends FileIndexFacade {
+  private final Module myModule;
+  private final Set<VirtualFile> myLibraryRoots = new HashSet<>();
+
+  public MockFileIndexFacade(final Project project) {
+    super(project);
+    myModule = null;  // TODO
+  }
+
+  @Override
+  public boolean isInContent(@NotNull VirtualFile file) {
+    return true;
+  }
+
+  @Override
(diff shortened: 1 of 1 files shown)


kotlin-analysis-api/src/main/java/com/intellij/mock/MockFileIndexFacade.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Override
+  public boolean isInLibraryClasses(@NotNull VirtualFile file) {
+    VirtualFile parent = file.getParent();
+    while (true) {
+      if (parent == null) return false;
+      if (myLibraryRoots.contains(parent)) return true;
+      parent = parent.getParent();
+    }
+  }
```
jaschdoc, line 52: This is safe but seems fragile. I know this is what was committed upstream, but would there be any performance issues with writing it in a safer way?

```java
VirtualFile parent = file.getParent();
while (parent != null) {
  if (myLibraryRoots.contains(parent)) return true;
  parent = parent.getParent();
}
return false;
```

jaschdoc: @bcorso, any thoughts on this?

bcorso: > would there be any performance issues with writing it in a safer way

I don't think there would be any performance issues with the change you suggested. 

However, while I do prefer your version stylistically, it probably doesn't make sense to change it here unless you're also going to try to commit the change upstream as well since it will eventually be replaced by the upstream version anyway.

ting-yuan:
Well, in addition to the style, the original commit has a bug. It always skips the last, leaf element of `file`. E.g., when `file` is`/x/y/z`, it only searches for `/x/y` and `/x`. Before that commit it searches for the leaf.

kotlin-analysis-api/src/main/java/com/intellij/mock/MockFileIndexFacade.java
ting-yuan: The original commit actually has a bug so there is an opportunity to update the style in the upstream too :)

bcorso: Nice catch! 

Let's update [IDEA-382356](https://youtrack.jetbrains.com/issue/IDEA-382356/Optimize-MockFileIndexFacadeisInLibraryClases) with the bug you found and you can suggest the style change there too.

ting-yuan: I filed a bug at https://youtrack.jetbrains.com/issue/IJPL-234752/MockFileIndexFacade.isInLibraryClasses-misses-leaves. An alternative way to do so might be sending a PR.

jaschdoc: Opened a PR here: https://github.com/JetBrains/intellij-community/pull/3422

PR #2788, opened 2026-02-13
Add CODEOWNERS file

jaschdoc:
This is a suggestion for a `CODEOWNERS` file. It's based on`git-log` / `git-blame`. Are there any areas you think should be differently covered or in greater / less detail?
The `CODEOWNERS` file is there intended to ensure every PR receives attention from familiar contributors. It should be viewed as a baseline. Reviews from other relevant people should be encouraged for all PRs.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,26 @@
+* @ting-yuan @hfmehmed @jaschdoc
+
+/.github/ @ting-yuan @hfmehmed @jaschdoc
+/buildSrc/ @ting-yuan @hfmehmed
+/gradle-plugin/ @ting-yuan @hfmehmed
+build.gradle.kts @hfmehmed
+settings.gradle.kts @hfmehmed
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,26 @@
...
+gradle.properties @hfmehmed
+gradlew @hfmehmed
+
+/api/ @ting-yuan @jaschdoc
+/symbol-processing/ @ting-yuan @jaschdoc
+/kotlin-analysis-api/ @ting-yuan @jaschdoc
+/symbol-processing-aa-embeddable/ @ting-yuan @hfmehmed @jaschdoc
+
+/integration-tests/ @hfmehmed @ting-yuan
+/benchmark/ @ting-yuan
```
ting-yuan: This is used by some internal nightly perf tests in AAR. I think you're the better owner than me, if those tests are still running and relevant.

Issue #2774, opened 2026-02-06
Module recompilation fails on Windows when KSP2 is enabled


ruxbit:
## Description
When running the same Gradle task twice in a row on Windows, the 2nd run fails with a file-lock error on `classes.jar`.

This is reproducible only on Windows and only with KSP2 enabled (`ksp.useKSP2=true`).
With KSP1 (`ksp.useKSP2=false`) the issue does not reproduce.

Minimal reproducer repo:
https://github.com/ruxbit/ksp2reproducer

## Environment
- JDK: JetBrains Runtime 21.0.8
- Gradle: 8.13
- Android Gradle Plugin: 8.10.0
- Kotlin Gradle Plugin: 2.2.21
- KSP Gradle Plugin: 2.2.21-2.0.5

## Steps to reproduce
1. Clone the reproducer:
   https://github.com/ruxbit/ksp2reproducer

2. Ensure KSP2 is enabled:
   - `ksp.useKSP2=true` (via `gradle.properties`)

3. Run this command (1st run succeeds):
   ./gradlew :lib:common:media:assembleDebug --rerun-tasks

4. Run the same command again (2nd run fails):
   ./gradlew :lib:common:media:assembleDebug --rerun-tasks

## Actual result
2nd run fails with:

> Execution failed for task ':lib:common:collections:bundleLibCompileToJarDebug'.
java.nio.file.FileSystemException: D:\Wildberries\ksp2reproducer\lib\common\collections\build\intermediates\compile_library_classes_jar\debug\bundleLibCompileToJarDebug\classes.jar: The process cannot access the file because it is being used by another process

## Expected result
The 2nd run should also succeed (no stale file handle / no locked `classes.jar`).

## Additional notes / observations
- Reproduces only on Windows (Linux/macOS do not reproduce).
- Reproduces only with KSP2:
  - `ksp.useKSP2=true` -> fails on 2nd run
  - `ksp.useKSP2=false` -> OK (both runs succeed)
- If I remove `ksp(libs.androidx.room.compiler)` from module `:lib:common:media`, the issue disappears.
- If I keep KSP enabled but use *any other* KSP processor instead of Room, the issue still reproduces.
- Project structure is intentionally minimal:
  - `:lib:common:collections` contains only a single Kotlin class `A`
  - `:lib:common:media` contains only `B(val a: A)`, depends on `collections`, applies KSP, and adds a KSP processor dependency

hfmehmed:
[...]

PR #2790, opened 2026-02-13
Fix potential file handle leaks by clearing IntelliJ VFS cache

hfmehmed:
Improves resource management in the KSP Analysis API implementation by ensuring IntelliJ file handlers are properly cleaned up. When using the Kotlin Analysis API, the underlying CoreJarFileSystem maintains a cache of file handlers. If not explicitly cleared, this leads to issues in Windows where the OS holds the lock on a specific file making it impossible to re-run the build unless the existing gradle daemon is terminated. Linux and Mac os are more lenient in terms of file locking so this does not seem to be an issue there

Fixes: #2774


Hamoo456 (changes requested):
good

PR #2791, opened 2026-02-16
Test for running same gradle task in a row on windows

hfmehmed:


hfmehmed:
Yes but this test didnt work as i expected.

PR #2804, opened 2026-02-18
Add new example project

jaschdoc:
This example project is part of updating the documentation on kotlin lang's website.


jaschdoc:
@ting-yuan TL;DR: I'm unsure whether to include gradle-specific files and binaries in this example. What are your thoughts?

I want to leave out `gradlew` scripts and the `gradle.jar` since these files will rarely be used (this is after all just a small demo / example) and it seems wasteful to store those files in the repo. It also means that we have to keep them up to date with gradle versions. On the other hand, it allows a reader to see what a complete runnable project looks like. If we run the examples in CI then it seems like a good idea to have the gradle scripts and jar in there (but it again the maintenance problem could probably also be automated in the future).

PR #2809, opened 2026-02-24
Fix kotlinx coroutines dependency

jaschdoc:
The `1.8.0-intellij-13` version is published under the `com.intellij.platform` group. This is the version that Kotlin uses, and consequently the one we need.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
@@ -79,6 +79,7 @@ import java.io.ByteArrayOutputStream
 import java.io.File
 import java.io.ObjectOutputStream
 import java.lang.reflect.InvocationTargetException
+import java.net.URI
 import java.net.URLClassLoader
 import java.nio.file.Paths
 import java.util.*
@@ -136,6 +137,7 @@ abstract class KspAATask @Inject constructor(
                         kspConfig.processorClasspath,
                     )
                 }
+
                 else -> {
                     if (
                         !inputChanges.isIncremental ||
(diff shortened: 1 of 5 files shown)


jaschdoc:
@ting-yuan, do you think we should merge this?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -183,11 +185,14 @@ abstract class KspAATask @Inject constructor(
                 ),
                 project.dependencies.create("org.jetbrains.kotlin:kotlin-stdlib:$KSP_KOTLIN_BASE_VERSION"),
                 project.dependencies.create(
-                    "org.jetbrains.intellij.deps.kotlinx:kotlinx-coroutines-core-jvm:$KSP_COROUTINES_VERSION"
+                    "com.intellij.platform:kotlinx-coroutines-core-jvm:$KSP_COROUTINES_VERSION"
```
ting-yuan: Can you try to see if the standard `org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.8.0` (only here, not in the build.gradle.kts) works, i.e., make the integration tests pass? The issue with intellij's fork of coroutines lib is that it contains some classes (e.g., `IntellijCoroutines`) that aren't available in the standard version. While this patch seems to align with the compiler better, I'm curious about what's the difference between the mid-stream version we're using. If the most-downstream version (this patch) from intellij platform still requires `IntellijCoroutines`, the depencency problem remains unresolved.

See this for details:
https://github.com/JetBrains/kotlin/commit/3534805795c66fd9edbafa7865c5b5e03a26bab9

jaschdoc: That results in test failures with the exact stack trace:
```
java.lang.NoClassDefFoundError: kotlinx/coroutines/internal/intellij/IntellijCoroutines
...
```

See github workflow run for https://github.com/google/ksp/commit/5ef29cac239d3c94c5b0f3f42800bcd95dba6a31

jaschdoc:
Will do :)

PR #2813, opened 2026-02-26
Fix getJvmName for @JvmRecord data class properties

### LINKED ISSUE #2812: getJvmName returns incorrect accessor names for @JvmRecord data classes
## Summary

`Resolver.getJvmName(accessor)` returns `getName` for a property getter on a `@JvmRecord` Kotlin data class, but the actual JVM bytecode uses record component accessors (`name()`, not `getName()`). This causes downstream frameworks (e.g. Micronaut Data) that rely on `getJvmName` to generate incorrect reflective calls, resulting in `NoSuchMethodError` at runtime.

## Environment

- KSP version: 2.2.21-2.0.5 (also reproducible on `main` at 2.3.0)
- Kotlin version: 2.2.21+
- JDK: 17+

## Reproducer

Given this Kotlin source:

```kotlin
@JvmRecord
data class Author(
    val id: Int,
    val name: String,
)
```

A KSP processor calling:

```kotlin
val cls = resolver.getClassDeclarationByName("Author")!!
val nameGetter = cls.getAllProperties().first { it.simpleName.asString() == "name" }.getter!!
val jvmName = resolver.getJvmName(nameGetter)
println(jvmName) // prints "getName" — should print "name"
```

## Expected behavior

`resolver.getJvmName(getter)` should return `"name"` (the record component accessor), matching the actual JVM bytecode produced by the Kotlin compiler for `@JvmRecord` classes.

## Actual behavior

Returns `"getName"` (the bean-style getter), which does not exist on the compiled Java record class.

## Root cause

In `ResolverAAImpl.getJvmName(accessor: KSPropertyAccessor)` (line 445 of `ResolverAAImpl.kt`), the name is computed as:

```kotlin
val name = accessor.receiver.simpleName.asString()
val prefixedName = when (accessor) {
    is KSPropertyGetter -> JvmAbi.getterName(name)   // unconditionally → "getName"
    is KSPropertySetter -> JvmAbi.setterName(name)
    else -> ""
}
```

`JvmAbi.getterName()` unconditionally adds a `get` prefix. There is no check for whether the containing class is a `@JvmRecord` class, which suppresses the prefix at the JVM level.

### Why KSP1 worked correctly

The KSP1 implementation (`ResolverImpl`) delegated to `typeMapper.mapFunctionName(descriptor, OwnerKind.IMPLEMENTATION)`, which is the Kotlin compiler's internal `KotlinTypeMapper`. That function explicitly checks for `java.lang.Record` supertypes and returns the bare property name when the class is a record. KSP2 replaced this with a manual `JvmAbi.getterName()` call and lost the record-awareness.

## Impact

Any KSP processor that uses `getJvmName` to determine how to reflectively access properties on `@JvmRecord` classes will get the wrong method name. Known affected framework: Micronaut Data's `micronaut-inject-kotlin` processor, which generates bean introspection code using `getJvmName` — causing `NoSuchMethodError` at runtime for all `@JvmRecord` entities.

## Proposed fix

In `ResolverAAImpl.getJvmName(accessor)`, before computing the prefixed name, check whether the containing class has the `kotlin.jvm.JvmRecord` annotation. If so, return the bare property name (matching the Kotlin compiler's bytecode output):

```kotlin
// Annotation classes and @JvmRecord data classes both use bare property names
// as accessor names (no get/set prefix).
if (containingClass?.classKind == ClassKind.ANNOTATION_CLASS ||
    containingClass != null && containingClass.annotations.any {
        it.annotationType.resolve().declaration.qualifiedName?.asString() == "kotlin.jvm.JvmRecord"
    }
) {
    return name
}
```

For setters, records don't have setters at all (record components are final), so the setter branch is unaffected.

MariusVolkhart:
@JvmRecord data classes compile to Java records, whose component accessors use bare property names (e.g. name()) rather than bean-style getters (getName()). ResolverAAImpl.getJvmName unconditionally used JvmAbi.getterName() which always added the get prefix.

Add a check for the @JvmRecord annotation alongside the existing annotation class check, since both use bare property names as accessor names.

Fixes #2812


MariusVolkhart:
CLA violation is due to Claude Code. I didn't see anything in the CONTRIBUTING.md indicating agentic AI was not allowed, and wanted to be clear on the fact that it was used as part of this change. Please advise if you want me to make changes or are unable to accept the PR as a result. Thank you.

hfmehmed:
To me it seems like the tool requires all the owners of the commit (you and anthropic agent) need to sign the CLA. Obviously, the agent itself can't sign it so probably to the time being the workaround is not commit code using the agent

kotlin-analysis-api/testData/jvmNameRecord.kt
```diff
@@ -0,0 +1,12 @@
...
+// (x, null), (y, null)
+// (x, null), (y, null)
+// END
+// MODULE: main
+// FILE: TestRecordClass.kt
+@JvmRecord
+data class TestRecordClass(val x: Int, val y: String)
+// FILE: TestLibRecordClass.kt
+@JvmRecord
+data class TestLibRecordClass(val x: Int, val y: String)
```
jaschdoc: How are these classes distinct except for the name? It's better if the classes are a bit more diverse. I'm thinking an approximation of a matrix of the following:
- Classes with different number of parameters
- Different property names
- Classes extending other types
- Classes with type parameters
- Classes marked as JvmRecords that override a val declared in a non-annotation interface, e.g., `interface Example { val x: Int }`
- Classes with different names (like you already have)
- Classes with `var` and `val` properties

If you have type parameters and or normal parameters and a class that extends another type, then it's also a good idea to override some parameters and add some locally, e.g.,
```kotlin
interface Example<A> {
    val x: A
}

data class Ext<A, B>(override val x: A, val y: B) : Example<A>
```

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/JvmNameRecordProcessor.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import com.google.devtools.ksp.KspExperimental
+import com.google.devtools.ksp.getClassDeclarationByName
+import com.google.devtools.ksp.processing.Resolver
+import com.google.devtools.ksp.symbol.KSAnnotated
+
+class JvmNameRecordProcessor : AbstractTestProcessor() {
+    val results = mutableListOf<String>()
+    override fun toResult(): List<String> {
+        return results
```
jaschdoc, line 29: This should be sorted to ensure results are reproducible.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -452,11 +452,18 @@ class ResolverAAImpl(
...
+
+        // Annotation classes and @JvmRecord data classes both use bare property names
+        // as accessor names (no get/set prefix).
+        if (containingClass?.classKind == ClassKind.ANNOTATION_CLASS ||
+            containingClass != null && containingClass.annotations.any {
+                it.annotationType.resolve().declaration.qualifiedName?.asString() == "kotlin.jvm.JvmRecord"
+            }
+        ) {
+            return name
+        }
```
jaschdoc: I think we can improve the readability of this by extracting some variables so it reads
```kotlin
if (isAnnotationClass || isJvmRecord) {
    return name
}
```

Additionally, we can shortcircuit the expensive type resolution by checking for the short name first in the lambda:
```kotlin
any {
    it.shortName.asString() == "JvmRecord" ||
        it.annotationType.resolve().declaration.qualifiedName?.asString() == "kotlin.jvm.Record
}
```

Lastly, maybe we can rename `name` to `propertyName` so it's clearer what the name. I know you didn't change it, but let's improve it.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPAA17Test.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.jetbrains.kotlin.config.JvmTarget
+import org.jetbrains.kotlin.test.TestMetadata
+import org.jetbrains.kotlin.test.builders.TestConfigurationBuilder
+import org.jetbrains.kotlin.test.directives.JvmEnvironmentConfigurationDirectives
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.parallel.Execution
+import org.junit.jupiter.api.parallel.ExecutionMode
+
+@Execution(ExecutionMode.SAME_THREAD)
+class KSPAA17Test : AbstractKSPAATest() {
```
jaschdoc: Did you try adding the test to the existing test suite?

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/JvmNameRecordProcessor.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @OptIn(KspExperimental::class)
+    override fun process(resolver: Resolver): List<KSAnnotated> {
+        listOf("TestRecordClass", "TestLibRecordClass").forEach { clsName ->
+            resolver.getClassDeclarationByName(clsName)?.let { cls ->
+                results.add(
+                    cls.getAllProperties().map {
+                        "(${it.getter?.let { resolver.getJvmName(it) }}, " +
+                            "${it.setter?.let { resolver.getJvmName(it) }})"
+                    }.toList().joinToString()
```
jaschdoc: It's more stable if you use `resolver.getSymbolsWithAnnotation("kotlin.jvm.JvmRecord")` and filter for class declaration instances.
It might also be a bit cleaner if you flatmap over `getAllProperties` and then return a list of size at most two, that contains the getter and setter names which you can the call `joinToString` on, e.g.,
```kotlin
getAllProperties().flatMap {
    // create a list that contains the jvm getter name if it exists and the setter name if it exists
}
```
The names should probably also be qualified / prefixed with the class names.

jaschdoc (commented):
Hey thank you for opening the PR and apologies for the slow response. I left some comments that aim to improve the test suite and readability. Cheers!

jaschdoc:
@MariusVolkhart, just pinging to see if you still want to contribute :)

kotlin-analysis-api/testData/jvmNameRecord.kt
MariusVolkhart: Done. Records can't extend classes or have var constructor properties, so those cells became interface supertypes (incl. the generic override shape) and a body var without a backing field — noted in the test file. Also added an aliased import, @get:JvmName, and a library-module record. Expected names verified against javap.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/JvmNameRecordProcessor.kt
MariusVolkhart: Done — results are sorted now.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
MariusVolkhart: Done, with one tweak: comparing source short names breaks on import aliases (import kotlin.jvm.JvmRecord as JR), so the check compares ClassIds on the analysis-api symbol instead — same as explictJvmName does for @JvmName just above in util.kt.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPAA17Test.kt
MariusVolkhart: Yes — a `// JVM_TARGET: 17` directive in the testData collides with the hardcoded default in `AbstractKSPTest` ("Too many values passed to JVM_TARGET: [1.8, 17]"); that's why the separate class existed. Removed the hardcoded default so file directives work, and the test now lives in KSPAATest. Isolated in [its own commit](https://github.com/google/ksp/pull/2813/changes/3dffe969897c81eaa5cdc0cfde8a6bcceed67cd9) — easy to drop if you'd rather keep this PR minimal.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/JvmNameRecordProcessor.kt
MariusVolkhart: Done. One wrinkle: `getSymbolsWithAnnotation` only returns current-compilation symbols, so the library-module record is looked up by name. That path exposed a real bug — `@JvmRecord` isn't retained in class files, so library records were reported as `getW`. Fixed in a separate commit by recognizing the `java.lang.Record` supertype instead.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
```diff
@@ -495,6 +495,12 @@ abstract class KSPUnitTestSuite(
         runTest("$AA_PATH/jvmName.kt")
     }
 
+    @TestMetadata("jvmNameRecord.kt")
+    @Test
+    fun testJvmNameRecord() {
+        runTest("../kotlin-analysis-api/testData/jvmNameRecord.kt")
```
jaschdoc: You can just use `"$AA_PATH/jvmNameRecord.kt"` here

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
```diff
@@ -495,6 +495,12 @@ abstract class KSPUnitTestSuite(
         runTest("$AA_PATH/jvmName.kt")
     }
 
+    @TestMetadata("jvmNameRecord.kt")
+    @Test
+    fun testJvmNameRecord() {
```
jaschdoc, line 501: It would be great if you could annotate the test with the `@Bug` annotation. That way, in the future we have some reference as to why the test is there. Example: https://github.com/google/ksp/blob/c116b824f301d6600846e1961ff04926cccd5e30/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt#L705

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/JvmNameRecordProcessor.kt
jaschdoc, line 1: Missing copyright header :)

kotlin-analysis-api/testData/jvmNameRecord.kt
jaschdoc, line 1: Can we add a test where one of the properties of the data class is also a data class / jvm record?

Also, missing copyright header :)

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AbstractKSPAATest.kt
```diff
@@ -86,14 +91,20 @@ abstract class AbstractKSPAATest(val experimentalPsiResolution: Boolean) : Abstr
         val compilerClass = URLClassLoader(arrayOf(), javaClass.classLoader).loadClass(K2JVMCompiler::class.java.name)
         val compiler = compilerClass.getDeclaredConstructor().newInstance()
         val execMethod = compilerClass.getMethod("exec", PrintStream::class.java, Array<String>::class.java)
-        execMethod.invoke(compiler, PrintStream(outStream), args.toTypedArray())
+        val exitCode = execMethod.invoke(compiler, PrintStream(outStream), args.toTypedArray())
+        check(exitCode.toString() == "OK") {
+            "Kotlin compilation failed with exit code $exitCode:\n$outStream"
+        }
```
jaschdoc: Please revert this :)

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/JvmNameRecordProcessor.kt
```diff
@@ -0,0 +1,36 @@
...
+        // so the record class from the library module is looked up explicitly.
+        val sourceRecords = resolver.getSymbolsWithAnnotation("kotlin.jvm.JvmRecord")
+            .filterIsInstance<KSClassDeclaration>()
+        val libRecords = listOfNotNull(resolver.getClassDeclarationByName("LibRecord"))
+        (sourceRecords + libRecords)
+            .flatMap { cls ->
+                cls.getAllProperties().map { property ->
+                    val accessorNames = listOfNotNull(
+                        property.getter?.let { resolver.getJvmName(it) },
+                        property.setter?.let { resolver.getJvmName(it) },
```
jaschdoc: I think we should also get the jvm name of the setter parameter.

kotlin-analysis-api/testData/jvmNameRecord.kt
```diff
@@ -0,0 +1,64 @@
...
+    var mutable: Int
+        get() = a
+        set(value) {}
+}
+
+// An explicit @JvmName takes precedence over the record accessor naming.
+@JvmRecord
+data class WithJvmName(@get:JvmName("customName") val n: Int)
+// FILE: aliased.kt
+import kotlin.jvm.JvmRecord as JR
```
jaschdoc, line 84: In addition to this, can we also create a test where we use a type alias for the annotation?

jaschdoc (commented):
Looks good. Only small nits :)

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
MariusVolkhart: Fixed — now uses `$AA_PATH/jvmNameRecord.kt`.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
MariusVolkhart: Added `@Bug("https://github.com/google/ksp/issues/2812", BugState.FIXED)`.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/JvmNameRecordProcessor.kt
MariusVolkhart: Done.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AbstractKSPAATest.kt
MariusVolkhart: Reverted.

kotlin-analysis-api/testData/jvmNameRecord.kt
MariusVolkhart: Copyright header added. Also added `WithRecordProp(val inner: Single, val n: Int)` where the property type is itself a record — accessor naming is unaffected by the property's type.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/JvmNameRecordProcessor.kt
MariusVolkhart: Added. There's no `getJvmName(KSValueParameter)` overload, so the setter's parameter name comes from `setter.parameter.name?.asString()` — gives `value` for the explicit setter in the test.

kotlin-analysis-api/testData/jvmNameRecord.kt
MariusVolkhart: Added `TypeAliased` via `typealias JvmRec = JvmRecord`. One wrinkle: PSI's `getSymbolsWithAnnotation` doesn't resolve typealias annotations, so `TypeAliased` is also looked up by name in the processor (same as `LibRecord`), with deduplication to avoid double-reporting in AA mode where it is found by both paths.

jaschdoc: Okay, thank you for pointing it out! That's a bug. I will open an issue.

jaschdoc (approved):
Looks good, thank you for the PR!

PR #2831, opened 2026-03-16
Upgrade all tests to JUnit 6.0.3

jaschdoc:
Depends on https://github.com/google/ksp/pull/2837


jaschdoc:
Closing this PR for now. We should try to bump each module independently or, in the case of integration tests, consider creating a new gradle module that uses the new junit version and incrementally migrate each test.

PR #2836, opened 2026-03-19
Bump caffeine to version 3.2.3

jaschdoc:


jaschdoc:
Requires upgrading from Java 8 to Java 11, but we should bump to Java 17. This means we can try to bump the version of caffeine again at that point.

PR #2837, opened 2026-03-19
Upgrade to Java 17

jaschdoc:
Can we do this or could there be downstream problems from this? Upgrading would allows us to upgrade some dependencies such as JUnit and caffeine.


hfmehmed:
It shouldn't affect user consuming ksp via gradle plugin but it will affect people depending on the ksp artifacts directly to build on top of them (basically it forces them to upgrade to jdk 17 as well).

Actually technically for gradle plugin consumers it also requires them to run gradle daemon with jdk 17 toolchain i believe - so it does affect them too in some way i guess but it seems less severe.

jaschdoc:
Closing this for now. I can reopen another PR when it's time. We can also upgrade some module independently, e.g., the integration test suite can run on Java 17 without bumping other places.

PR #2848, opened 2026-03-25
Fix type arguments for annotations

### LINKED ISSUE #2622: typeArguments returns an empty list
In my library I need to inspect generic type arguments from an annotation.
Until version `2.0.21-1.0.28` this worked fine, but starting with version `2.2.0-*` it seems broken.

**Example**
```kotlin
private fun associateWithFor(item: KSAnnotation): List<ClassName> {
    if (item.annotationType.toString() == MapConvert::class.simpleName) {
        return item.annotationType.element!!
            .typeArguments
            .take(2)
            .map { it.type!!.resolve().toClassName() }
    }
    return emptyList()
}
```
**Annotation definition**
```kotlin
@Target(AnnotationTarget.PROPERTY)
@Retention(AnnotationRetention.SOURCE)
@Repeatable
annotation class MapConvert<S: Any, D: Any, C : KOMMConverter<S, *, D, *>>(
    val converter: KClass<C>, 
    val name: String = ""
)
```
**Usage**
```kotlin
@MapConvert<SourceObject, JvmDestinationObject, JvmDestinationCostConverter>(
    converter = JvmDestinationCostConverter::class
)
```
**Problem**

- On `2.0.21-1.0.28`: `item.annotationType.element!!.typeArguments` contains the expected generic arguments (SourceObject, JvmDestinationObject).

- On `2.2.0-*` (and newer): `item.annotationType.element!!.typeArguments` is always empty.

**Expected behavior**
Generic type arguments provided to an annotation should be accessible via `annotationType.element!!.typeArguments`, just like before.

jonamireh:
## Summary
This PR fixes #2622 to allow type arguments to be read from parameterized annotations. Additionally, this PR also allows them to be read when the parameterized annotation is an argument to another annotation

## Test Plan
Compiler tests were added for all of these cases

🤖 Co-authored with Codex

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
@@ -24,10 +24,13 @@ import com.intellij.psi.PsiArrayInitializerMemberValue
 import com.intellij.psi.PsiClass
 import com.intellij.psi.impl.compiled.ClsClassImpl
 import org.jetbrains.kotlin.analysis.api.KaImplementationDetail
+import org.jetbrains.kotlin.analysis.api.KaExperimentalApi
+import org.jetbrains.kotlin.analysis.api.components.resolveCall
 import org.jetbrains.kotlin.analysis.api.annotations.KaAnnotation
 import org.jetbrains.kotlin.analysis.api.impl.base.annotations.KaBaseNamedAnnotationValue
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolOrigin
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationUseSiteTarget.*
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtFile
 
 class KSAnnotationResolvedImpl private constructor(
@@ -43,9 +46,18 @@ class KSAnnotationResolvedImpl private constructor(
             }
     }
 
+    @OptIn(KaExperimentalApi::class)
     override val annotationType: KSTypeReference by lazy {
         analyze {
-            KSTypeReferenceResolvedImpl.getCached(
+            // Try source-level type reference first (preserves type arguments from source),
+            // then fall back to resolved call signature, and finally build from classId.
+            val sourceTypeReference = lazy { (annotationApplication.psi as? KtAnnotationEntry)?.typeReference }
+            val resolvedReturnType = lazy { annotationApplication.psi?.resolveCall()?.signature?.returnType }
+            sourceTypeReference.value?.let {
+                KSTypeReferenceImpl.getCached(it, this@KSAnnotationResolvedImpl)
+            } ?: resolvedReturnType.value?.let {
+                KSTypeReferenceResolvedImpl.getCached(it, parent = this@KSAnnotationResolvedImpl)
+            } ?: KSTypeReferenceResolvedImpl.getCached(
                 buildClassType(annotationApplication.classId!!),
                 parent = this@KSAnnotationResolvedImpl
             )
--- a/kotlin-analysis-api/testData/annotationTypeArguments.kt
+++ b/kotlin-analysis-api/testData/annotationTypeArguments.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2026 Google LLC
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// WITH_RUNTIME
+// TEST PROCESSOR: AnnotationTypeArgumentsProcessor
+// EXPECTED:
+// ContainerTarget.nested.annotationType: MapConvert<SourceObject, DestinationObject, DestinationConverter>
+// ContainerTarget.nested.typeArgCount: 3
+// ContainerTarget.nested.typeArgs: SourceObject, DestinationObject, DestinationConverter
+// LibClass.annotationType: LibAnno
+// LibClass.typeArgCount: 0
+// NestedTarget.annotationType: MapConvert<INVARIANT List<INVARIANT SourceObject>, INVARIANT Map<INVARIANT String, INVARIANT DestinationObject>, INVARIANT NestedDestinationConverter>
+// NestedTarget.typeArgCount: 3
+// NestedTarget.typeArgs: kotlin.collections.List, kotlin.collections.Map, NestedDestinationConverter
+// Target.annotationType: MapConvert<INVARIANT SourceObject, INVARIANT DestinationObject, INVARIANT DestinationConverter>
+// Target.typeArgCount: 3
+// Target.typeArgs: SourceObject, DestinationObject, DestinationConverter
+// END
+
+// MODULE: lib
+// FILE: LibAnno.kt
+package com.example
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.BINARY)
+annotation class LibAnno<A : Any>
+
+// FILE: LibAnnoValue.kt
+package com.example
+
+class LibAnnoValue
+
+// FILE: LibClass.kt
+package com.example
+
+@LibAnno<LibAnnoValue>
+class LibClass
+
+// MODULE: main(lib)
+interface Converter<S : Any, D : Any>
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.SOURCE)
+annotation class MapConvert<S : Any, D : Any, C : Converter<S, D>>
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.SOURCE)
+annotation class Container(
+    val nested: MapConvert<out Any, out Any, out Converter<out Any, out Any>>,
+)
+
+class SourceObject
+class DestinationObject
+class DestinationConverter : Converter<SourceObject, DestinationObject>
+class NestedDestinationConverter : Converter<List<SourceObject>, Map<String, DestinationObject>>
+
+@MapConvert<SourceObject, DestinationObject, DestinationConverter>
+class Target
+
+@MapConvert<List<SourceObject>, Map<String, DestinationObject>, NestedDestinationConverter>
+class NestedTarget
+
+@Container(
+    nested = MapConvert<SourceObject, DestinationObject, DestinationConverter>()
+)
+class ContainerTarget
--- a/kotlin-analysis-api/testData/annotationTypeArguments.kt
+++ b/kotlin-analysis-api/testData/annotationTypeArguments.kt
@@ -21,8 +21,6 @@
 // ContainerTarget.nested.annotationType: MapConvert<SourceObject, DestinationObject, DestinationConverter>
 // ContainerTarget.nested.typeArgCount: 3
 // ContainerTarget.nested.typeArgs: SourceObject, DestinationObject, DestinationConverter
-// LibClass.annotationType: LibAnno
-// LibClass.typeArgCount: 0
 // NestedTarget.annotationType: MapConvert<INVARIANT List<INVARIANT SourceObject>, INVARIANT Map<INVARIANT String, INVARIANT DestinationObject>, INVARIANT NestedDestinationConverter>
 // NestedTarget.typeArgCount: 3
 // NestedTarget.typeArgs: kotlin.collections.List, kotlin.collections.Map, NestedDestinationConverter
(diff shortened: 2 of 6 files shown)


jonamireh:
@ting-yuan @jaschdoc I rebased with the correct email for CLA, let me know if there's anything I need!

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
```diff
@@ -43,9 +46,14 @@ class KSAnnotationResolvedImpl private constructor(
...
 
+    @OptIn(KaExperimentalApi::class)
     override val annotationType: KSTypeReference by lazy {
         analyze {
-            KSTypeReferenceResolvedImpl.getCached(
+            (annotationApplication.psi as? KtAnnotationEntry)?.typeReference?.let {
+                KSTypeReferenceImpl.getCached(it, this@KSAnnotationResolvedImpl)
+            } ?: annotationApplication.psi?.resolveCall()?.signature?.returnType?.let {
+                KSTypeReferenceResolvedImpl.getCached(it, parent = this@KSAnnotationResolvedImpl)
+            } ?: KSTypeReferenceResolvedImpl.getCached(
```
jaschdoc: Could you either name things a bit more or explain in comments what happens here?

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/AnnotationTypeArgumentsProcessor.kt
```diff
@@ -0,0 +1,57 @@
...
+        results.add(
+            "$label.typeArgs: ${
+            typeArguments.joinToString { argument ->
+                argument.type!!.resolve().declaration.qualifiedName!!.asString()
+            }
+            }"
+        )
+    }
+
+    override fun toResult(): List<String> = results
```
jaschdoc: This should be sorted to ensure results are stable across changes to the resolver. You might also need to update the expected output in `annotationTypeArguments.kt`.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/AnnotationTypeArgumentsProcessor.kt
```diff
@@ -0,0 +1,57 @@
...
+        results.add("$label.annotationType: ${annotation.annotationType}")
+        results.add("$label.typeArgCount: ${typeArguments.size}")
+        results.add(
+            "$label.typeArgs: ${
+            typeArguments.joinToString { argument ->
+                argument.type!!.resolve().declaration.qualifiedName!!.asString()
+            }
+            }"
+        )
+    }
```
jaschdoc: Nit: Could you split rendering and adding to list into separate functions or just add the list of strings to the result at the call site? Example:
```kotlin
renderAnnotation(propertyName, annotation).forEach { results.add(it) }
```

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/AnnotationTypeArgumentsProcessor.kt
```diff
@@ -0,0 +1,57 @@
...
+
+class AnnotationTypeArgumentsProcessor : AbstractTestProcessor() {
+    private val results = mutableListOf<String>()
+
+    override fun process(resolver: Resolver): List<KSAnnotated> {
+        listOf("target", "nestedTarget").forEach { propertyName ->
+            val property = resolver.getPropertyDeclarationByName(resolver.getKSNameFromString(propertyName), true)!!
+            val annotation = property.annotations.single { it.shortName.asString() == "MapConvert" }
+            renderAnnotation(propertyName, annotation)
+        }
```
jaschdoc, line 40: Ideally, you should obtain these properties by calling `resolver.getSymbolsWithAnnotation`, so if the test is ever updated, the processor will catch the change.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/AnnotationTypeArgumentsProcessor.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        val containerProperty =
+            resolver.getPropertyDeclarationByName(resolver.getKSNameFromString("containerTarget"), true)!!
+        val containerAnnotation = containerProperty.annotations.single { it.shortName.asString() == "Container" }
+        val nestedAnnotation = containerAnnotation.arguments.single().value as KSAnnotation
+        renderAnnotation("containerTarget.nested", nestedAnnotation)
```
jaschdoc: The same thing applies here. You should obtain `containerTarget` by calling `resolver.getSymbolsWithAnnotation("Container")`

jaschdoc (commented):
Hi @jonamireh, thank you for the PR! Looks good overall. I left some feedback to improve readability and tests.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/AnnotationTypeArgumentsProcessor.kt
jaschdoc: You could also change the targets to classes instead of val-bound unit values.

jonamireh:
Thanks for the feedback! Addressed them all 🙂

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
```diff
@@ -43,9 +46,18 @@ class KSAnnotationResolvedImpl private constructor(
...
     }
 
+    @OptIn(KaExperimentalApi::class)
     override val annotationType: KSTypeReference by lazy {
         analyze {
-            KSTypeReferenceResolvedImpl.getCached(
+            // Try source-level type reference first (preserves type arguments from source),
+            // then fall back to resolved call signature, and finally build from classId.
+            val sourceTypeReference = (annotationApplication.psi as? KtAnnotationEntry)?.typeReference
+            val resolvedReturnType = annotationApplication.psi?.resolveCall()?.signature?.returnType
```
jaschdoc: This is expensive to compute, so this should be moved down after `sourceTypeReference?.let { ... }`. I know this is what you had originally, but you could wrap this declaration in a lazy block.

kotlin-analysis-api/testData/annotationTypeArguments.kt
```diff
@@ -0,0 +1,66 @@
...
+// END
+
+import kotlin.reflect.KClass
+
+interface Converter<S : Any, D : Any>
+
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.SOURCE)
+annotation class MapConvert<S : Any, D : Any, C : Converter<S, D>>(
+    val converter: KClass<C>,
```
jaschdoc: Is this property required? It does not seem to be used anywhere except assigned to.

kotlin-analysis-api/testData/annotationTypeArguments.kt
```diff
@@ -0,0 +1,66 @@
...
+// Target.annotationType: MapConvert<INVARIANT SourceObject, INVARIANT DestinationObject, INVARIANT DestinationConverter>
+// Target.typeArgCount: 3
+// Target.typeArgs: SourceObject, DestinationObject, DestinationConverter
+// END
+
+import kotlin.reflect.KClass
+
+interface Converter<S : Any, D : Any>
+
+@Target(AnnotationTarget.PROPERTY)
```
jaschdoc: This should be `AnnotationTarget.CLASS`.

jaschdoc (commented):
Cool, thanks for the changes. I have a few more comments that I didn't catch earlier.

kotlin-analysis-api/testData/annotationTypeArguments.kt
jonamireh: Good catch, removed!

kotlin-analysis-api/testData/annotationTypeArguments.kt
jonamireh: Whoops. Fixed!

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/resolved/KSAnnotationResolvedImpl.kt
jonamireh: Yeup, lost that during the change. Fixed!

kotlin-analysis-api/testData/annotationTypeArguments.kt
```diff
@@ -0,0 +1,60 @@
...
+// ContainerTarget.nested.typeArgs: SourceObject, DestinationObject, DestinationConverter
+// NestedTarget.annotationType: MapConvert<INVARIANT List<INVARIANT SourceObject>, INVARIANT Map<INVARIANT String, INVARIANT DestinationObject>, INVARIANT NestedDestinationConverter>
+// NestedTarget.typeArgCount: 3
+// NestedTarget.typeArgs: kotlin.collections.List, kotlin.collections.Map, NestedDestinationConverter
+// Target.annotationType: MapConvert<INVARIANT SourceObject, INVARIANT DestinationObject, INVARIANT DestinationConverter>
+// Target.typeArgCount: 3
+// Target.typeArgs: SourceObject, DestinationObject, DestinationConverter
+// END
+
+import kotlin.reflect.KClass
```
jaschdoc: I believe this import can be removed now :)

jaschdoc (commented):
This looks good. I have one last comment and one more consideration. Apologies for the slow back and forth!

Do you know how this interacts with compiled libraries? I assume the type arguments would be erased. Could you add another section to the test that checks this?
If you add the following declaration `interface Converter`, i.e.,:
```kt
// MODULE: lib
// FILE: LibAnno.kt

@Target(AnnotationTarget.CLASS)
@Retention(AnnotationRetention.BINARY)
annotation class LibAnno<A : Any>

// FILE: LibAnnoValue.kt
package com.example

class LibAnnoValue

// FILE: LibClass.kt
package com.example

@LibAnno<LibAnnoValue>
class LibClass

// MODULE: main(lib)

interface Converter<S : Any, D : Any>
// The rest of your current test
```
This will create a library / jar with the new parts. Note that you can also move the library part below you current test program if you think that's prettier. Then the symbol processor would have to ask for a class with name `com.example.LibClass`.

Oh and it would be great if you could fixup / squash the commits that address feedback into the previous commits.

jonamireh:
>Do you know how this interacts with compiled libraries? 

It seems like they can't be resolved? The processor reports no type arguments for annotations on classes that have been compiled

jaschdoc:
My bad sorry, the annotation retention should be `RUNTIME` instead of `BINARY`, so it's available for reflection. Maybe it's good to actually cover [all three retention cases](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.annotation/-annotation-retention) in libraries. It makes sense that for `SOURCE` and `BINARY`, they are completely erased, but for `RUNTIME` there should at least be something. Maybe that's an erased type, i.e., `Object`.

jonamireh:
> It makes sense that for `SOURCE` and `BINARY`, they are completely erased, but for `RUNTIME` there should at least be something. Maybe that's an erased type, i.e., `Object`.

Unfortunately, no. It doesn't seem to matter whether it's `RUNTIME` or `BINARY`, KSP still can't seem to read the type arguments of `LibAnno` once it's applied and compiled. I think this makes sense to me? Of course, when it's `SOURCE`, it can't even detect the annotation on compiled source

jaschdoc:
> > It makes sense that for `SOURCE` and `BINARY`, they are completely erased, but for `RUNTIME` there should at least be something. Maybe that's an erased type, i.e., `Object`.
> 
> Unfortunately, no. It doesn't seem to matter whether it's `RUNTIME` or `BINARY`, KSP still can't seem to read the type arguments of `LibAnno` once it's applied and compiled. I think this makes sense to me? Of course, when it's `SOURCE`, it can't even detect the annotation on compiled source

Yes, you're right. I don't know why I wrote `BINARY` when I meant to write `RUNTIME`, apologies. With `RUNTIME` annotation retention, it's part of the `RuntimeVisibleAnnotations` in the byte code so it's both stored in the byte code and also publicly visible for reflection. KSP should be able to at least load in the annotation, but it's unclear what the value for the type argument is. Usually, type arguments are not visible in class files, so I suspect it's the type at the declaration site of the annotation.

jonamireh:
> Yes, you're right. I don't know why I wrote `BINARY` when I meant to write `RUNTIME`, apologies. With `RUNTIME` annotation retention, it's part of the `RuntimeVisibleAnnotations` in the byte code so it's both stored in the byte code and also publicly visible for reflection. KSP should be able to at least load in the annotation, but it's unclear what the value for the type argument is. Usually, type arguments are not visible in class files, so I suspect it's the type at the declaration site of the annotation.

Is there an action you'd like me to take to move this PR forward? The PR is intended for type arguments at a use-site, not a declaration site

jaschdoc:
> Is there an action you'd like me to take to move this PR forward? The PR is intended for type arguments at a use-site, not a declaration site

Yes, but the use-site type arguments in libraries are important to cover if this feature is to be supported. It'd be great if you added another test file (and perhaps also test processor) that covers the library case with all three retention cases. Of course, as discussed, the expected output is no output for annotations with retention `SOURCE` and `BINARY`, but for the annotation with retention `RUNTIME`, we at least expect the annotation to be visible but for the type arguments its unclear so I suggest we just see what the test produces and we discuss from there.

I know there's been a lot back and forth. I appreciate the patience and work and this a good use case to cover, but I want to do so on solid ground. Once we have those tests, it should be close to ready! Cheers :)

jonamireh:
> It'd be great if you added another test file (and perhaps also test processor) that covers the library case with all three retention cases. Of course, as discussed, the expected output is no output for annotations with retention `SOURCE` and `BINARY`, but for the annotation with retention `RUNTIME`, we at least expect the annotation to be visible but for the type arguments its unclear so I suggest we just see what the test produces and we discuss from there.

Done!

jaschdoc:
Awesome, thanks! I'm holding off moving forward with this until the discussion in #2622 is clearer.

jonamireh:
@jaschdoc It seems _something_ was merged upstream that already fixes the `NestedTarget.typeArgs` and `Target.typeArgs` tests. Do you know what it was? I did check – all of these tests failed when the PR was opened

However, this PR still fixes the nested case where an annotation accepts another parameterized annotation as an argument (so `ContainerTarget.nested.typeArgs` in these tests) and [my usecase](https://github.com/google/ksp/issues/2622#issuecomment-4359554231)

jaschdoc:
Hi again @jonamireh, are you still interested in contributing this? I know there was a bit of back and forth on the issue, but if you add a test that clearly shows the defect in a separate issue and PR then it might be easier to move forward with this. Cheers :)

PR #2850, opened 2026-03-26
Update tests with anonymous classes

jaschdoc:
@ting-yuan are we sure this is the right behavior? It seems to me that when `inDepth = true`, the processor should pick up the `Anno` annotations on the `run` methods of the anonymous classes. Note that I have not added the newly annotated symbols to the expected test results (only updated line numbers in `locations.kt`), yet the tests still pass.

Related to https://github.com/google/ksp/pull/2846

### DIFF
--- a/test-utils/testData/api/getSymbolsFromAnnotation.kt
+++ b/test-utils/testData/api/getSymbolsFromAnnotation.kt
@@ -47,6 +47,14 @@
 // p1:KSValueParameter
 // param:KSValueParameter
 // propertyFoo:KSPropertyDeclaration
+// run:KSFunctionDeclaration
+// run:KSFunctionDeclaration
+// run:KSFunctionDeclaration
+// run:KSFunctionDeclaration
+// runJava:KSFunctionDeclaration
+// runJava2:KSFunctionDeclaration
+// runKotlin:KSFunctionDeclaration
+// runKotlin2:KSFunctionDeclaration
 // ==== Bnno inDepth = false ====
 // <init>:KSFunctionDeclaration
 // File: Foo.kt:KSFile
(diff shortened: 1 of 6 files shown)


test-utils/testData/api/locations.kt
```diff
@@ -76,6 +86,37 @@ class K<@Location T> {
...
+    private val backing = 42
+
+    @delegate:Location
+    val delegateProp: Int by backing
+
+    var setParamProp: Int = 0
+        set(@setparam:Location value) { field = value }
 }
+
+fun @receiver:Location String.extFun() = Unit
```
jaschdoc, line 122: This is just a copy from `kotlin-analysis-api/testData/locations.kt`

ting-yuan (dismissed):
`inDepth=true` is rarely used and unsurprisingly there are bugs.

btw, one more future, refactoring task is to merge the two `testData` directory. They were forked because of KSP1 and KSP2.

test-utils/testData/api/getSymbolsFromAnnotation.kt
jaschdoc, line 1: This file is a copy of `kotlin-analysis-api/testData/getSymbolsFromAnnotation.kt`

jaschdoc:
@ting-yuan I gave it some more thought. I'm not sure anonymous classes should be supported. They are sort of like lambdas; it's an _expression_ that immediately declares and returns a value. However, nested functions with proper names are declarations, since the declaration does not evaluate to a value whereas constructing a lambda is directly an expression that returns a value. The same goes for anonymous classes and objects. **Thus, I believe it's correct to not return any annotations on members of anonymous classes / obejcts.**

jaschdoc:
This PR would introduce a bug.

PR #2847, opened 2026-03-24
Add lazy caching in `PsiResolutionStrategy`

jaschdoc:
Related to https://github.com/google/ksp/issues/2816

Issue #2816, opened 2026-02-27
[Performance]: getSymbolsWithAnnotation resolves symbols too eagerly.


bcorso:
I'm seeing [`ResolverAAImpl#getSymbolsWithAnnotation()`](https://github.com/google/ksp/blob/main/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt#L546) take ~50% of the processing time in some cases (see numbers below). Note that calling `getSymbolsWithAnnotation` is literally the first thing a processor does before doing any work itself, so the fact that this call alone takes 50% of the processing time is substantial.

It looks like the main culprit is [`CollectAnnotatedSymbolsVisitor`](https://github.com/google/ksp/blob/main/common-util/src/main/kotlin/com/google/devtools/ksp/common/visitor/CollectAnnotatedSymbolsVisitor.kt), which is used to walk the AST and find which symbols are annotated. However, this visitor requires using the Analysis API to first resolve each symbol to a `KaSymbol` and then wrap in `KSAnnotated`, which is very expensive.

It should be much cheaper to use a psi visitor, e.g. [`PsiRecursiveElementWalkingVisitor`](https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/psi/PsiRecursiveElementWalkingVisitor.java) to find the elements that are annotated. The PsiElement representation should be much cheaper because it's purpose is just to represent the structure of the source file and avoids expensive type resolution/validation that is needed for `KaSymbol`.

However, one issue with a pure psi visitor approach is that it doesn't take into account type aliases, e.g. if an element is annotated with `@MyAlias` which is an alias of `@MyAnnotation`, calling `resolver.getSymbolsWithAnnotation("MyAnnotation")` wouldn't return symbols annotated with `@MyAlias` because it doesn't take into account type aliases. In order to fix this, we would need to use the Analysis API to resolve every annotation to fully expand its type aliases. However, this should still be much, much cheaper than resolving every symbol in the AST.

I've created a prototype of both the pure psi approach (without type alias support) and the "fixed" approach (with type alias support). The performance comparisons are shown below:

* **Without optimization:**
  * total: `16.5s`
  * getSymbolsWithAnnotation: `8.67s`
* **With PsiElement optimization with type alias support:**
  * total: `8.39s`
  * getSymbolsWithAnnotation: `1.84s`
* **With PsiElement optimization without type alias support**
  * total: `6.98s`
[...]

PR #2860, opened 2026-04-08
Lazily compute fully qualified Java annotation names

jaschdoc:
Depends on https://github.com/google/ksp/pull/2847

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
@@ -75,6 +75,9 @@ class PsiResolutionStrategy(
     override val deferredSymbols: Map<SymbolProcessor, List<Restorable>>
 ) : AnnotationResolutionStrategy {
 
+    /**
+     * Returns all symbols annotated with [annotationName].
+     */
     override fun getSymbolsWithAnnotation(annotationName: String, inDepth: Boolean): Sequence<KSAnnotated> =
         if (inDepth)
             aaResolutionStrategy.getSymbolsWithAnnotation(annotationName, inDepth = true)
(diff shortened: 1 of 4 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -75,11 +75,16 @@ class PsiResolutionStrategy(
...
+     * Returns all symbols annotated with [annotationName].
+     */
     override fun getSymbolsWithAnnotation(annotationName: String, inDepth: Boolean): Sequence<KSAnnotated> =
         if (inDepth)
             aaResolutionStrategy.getSymbolsWithAnnotation(annotationName, inDepth = true)
-        else
-            getAnnotatedSymbols(annotationName)
+        else {
+            val result = getAnnotatedSymbols(annotationName)
+            result
```
bcorso: remove variable?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -92,35 +97,104 @@ class PsiResolutionStrategy(
...
      */
-    private fun getAnnotatedSymbols(annotationName: String): Sequence<KSAnnotated> {
-        val newSymbols = annotationToSymbolsCache[annotationName]?.value ?: emptyList()
-        val restoredSymbols = deferredSymbolsGroupedByAnnotations[annotationName] ?: emptyList()
-        // N.B.: Convert `newSymbols` to sequence **before** appending `restoredSymbols`
-        // to avoid eagerly appending lists.
-        return newSymbols.asSequence() + restoredSymbols
+    private fun getAnnotatedJavaSymbols(annotationName: String): Sequence<KSAnnotated> {
+        val annotationShortName = annotationName.substringAfterLast('.')
+        val result = annotatedJavaElementsByShortName[annotationShortName]?.filter { element ->
```
bcorso: Hmm, can we add a `fullyQualifiedJavaAnnotationNamesByElements` cache that maps from fqn to `KSAnnotated`?

For example,

```kotlin
private val annotatedJavaElementsByFullyQualifiedName
    : MutableMap<String, Lazy<Collection<KSAnnotated>>>
```

The map would need to be mutable and each key calculated lazily, but I think this would be a better way to organize it since it mimics the map you already have on the Kotlin side already:

```kotlin
private val annotatedKotlinElementsByFullyQualifiedName
    : Map<String, Lazy<Collection<KSAnnotated>>> by lazy { ... }
```

In addition, this approach would cache the call to `PsiElement.resolveToKSAnnotated()`, which is also expensive, so that it's not called again on subsequent calls to `getAnnotatedJavaSymbols`.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -402,6 +454,14 @@ class PsiResolutionStrategy(
         return listOf(typeParamSym, clone)
     }
 
+    /**
+     * Returns the short name used at the annotation site.
+     */
+    private val PsiAnnotation.shortName: String
+        get() = text.substringAfterLast('@') // Remove annotation delimiter
```
bcorso: String parsing is super fragile. For example, this would fail for cases like 

```java
@MyAnnotation(value = "user@domain.com")
```

Can you just call `nameReferenceElement?.referenceName` instead? This should give you the short name directly and avoid any sort of string parsing.

bcorso (changes requested):
Thanks!

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
jaschdoc: Ah thanks! I was hesitant to commit this too, but I'm still not too familiar with the API so I wasn't sure what to do here.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
jaschdoc: Good point! I forgot about storing the results after resolving to `KSAnnotated`. However, I believe the mutable map for Java elements should not have lazy collections as it populated after resolving the relevant elements to `KSAnnotated`. Thus, at the point where we update the map, we have already computed what the lazy block would try to suspend.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -136,50 +183,78 @@ class PsiResolutionStrategy(
...
         }
+    }
 
-        else -> error("Unexpected PsiElement at ${element.toLocation()}: ${element.javaClass}")
+    /**
+     * A map from Java elements to their annotations' fully qualified names, indexed by short name.
+     */
+    private val fullyQualifiedJavaAnnotationNamesByElements: Map<PsiElement, Map<String, Lazy<String>>> by lazy {
+        annotatedJavaElements.associate { element ->
+            element to element.annotations.associate {
```
bcorso: Note that short names are not unique, so you can have multiple short names mapping to different fully qualified names. For example, if you have nested annotations like below, both annotations will have the same short name `Bar` but will map to two different fully qualified names:

```kotlin
import com.foo.Foo1
import com.foo.Foo2

@Foo1.Bar
@Foo2.Bar
val prop: MyProp
```

Thus, I think you'll need to change the map's value from `Lazy<String>` to `Lazy<Set<String>>`

It's probably be good to add a test case for this as well.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -400,6 +472,25 @@ class PsiResolutionStrategy(
         return listOf(typeParamSym, clone)
     }
 
+    /**
+     * Returns the short name used at the annotation site.
+     */
+    private val PsiAnnotation.shortName: String?
```
bcorso: Do we need both `shortName` and `shortNameUnsafe`? It seems you probably don't want anyone calling the nullable version so maybe just merge the two into a single function.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -92,35 +95,79 @@ class PsiResolutionStrategy(
...
+        }?.flatMap {
+            it.resolveToKSAnnotated()
+        } ?: emptyList()
+    }
+
+    /**
+     * Returns all [PsiElement]s that are annotated in [file] except for declarations
+     * in function bodies or property accessors.
+     */
+    private fun collectAnnotatedPsiElementsIn(file: KSFile): Collection<PsiElement> {
```
bcorso, line 163: style nit: can you move this directly below it's usage in `annotatedPsiElements` (It's typically easier to follow the flow of the code when things read in order from top down).

bcorso (approved):
Left a few comments, but overall LGTM. Thanks!

ting-yuan (dismissed):

troelsbjerre (approved):

PR #2862, opened 2026-04-09
Handle missing cases in `CollectAnnotatedSymbolsPsiVisitor`

jaschdoc:
I studied the structure of `PsiAnnotation` a bit further and found what is hopefully an exhaustive solution.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
@@ -11,9 +11,12 @@ import com.intellij.psi.PsiMethod
 import com.intellij.psi.PsiModifierList
 import com.intellij.psi.PsiPackageStatement
 import com.intellij.psi.PsiRecursiveElementWalkingVisitor
+import com.intellij.psi.PsiType
+import com.intellij.psi.PsiTypeElement
 import com.intellij.psi.PsiTypeParameter
 import com.intellij.psi.javadoc.PsiDocComment
 import org.jetbrains.kotlin.kdoc.psi.api.KDoc
+import org.jetbrains.kotlin.psi.KtAnnotatedExpression
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtDeclarationModifierList
 import org.jetbrains.kotlin.psi.KtExpression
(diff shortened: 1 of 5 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
```diff
@@ -36,31 +38,48 @@ class CollectAnnotatedSymbolsPsiVisitor : PsiRecursiveElementWalkingVisitor() {
...
                 }
-                is PsiModifierList -> {
-                    result.add(parent.parent)
+
+                is PsiTypeElement -> {
+                    result.add(owner)
                     return
                 }
+
+                null ->
```
bcorso: Why not crash for `null` and `PsiType`?

Do you have examples what code leads to these cases?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
```diff
@@ -36,31 +38,48 @@ class CollectAnnotatedSymbolsPsiVisitor : PsiRecursiveElementWalkingVisitor() {
...
-            //   PsiAnnotation -> Annotated Element
-            // For all other annotated elements, we expect the structure:
-            //   PsiAnnotation -> PsiModifierList -> Annotated Element
-            is PsiAnnotation -> when (val parent = element.parent) {
+            // For Java sources, we expect annotations to have three types of owners:
+            //  - PsiAnnotation -> PsiModifierList -> Annotated Element
+            //  - PsiAnnotation -> PsiTypeParameter (is itself annotated)
+            //  - PsiAnnotation -> PsiTypeElement (is itself annotated)
+            is PsiAnnotation -> when (val owner = element.owner) {
+                is PsiModifierList -> {
```
bcorso, line 78: Can we add an example source code that leads to each of these cases, e.g. this was generated by AI:

```
is PsiAnnotation -> when (val owner = element.owner) {
    // e.g. @MyAnnotation class Foo {}
    // e.g. @MyAnnotation String myField;
    // e.g. void myMethod(@MyAnnotation int param)
    is PsiModifierList -> {...}

    // e.g. class Foo<@MyAnnotation T> {}
    is PsiTypeParameter -> {...}

    // Type-use annotations
    // e.g. List<@MyAnnotation String> foo;
    // e.g. class Foo extends @MyAnnotation Bar {}
    is PsiTypeElement -> {...}

    // Malformed/incomplete code, e.g. a dangling "@MyAnnotation" at the end of a file
    null -> {...}

    // Programmatic/synthetic types (PsiType is not a PsiElement, so this is rare in a PSI visitor,
    // but it implements PsiAnnotationOwner and can theoretically own annotations in memory).
    is PsiType -> {...}

    else -> ...
}
```

bcorso (approved):
LGTM, thanks!

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
bcorso: In particular, I'm curious if you've actually hit issues with this or if this is just to preemptively handle these cases in case they happen. Note that I don't think we technically need to handle every case exhaustively since our visitor itself skips a lot cases we don't care about. Especially, after https://github.com/google/ksp/pull/2863 where you improve the skip logic, I would be curious if we need to handle `null` and `PsiType`. If we don't have explicit failure cases for these then I would lean towards just crashing until someone reports a case and we can investigate the specific issue.

ting-yuan (dismissed):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
jaschdoc: This case was observed when returning an annotation as a default value for another annotation, e.g.,
```java
public @interface JavaAnnotationWithDefaults {
    OtherAnnotation otherAnnotationVal() default @OtherAnnotation("def");
}
```

See [`annotationValue_java.kt`](https://github.com/google/ksp/blob/94987c34c5f446f7a1255ebdedda8a44adcff11f/kotlin-analysis-api/testData/annotationValue/annotationValue_java.kt#L158). In this example, the default value for `otherAnnotationVal` is `@OtherAnnotation("def")`

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
jaschdoc: Sounds good. I will update each case.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
bcorso: Thanks!

bcorso (approved):

ting-yuan (approved):

PR #2871, opened 2026-04-16
Bump AA to `2.4.20-dev-835`

jaschdoc:
This PR bumps the analysis api to `2.4.20-dev-835`. Since [KEEP 402](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0402-annotation-target-in-properties.md) is enabled by default in `2.4.20`, some tests need to be updated.

PR #2872, opened 2026-04-16
Update test for annotations in dependencies

jaschdoc:
Adds line numbers to the test output so changes can be debugged more easily.

Related to https://github.com/google/ksp/pull/2871


kotlin-analysis-api/testData/annotationInDependencies.kt
```diff
@@ -19,72 +19,72 @@
...
-// parameter constructorParam : annotations.AllTarget{[value = all:]}
-// parameter constructorParam : annotations.NoTargetAnnotation{[value = onConstructorParam]}
-// parameter value : annotations.AllTarget{[value = all:]}
-// property constructorParam : annotations.AllTarget{[value = all:]}
-// property constructorParam : annotations.AllTarget{[value = all:]}
-// property constructorParam : annotations.FieldTarget2{[value = field:]}
-// property constructorParam : annotations.FieldTarget{[value = onConstructorParam]}
-// property constructorParam : annotations.PropertyTarget{[value = onConstructorParam]}
-// setter of property constructorParam : annotations.PropertySetterTarget{[value = set:]}
+// class lib.DataClass <no line> : annotations.ClassTarget{[value = onDataClass : <no line>]} : <no line>
```
troelsbjerre, line 75: Why doesn't this have a line number?

jaschdoc: That is because it is declared in a jar / library, so there is no source line available for it. Tests that begin with `// MODULE: <module name>` are compiled before KSP gets access to it. Only the `main` module is available as source files for KSP. There are also some symbols that do not have source line numbers / source positions because they are synthetic / generated. This goes for the `value` symbol which is targeted by the `setparam` use site target as seen in line 208. The `value` symbol is the parameter name for the parameter in the setter method.

troelsbjerre (approved):

PR #2877, opened 2026-04-21
Add PSI element implementations of KSFunctionDeclarationImpl and KSPropertyDeclarationJavaImpl

bcorso:
This change adds a PSI element implementation of `KSFunctionDeclarationImpl` and `KSPropertyDeclarationJavaImpl` to work around performance bugs in the Analysis API for java source/class file declaration resolution (https://youtrack.jetbrains.com/issue/KT-85692).

To use the new implementation, enable the `experimentalPsiResolution` flag.

Bug: https://youtrack.jetbrains.com/issue/KT-85692

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
@@ -86,6 +86,7 @@ import com.google.devtools.ksp.isProtected
 import com.google.devtools.ksp.isPublic
 import com.google.devtools.ksp.isVisibleFrom
 import com.google.devtools.ksp.processing.KSBuiltIns
+import com.google.devtools.ksp.processing.KSPConfig
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.symbol.ClassKind
 import com.google.devtools.ksp.symbol.KSAnnotated
@@ -159,21 +160,30 @@ class ResolverAAImpl(
 ) : Resolver {
     companion object {
         private val instance_prop: ThreadLocal<ResolverAAImpl> = ThreadLocal()
-        private val ktModule_prop: ThreadLocal<KaSourceModule> = ThreadLocal()
         var instance: ResolverAAImpl
             get() = instance_prop.get()
             set(value) {
                 instance_prop.set(value)
             }
+
+        private val ktModule_prop: ThreadLocal<KaSourceModule> = ThreadLocal()
         var ktModule: KaSourceModule
             get() = ktModule_prop.get()
             set(value) {
                 ktModule_prop.set(value)
             }
 
+        private val kspConfig_prop: ThreadLocal<KSPConfig> = ThreadLocal()
+        var kspConfig: KSPConfig
+            get() = kspConfig_prop.get()
+            set(value) {
+                kspConfig_prop.set(value)
+            }
+
         fun tearDown() {
             instance_prop.remove()
             ktModule_prop.remove()
+            kspConfig_prop.remove()
         }
     }
 
@@ -583,6 +593,12 @@ class ResolverAAImpl(
 
     // TODO: handle library symbols
     override fun getJvmName(declaration: KSFunctionDeclaration): String {
+        // FAST PATH: Java methods do not have Kotlin name mangling or @JvmName.
+        // Their JVM name is always exactly their simple name.
+        if (declaration.origin == Origin.JAVA || declaration.origin == Origin.JAVA_LIB) {
+            return declaration.simpleName.asString()
+        }
+
         val symbol: KaFunctionSymbol? = when (declaration) {
             is KSFunctionDeclarationImpl -> declaration.ktFunctionSymbol
             else -> null
@@ -814,10 +830,6 @@ class ResolverAAImpl(
         fun KSType.toSignature(): String {
             return if (this is KSTypeImpl) {
                 analyze {
-                    val decl = (this@toSignature.declaration as? KSClassDeclaration)
-                    // Force inline value class to be unwrapped.
-                    // Do not remove until AA fixes this.
-                    decl?.primaryConstructor
                     type.toSignature()
                 }
             } else {
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
@@ -21,6 +21,7 @@ import com.google.devtools.ksp.common.KSObjectCache
 import com.google.devtools.ksp.common.errorTypeOnInconsistentArguments
 import com.google.devtools.ksp.common.impl.KSNameImpl
 import com.google.devtools.ksp.common.impl.KSTypeReferenceSyntheticImpl
+import com.google.devtools.ksp.common.isObjectOverride
 import com.google.devtools.ksp.common.lazyMemoizedSequence
 import com.google.devtools.ksp.impl.ResolverAAImpl
 import com.google.devtools.ksp.impl.recordGetSealedSubclasses
@@ -29,17 +30,25 @@ import com.google.devtools.ksp.impl.recordLookupForGetAllFunctions
 import com.google.devtools.ksp.impl.recordLookupForGetAllProperties
 import com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSTypeReferenceResolvedImpl
 import com.google.devtools.ksp.symbol.*
+import com.intellij.psi.PsiClass
 import org.jetbrains.kotlin.analysis.api.KaExperimentalApi
 import org.jetbrains.kotlin.analysis.api.KaImplementationDetail
 import org.jetbrains.kotlin.analysis.api.impl.base.types.KaBaseStarTypeProjection
 import org.jetbrains.kotlin.analysis.api.symbols.*
 import org.jetbrains.kotlin.analysis.api.types.abbreviationOrSelf
+import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.KtClassOrObject
 
 class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSymbol: KaClassSymbol) :
     KSClassDeclaration,
-    AbstractKSDeclarationImpl(ktClassOrObjectSymbol),
+    AbstractKSDeclarationImpl(),
     KSExpectActual by KSExpectActualImpl(ktClassOrObjectSymbol) {
+    private val isExperimentalPsiResolutionEnabled: Boolean by lazy {
+        ResolverAAImpl.kspConfig.experimentalPsiResolution
+    }
+
+    override val ktDeclarationSymbol = ktClassOrObjectSymbol
+
     companion object : KSObjectCache<KaClassSymbol, KSClassDeclarationImpl>() {
         fun getCached(ktClassOrObjectSymbol: KaClassSymbol) =
             cache.getOrPut(ktClassOrObjectSymbol) { KSClassDeclarationImpl(ktClassOrObjectSymbol) }
@@ -180,8 +189,88 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
     }
 
     override val declarations: Sequence<KSDeclaration> by lazyMemoizedSequence {
+        // FAST PATH: Use PSI directly for Java classes
+        if (isExperimentalPsiResolutionEnabled) {
+            ktClassOrObjectSymbol.psi?.let { psi ->
+                if (psi is PsiClass && canUsePsiResolution()) {
+                    return@lazyMemoizedSequence psi.getDeclarationsFromPsi()
+                }
+            }
+        }
+
+        // SLOW PATH: Original AA fallback for Kotlin classes (or if ASM failed)
+        getDeclarationsFromAnalysisApi()
+    }
+
+    private fun canUsePsiResolution(): Boolean {
+        return (origin == Origin.JAVA || origin == Origin.JAVA_LIB) &&
+            // Java annotations, enums, and types that map to Kotlin types (e.g. java.lang.String -> kotlin.String) have
+            // additional properties added by the Analysis API that don't actually exist in the Java type so just skip
+            // them to avoid this complexity.
+            classKind != ClassKind.ANNOTATION_CLASS &&
+            classKind != ClassKind.ENUM_CLASS &&
+            !ktClassOrObjectSymbol.isOrInheritsFromKotlinAlteredType()
+    }
+
+    private fun PsiClass.getDeclarationsFromPsi(): Sequence<KSDeclaration> {
+        // 1. Fields
+        val psiFields = fields.asSequence()
+            .map { KSPropertyDeclarationJavaImpl.getCached(psi = it, parent = this@KSClassDeclarationImpl) }
+
+        // 2. Methods (Non-constructors)
+        val psiMethods = methods.asSequence()
+            // Mimic FIR dropping malformed constructors that don't match class name, e.g. `class Foo { Bar() {} }`
+            .filter { !it.isConstructor }
+            // Mimic FIR dropping Object overrides in Java interfaces (e.g. `@Override boolean equals(Object)`).
+            .filter { !(isInterface && it.isObjectOverride()) }
+            .map { KSFunctionDeclarationImpl.getCached(psi = it, parent = this@KSClassDeclarationImpl) }
+
+        // 3. Constructors
+        // IntelliJ's PSI lacks implicit constructors. If none are explicitly declared,
+        // we must fetch the synthesized constructor from the Analysis API.
+        val psiConstructors = if (constructors.isEmpty() && !isInterface) {
+            analyze {
+                ktClassOrObjectSymbol.declaredMemberScope.constructors.map {
+                    KSFunctionDeclarationImpl.getCached(it)
+                }
+            }
+        } else {
+            constructors.asSequence()
+                .map { KSFunctionDeclarationImpl.getCached(psi = it, parent = this@KSClassDeclarationImpl) }
+        }
+
+        // 4. Inner Classes
+        val psiInnerClasses = innerClasses.asSequence()
+            .mapNotNull { psiInnerClass ->
+                analyze {
+                    val name = psiInnerClass.name?.let { Name.identifier(it) } ?: return@analyze null
+                    val instanceSymbols = ktClassOrObjectSymbol.declaredMemberScope.classifiers(name)
+                    val staticSymbols = ktClassOrObjectSymbol.staticDeclaredMemberScope.classifiers(name)
+                    (instanceSymbols + staticSymbols)
+                        .filterIsInstance<KaNamedClassSymbol>()
+                        .find { it.psi == psiInnerClass || it.psi?.isEquivalentTo(psiInnerClass) == true }
+                        ?.let { KSClassDeclarationImpl.getCached(it) }
+                }
+            }
+
+        // 5. Partition members into static and instance members to emulate FIR's ordering.
+        val (psiStaticMembers, psiInstanceMembers) = (psiFields + psiMethods + psiInnerClasses).partition {
+            Modifier.JAVA_STATIC in it.modifiers ||
+                // Nested classes in KSP don't get the static modifier.
+                (it is KSClassDeclaration && Modifier.INNER !in it.modifiers)
+        }
+
+        // Note: We could return declarations in source order by iterating over PsiClass.declarations directly.
+        // However, this ordering matches how things are done using the Analysis API to keep things consistent.
+        return psiInstanceMembers.asSequence() + psiConstructors + psiStaticMembers.asSequence()
+    }
+
+    private fun getDeclarationsFromAnalysisApi(): Sequence<KSDeclaration> {
         val decls = ktClassOrObjectSymbol.declarations()
-        if (origin == Origin.JAVA && classKind != ClassKind.ANNOTATION_CLASS) {
+            // The Analysis API is known to leak certain inherited members, so we filter those cases out here.
+            .filter { !it.isLeakedInheritedMember(this) }
+
+        return if ((origin == Origin.JAVA || origin == Origin.JAVA_LIB) && classKind != ClassKind.ANNOTATION_CLASS) {
             decls.flatMap { decl ->
                 if (decl is KSPropertyDeclarationImpl && decl.ktPropertySymbol is KaSyntheticJavaPropertySymbol) {
                     sequenceOf(decl.getter, decl.setter).mapNotNull { accessor ->
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
@@ -22,9 +22,11 @@ import com.google.devtools.ksp.closestClassDeclaration
 import com.google.devtools.ksp.common.KSObjectCache
 import com.google.devtools.ksp.common.impl.KSNameImpl
 import com.google.devtools.ksp.common.lazyMemoizedSequence
+import com.google.devtools.ksp.common.toKSModifiers
 import com.google.devtools.ksp.impl.ResolverAAImpl
 import com.google.devtools.ksp.impl.recordLookupForPropertyOrMethod
 import com.google.devtools.ksp.impl.recordLookupWithSupertypes
+import com.google.devtools.ksp.impl.symbol.java.KSAnnotationJavaImpl
 import com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSTypeReferenceResolvedImpl
 import com.google.devtools.ksp.isConstructor
 import com.google.devtools.ksp.isPublic
@@ -33,10 +35,11 @@ import com.google.devtools.ksp.symbol.FunctionKind
 import com.google.devtools.ksp.symbol.KSAnnotation
 import com.google.devtools.ksp.symbol.KSClassDeclaration
 import com.google.devtools.ksp.symbol.KSDeclaration
-import com.google.devtools.ksp.symbol.KSExpectActual
+import com.google.devtools.ksp.symbol.KSFile
 import com.google.devtools.ksp.symbol.KSFunction
 import com.google.devtools.ksp.symbol.KSFunctionDeclaration
 import com.google.devtools.ksp.symbol.KSName
+import com.google.devtools.ksp.symbol.KSPropertyDeclaration
 import com.google.devtools.ksp.symbol.KSType
 import com.google.devtools.ksp.symbol.KSTypeReference
 import com.google.devtools.ksp.symbol.KSValueParameter
@@ -45,6 +48,8 @@ import com.google.devtools.ksp.symbol.Modifier
 import com.google.devtools.ksp.symbol.Origin
 import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiMethod
+import com.intellij.psi.PsiModifier
+import org.jetbrains.kotlin.analysis.api.KaSession
 import org.jetbrains.kotlin.analysis.api.symbols.KaConstructorSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaDestructuringDeclarationSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaFunctionSymbol
@@ -52,6 +57,8 @@ import org.jetbrains.kotlin.analysis.api.symbols.KaNamedFunctionSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaPropertyAccessorSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaPropertyGetterSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaPropertySetterSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaPropertySymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolLocation
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolModality
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolOrigin
@@ -59,22 +66,36 @@ import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolVisibility
 import org.jetbrains.kotlin.analysis.api.symbols.name
 import org.jetbrains.kotlin.analysis.api.symbols.receiverType
 import org.jetbrains.kotlin.analysis.api.types.abbreviationOrSelf
+import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.KtDeclaration
 import org.jetbrains.kotlin.psi.KtFunction
 
-class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbol: KaFunctionSymbol) :
-    KSFunctionDeclaration,
-    AbstractKSDeclarationImpl(ktFunctionSymbol),
-    KSExpectActual by KSExpectActualImpl(ktFunctionSymbol) {
-    companion object : KSObjectCache<KaFunctionSymbol, KSFunctionDeclarationImpl>() {
-        fun getCached(ktFunctionSymbol: KaFunctionSymbol) =
-            cache.getOrPut(ktFunctionSymbol) { KSFunctionDeclarationImpl(ktFunctionSymbol) }
+sealed class KSFunctionDeclarationImpl : KSFunctionDeclaration, AbstractKSDeclarationImpl() {
+    companion object {
+        // Factory for AA Slow Path
+        fun getCached(symbol: KaFunctionSymbol): KSFunctionDeclarationImpl =
+            KSFunctionDeclarationAAImpl.getCached(symbol)
+
+        // Factory for PSI Fast Path
+        fun getCached(psi: PsiMethod, parent: KSClassDeclarationImpl): KSFunctionDeclarationImpl =
+            KSFunctionDeclarationPsiImpl.getCached(psi, parent)
     }
+
+    abstract val ktFunctionSymbol: KaFunctionSymbol
+
+    override val ktDeclarationSymbol: KaFunctionSymbol get() = ktFunctionSymbol
+
+    // Manual delegation for KSExpectActual to avoid eager evaluation in the class header
+    private val expectActualImpl by lazy { KSExpectActualImpl(ktFunctionSymbol) }
+    override val isActual: Boolean get() = expectActualImpl.isActual
+    override val isExpect: Boolean get() = expectActualImpl.isExpect
+    override fun findActuals(): Sequence<KSDeclaration> = expectActualImpl.findActuals()
+    override fun findExpects(): Sequence<KSDeclaration> = expectActualImpl.findExpects()
 
     override val functionKind: FunctionKind by lazy {
         when (ktFunctionSymbol.location) {
             KaSymbolLocation.CLASS -> {
-                if (ktFunctionSymbol is KaNamedFunctionSymbol && ktFunctionSymbol.isStatic) {
+                if ((ktFunctionSymbol as? KaNamedFunctionSymbol)?.isStatic == true) {
                     FunctionKind.STATIC
                 } else {
                     FunctionKind.MEMBER
@@ -171,7 +192,7 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
     }
 
     override val simpleName: KSName by lazy {
-        when (ktFunctionSymbol) {
+        when (val ktFunctionSymbol = this.ktFunctionSymbol) {
             is KaNamedFunctionSymbol -> KSNameImpl.getCached(ktFunctionSymbol.name.asString())
             is KaPropertyAccessorSymbol -> when (val psi = ktFunctionSymbol.psi) {
                 is PsiMethod -> KSNameImpl.getCached(psi.name)
@@ -286,6 +307,175 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
     val isSetter: Boolean by lazy {
         ktFunctionSymbol is KaPropertySetterSymbol
     }
+}
+
+private class KSFunctionDeclarationAAImpl(
+    override val ktFunctionSymbol: KaFunctionSymbol
+) : KSFunctionDeclaration, KSFunctionDeclarationImpl() {
+    companion object : KSObjectCache<KaFunctionSymbol, KSFunctionDeclarationAAImpl>() {
+        fun getCached(symbol: KaFunctionSymbol) = cache.getOrPut(symbol) {
+            KSFunctionDeclarationAAImpl(symbol)
+        }
+    }
+}
+
+private class KSFunctionDeclarationPsiImpl(
+    val psiMethod: PsiMethod,
+    val parentClass: KSClassDeclarationImpl
+) : KSFunctionDeclaration, KSFunctionDeclarationImpl() {
+    companion object : KSObjectCache<PsiMethod, KSFunctionDeclarationPsiImpl>() {
+        fun getCached(psiMethod: PsiMethod, parentClass: KSClassDeclarationImpl) = cache.getOrPut(psiMethod) {
+            KSFunctionDeclarationPsiImpl(psiMethod, parentClass)
+        }
+    }
+
+    override val ktFunctionSymbol: KaFunctionSymbol by lazy {
+        analyze { findKaFunctionSymbol(psiMethod, parentClass) }
+            ?: throw InternalKSPException(
+                "Failed to resolve KaFunctionSymbol for method ${parentClass.simpleName.asString()}.${psiMethod.name}",
+                psiMethod.toLocation(),
+                psiMethod.javaClass,
+            )
+    }
+
+    override val simpleName: KSName by lazy {
+        if (psiMethod.isConstructor) KSNameImpl.getCached("<init>") else KSNameImpl.getCached(psiMethod.name)
+    }
+
+    override val origin: Origin by lazy {
+        when (parent) {
+            // If our lazy parent resolution determined that this method belongs to a Kotlin property,
+            // it means FIR considers it a synthetic property accessor.
+            is KSPropertyDeclaration -> Origin.SYNTHETIC
+            else -> parentClass.origin
+        }
+    }
+
+    override val containingFile: KSFile? get() = parentClass.containingFile
+
+    override val packageName: KSName get() = parentClass.packageName
+
+    override val isAbstract: Boolean by lazy { psiMethod.hasModifierProperty(PsiModifier.ABSTRACT) }
+
+    override val functionKind: FunctionKind by lazy {
+        if (psiMethod.hasModifierProperty(PsiModifier.STATIC)) FunctionKind.STATIC else FunctionKind.MEMBER
+    }
+
+    override val modifiers: Set<Modifier> by lazy {
+        if (origin == Origin.JAVA) {
+            return@lazy psiMethod.toKSModifiers()
+        }
+
+        val psiModifiers = psiMethod.toKSModifiers().toMutableSet()
+
+        // Emulate AA: Strip JVM-specific modifiers from compiled Java methods
+        psiModifiers.remove(Modifier.JAVA_TRANSIENT)
+        psiModifiers.remove(Modifier.JAVA_VOLATILE)
+        psiModifiers.remove(Modifier.JAVA_SYNCHRONIZED)
+        psiModifiers.remove(Modifier.JAVA_NATIVE)
+        psiModifiers.remove(Modifier.JAVA_STRICT)
+        psiModifiers.remove(Modifier.JAVA_DEFAULT)
+
+        when {
+            psiMethod.isConstructor -> {
+                // Constructors are always final
+                psiModifiers.add(Modifier.FINAL)
+            }
+            psiMethod.hasModifierProperty(PsiModifier.STATIC) -> {
+                // Emulate AA: Static methods in compiled Java libraries are treated as final in Kotlin
+                psiModifiers.add(Modifier.FINAL)
+            }
+            !psiMethod.hasModifierProperty(PsiModifier.FINAL) &&
+                !psiMethod.hasModifierProperty(PsiModifier.PRIVATE) -> {
+                // Non-final, non-static, non-private Java instance methods are open in Kotlin
+                psiModifiers.add(Modifier.OPEN)
+            }
+        }
+
+        return@lazy psiModifiers
+    }
+
+    override val annotations: Sequence<KSAnnotation> by lazyMemoizedSequence {
+        psiMethod.annotations.asSequence().map { KSAnnotationJavaImpl.getCached(it, this) }
+    }
+
+    // Java doesn't have extension receivers.
+    override val extensionReceiver: KSTypeReference? = null
+}
+
+/**
+ * Resolves the [KaFunctionSymbol] that corresponds to the given [PsiMethod] within the [parent] class.
+ *
+ * This method searches for a symbol whose underlying PSI matches the provided [psiMethod] using
+ * exact identity or equivalence. It handles several Kotlin Analysis API (FIR) interop edge cases:
+ *
+ * 1. **Performance Optimization (KT-85692):** For non-constructor methods, it queries `memberScope`
+ *    and `staticMemberScope` instead of `declaredMemberScope`. Querying the declared member scope
+ *    on Java classes triggers a severe performance bottleneck in the Analysis API, as it lacks a
+ *    dedicated declared enhancement scope for Java classes and falls back to resolving the entire
+ *    inherited use-site scope recursively.
+ * 2. **Property Accessor Fallback:** If a Java method follows getter/setter conventions (e.g., `getFoo()`)
+ *    and overrides a Kotlin property (`val foo`), FIR hides the method from the standard function scope
+ *    and exposes it as a property symbol instead. If the initial function match fails, this method
+ *    falls back to searching for property accessors via `findKaPropertyAccessorSymbols`.
+ *
+ * @param psiMethod The Java PSI method to resolve.
+ * @param parent The enclosing class declaration proxy.
+ * @return The resolved [KaFunctionSymbol], or null if the symbol could not be found.
+ */
+fun KaSession.findKaFunctionSymbol(psiMethod: PsiMethod, parentClass: KSClassDeclarationImpl): KaFunctionSymbol? {
+    fun Sequence<KaSymbol>.firstMatchingPsiMethod(psi: PsiMethod): KaFunctionSymbol? {
+        return filterIsInstance<KaFunctionSymbol>().find { it.psi == psi || it.psi?.isEquivalentTo(psi) == true }
+    }
+    val parentSymbol = parentClass.ktClassOrObjectSymbol
+    return if (psiMethod.isConstructor) {
+        parentSymbol.declaredMemberScope.constructors.firstMatchingPsiMethod(psiMethod)
+    } else {
+        val methodName = Name.identifier(psiMethod.name)
+        // Note: Technically, declaredMemberScope and staticDeclaredMemberScope could be used here since the psi method
+        // is declared in the parent. However, those methods trigger the performance issue mentioned in
+        // https://youtrack.jetbrains.com/issue/KT-85692, so stick with memberScope/staticMemberScope instead.
+        parentSymbol.memberScope.callables(methodName).firstMatchingPsiMethod(psiMethod)
+            ?: parentSymbol.staticMemberScope.callables(methodName).firstMatchingPsiMethod(psiMethod)
+            // If the above match failed, it means that FIR hid the method because it overrides a Kotlin property (e.g.
+            // `getX` -> `x`), so we need to find the property and get the KaFunctionSymbol from it instead.
+            ?: findKaPropertyAccessorSymbols(psiMethod, parentClass).firstMatchingPsiMethod(psiMethod)
+    }
+}
+
+private fun KaSession.findKaPropertyAccessorSymbols(
+    psiMethod: PsiMethod,
+    parentClass: KSClassDeclarationImpl
+): Sequence<KaPropertyAccessorSymbol> {
+    val methodName = psiMethod.name
+    val namesToSearch = when {
+        // A standard Java getter like `getFoo()` overrides a Kotlin property named `foo` or `Foo`.
+        methodName.startsWith("get") && methodName.length > 3 -> {
+            val accessorBase = methodName.substring(3)
+            val propertyName = accessorBase.replaceFirstChar { it.lowercaseChar() }
+            sequenceOf(propertyName, accessorBase)
+        }
+        // A Java boolean getter like `isFoo()` can override a Kotlin property named `foo` or `Foo`,
+        // OR a property explicitly named `isFoo`. All are valid in Kotlin interoperability.
+        methodName.startsWith("is") && methodName.length > 2 -> {
+            val accessorBase = methodName.substring(2)
+            val propertyName = accessorBase.replaceFirstChar { it.lowercaseChar() }
+            sequenceOf(propertyName, accessorBase, methodName)
+        }
+        // A Java setter like `setFoo()` can override a standard Kotlin property named `foo` or `Foo`,
+        // OR it can override a boolean property explicitly named `isFoo`.
+        methodName.startsWith("set") && methodName.length > 3 -> {
+            val accessorBase = methodName.substring(3)
+            val propertyName = accessorBase.replaceFirstChar { it.lowercaseChar() }
+            sequenceOf(propertyName, accessorBase, "is$accessorBase")
+        }
+        else -> return emptySequence()
+    }
+    return namesToSearch
+        .flatMap { parentClass.ktClassOrObjectSymbol.memberScope.callables(Name.identifier(it)) }
+        .filterIsInstance<KaPropertySymbol>()
+        .flatMap { sequenceOf(it.getter, it.setter) }
+        .filterNotNull()
 }
 
 internal fun KaFunctionSymbol.toModifiers(): Set<Modifier> {
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
@@ -1,22 +1,45 @@
 package com.google.devtools.ksp.impl.symbol.kotlin
 
+import com.google.devtools.ksp.InternalKSPException
 import com.google.devtools.ksp.common.KSObjectCache
 import com.google.devtools.ksp.common.impl.KSNameImpl
+import com.google.devtools.ksp.common.lazyMemoizedSequence
+import com.google.devtools.ksp.common.toKSModifiers
 import com.google.devtools.ksp.impl.ResolverAAImpl
+import com.google.devtools.ksp.impl.symbol.java.KSAnnotationJavaImpl
 import com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSTypeReferenceResolvedImpl
 import com.google.devtools.ksp.symbol.*
+import com.intellij.psi.PsiField
+import com.intellij.psi.PsiModifier
+import org.jetbrains.kotlin.analysis.api.symbols.KaDeclarationSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaJavaFieldSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolModality
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolVisibility
+import org.jetbrains.kotlin.name.Name
 
-class KSPropertyDeclarationJavaImpl private constructor(val ktJavaFieldSymbol: KaJavaFieldSymbol) :
-    KSPropertyDeclaration,
-    AbstractKSDeclarationImpl(ktJavaFieldSymbol),
-    KSExpectActual by KSExpectActualImpl(ktJavaFieldSymbol) {
-    companion object : KSObjectCache<KaJavaFieldSymbol, KSPropertyDeclaration>() {
-        fun getCached(ktJavaFieldSymbol: KaJavaFieldSymbol): KSPropertyDeclaration =
-            cache.getOrPut(ktJavaFieldSymbol) { KSPropertyDeclarationJavaImpl(ktJavaFieldSymbol) }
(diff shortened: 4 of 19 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -21,37 +21,55 @@ import com.google.devtools.ksp.*
...
+    override val isActual: Boolean get() = expectActualImpl.isActual
+    override val isExpect: Boolean get() = expectActualImpl.isExpect
+    override fun findActuals(): Sequence<KSDeclaration> = expectActualImpl.findActuals()
+    override fun findExpects(): Sequence<KSDeclaration> = expectActualImpl.findExpects()
+
     override val functionKind: FunctionKind by lazy {
         when (ktFunctionSymbol.location) {
             KaSymbolLocation.CLASS -> {
-                if (ktFunctionSymbol is KaNamedFunctionSymbol && ktFunctionSymbol.isStatic) {
+                if ((ktFunctionSymbol as? KaNamedFunctionSymbol)?.isStatic == true) {
```
jaschdoc, line 98: Why is this change preferable to the old style? Is it faster than using `is`?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -65,6 +83,8 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
         if (isConstructor()) {
             val ksClassDeclaration = parentDeclaration as KSClassDeclaration
             if (ksClassDeclaration.classKind == ClassKind.ENUM_CLASS) {
+                // receivers are modeled as parameter in AA therefore annotations are stored in
+                // the corresponding receiver parameter, need to pass it to the `KSTypeReferenceImpl`
                 setOf(Modifier.FINAL, Modifier.PRIVATE)
```
jaschdoc, line 122: Why was this comment moved here? :)

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -126,7 +144,7 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
         recordLookupForPropertyOrMethod(this)
         return analyze {
             if (ktFunctionSymbol is KaPropertyAccessorSymbol) {
-                (parentDeclaration as? KSPropertyDeclarationImpl)?.ktPropertySymbol
+                ktFunctionSymbol.containingDeclaration as? KaPropertySymbol
```
jaschdoc: How does this differ (in the set of returned values) from the old implementation? I.e., does `ktFunctionSymbol.containingDeclaration` have different semantics than `parentDeclaration.ktPropertySymbol`?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -139,12 +157,12 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
 
     override val simpleName: KSName by lazy {
         when (ktFunctionSymbol) {
-            is KaNamedFunctionSymbol -> KSNameImpl.getCached(ktFunctionSymbol.name.asString())
+            is KaNamedFunctionSymbol ->
+                KSNameImpl.getCached((ktFunctionSymbol as KaNamedFunctionSymbol).name.asString())
```
jaschdoc: I think we can just bind the subject in the `when` expression and avoid the type cast.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -234,6 +251,141 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
     }
 }
 
+private class KSFunctionDeclarationAAImpl(
+    override val ktFunctionSymbol: KaFunctionSymbol
+) : KSFunctionDeclaration, KSFunctionDeclarationImpl() {
```
jaschdoc, line 314: Is it necessary to declare it as a `KSFunctionDeclaration` here? Isn't that handled by subtyping since it's a `KSFunctionDeclarationImpl` is a subtype of `KSFunctioNDeclaration`?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -234,6 +251,141 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+    companion object : KSObjectCache<KaFunctionSymbol, KSFunctionDeclarationAAImpl>() {
+        fun getCached(symbol: KaFunctionSymbol) = cache.getOrPut(symbol) {
+            KSFunctionDeclarationAAImpl(symbol)
+        }
+    }
+}
```
jaschdoc, line 320: If this class' implementation is all inherited, why are we changing methods / properties in the super class?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -234,6 +251,141 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+
+    override val isAbstract: Boolean by lazy { psiMethod.hasModifierProperty(PsiModifier.ABSTRACT) }
+
+    override val functionKind: FunctionKind by lazy {
+        if (psiMethod.hasModifierProperty(PsiModifier.STATIC)) FunctionKind.STATIC else FunctionKind.MEMBER
+    }
+
+    override val modifiers: Set<Modifier> by lazy {
+        if (origin == Origin.JAVA) {
+            return@lazy (psiMethod as PsiModifierListOwner).toKSModifiers()
```
jaschdoc: Why is the cast `psiMethod as PsiModifierListOwner` necessary? Shouldn't it be handled by subtyping or are there quirks with extension methods and subtyping?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -234,6 +251,141 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+        }
+
+        val psiModifiers = (psiMethod as PsiModifierListOwner).toKSModifiers().toMutableSet()
```
jaschdoc: Same comment as above :)

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -234,6 +251,141 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+                // Emulate AA: Static methods in compiled Java libraries are treated as final in Kotlin
+                psiModifiers.add(Modifier.FINAL)
+            } else if (!psiMethod.hasModifierProperty(PsiModifier.FINAL) &&
+                !psiMethod.hasModifierProperty(PsiModifier.PRIVATE)
+            ) {
+                // Non-final, non-static, non-private Java instance methods are open in Kotlin
+                psiModifiers.add(Modifier.OPEN)
+            }
+        }
+
```
jaschdoc: These branches are mutually exclusive so either we can return directly from each branch or rewrite it to a `when` expression to avoid nesting `if`s.
```kt
if (psiMethod.isConstructor) {
    //...
    return@lazy psiModifiers
}
if (psiMethod.hasModifierProperty(PsiModifier.STATIC)) {
    //...
    return@lazy psiModifiers
}
if (!psiMethod.hasModifierProperty(PsiModifier.FINAL) && !psiMethod.hasModifierProperty(PsiModifier.PRIVATE)) {
    //...
    return@lazy psiModifiers // Optional return here
}
psiModifiers // I don't think we need an explicit return here?
```
or
```kt
when {
    psiMethod.isConstructor -> // ...
    psiMethod.hasModifierProperty(PsiModifier.STATIC) -> // ...
    !psiMethod.hasModifierProperty(PsiModifier.FINAL) && !psiMethod.hasModifierProperty(PsiModifier.PRIVATE) -> // ...
}
psiModifiers
```
Maybe we could also extract some methods / properties for the predicates to make it a bit more readable.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -234,6 +251,141 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+    }
+
+    // Java doesn't have extension receivers.
+    override val extensionReceiver: KSTypeReference? = null
+}
+
+private fun KaSession.findKaFunctionSymbol(psiMethod: PsiMethod, parent: KSClassDeclarationImpl): KaFunctionSymbol? {
+    fun Sequence<KaSymbol>.firstMatchingPsiMethod(psi: PsiMethod): KaFunctionSymbol? {
+        return filterIsInstance<KaFunctionSymbol>().find { it.psi == psi || it.psi?.isEquivalentTo(psi) == true }
+    }
```
jaschdoc, line 429: Does this function implicitly depend on the `KaSession` or can it be lifted outside `findKaFunctionSymbol`?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -234,6 +251,141 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+        // Note: Technically, declaredMemberScope and staticDeclaredMemberScope could be used here since the psi method
+        // is declared in the parent. However, those methods trigger the performance issue mentioned in
+        // https://youtrack.jetbrains.com/issue/KT-85692, so stick with memberScope/staticMemberScope instead.
+        parent.ktClassOrObjectSymbol.memberScope.callables(methodName).firstMatchingPsiMethod(psiMethod)
+            ?: parent.ktClassOrObjectSymbol.staticMemberScope.callables(methodName).firstMatchingPsiMethod(psiMethod)
+            // If the above match failed, it means that FIR hid the method because it overrides a Kotlin property (e.g.
+            // `getX` -> `x`), so we need to find the property and get the KaFunctionSymbol from it instead.
+            ?: findKaPropertyAccessorSymbols(psiMethod, parent).firstMatchingPsiMethod(psiMethod)
+    }
+}
```
jaschdoc, line 444: Would it be possible to add a comment right after the else branch that briefly describes what the overall strategy is? E.g.,
Look up the method on the parent in the following order:
1. Check class members, including inherited members
2. Check static class members
3. Property accessors

jaschdoc: Or maybe add it as a doc comment :)

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -234,6 +251,141 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+
+private fun KaSession.findKaPropertyAccessorSymbols(
+    psiMethod: PsiMethod,
+    parent: KSClassDeclarationImpl
+): Sequence<KaPropertyAccessorSymbol> {
+    val methodName = psiMethod.name
+    val namesToSearch = when {
+        // A standard Java getter like `getFoo()` overrides a Kotlin property named `foo`.
+        methodName.startsWith("get") && methodName.length > 3 -> {
+            val propertyName = methodName.substring(3).replaceFirstChar { it.lowercaseChar() }
```
jaschdoc: Could a Kotlin property also be named `Foo`, i.e., with the first character being upper case?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -619,21 +627,19 @@ class ResolverAAImpl(
...
-                overrider.ktFunctionSymbol
+        fun KSDeclaration.getCallableSymbol(): KaCallableSymbol? {
+            return when (this) {
+                is KSFunctionDeclarationImpl -> if (ktFunctionSymbol is KaPropertyAccessorSymbol) {
+                    analyze { ktFunctionSymbol.containingDeclaration as KaPropertySymbol }
+                } else {
+                    ktFunctionSymbol
+                }
+                is KSPropertyDeclarationImpl -> ktPropertySymbol
+                else -> null
```
jaschdoc: I think this function can be moved to `util.kt` in `impl/symbol`.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -704,10 +710,6 @@ class ResolverAAImpl(
         fun KSType.toSignature(): String {
             return if (this is KSTypeImpl) {
                 analyze {
-                    val decl = (this@toSignature.declaration as? KSClassDeclaration)
-                    // Force inline value class to be unwrapped.
-                    // Do not remove until AA fixes this.
-                    decl?.primaryConstructor
```
jaschdoc, line 820: Nice 🥳

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
```diff
@@ -67,6 +68,7 @@ class ResolverAAImpl(
     val allKSFiles: List<KSFile>,
     val project: Project,
     val incrementalContext: IncrementalContextAA,
+    val kspConfig: KSPConfig,
```
jaschdoc: Instead of adding `KSPConfig` to the `Resolver`, should we not refactor `KSPConfig` to have an `instance` property on the companion object?
To me it seems that it's okay for every `KSNode` and implementation of the API to depend on a global configuration, whereas depending on the `Resolver` everywhere is not. My concern is that one easily adds seemingly unrelated things to the `Resolver`. Or in other words, `Resolver`'s scope slowly increases and eventually everything will depend on it.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -28,18 +28,30 @@ import com.google.devtools.ksp.impl.recordLookup
...
+import org.jetbrains.org.objectweb.asm.Opcodes
 
 class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSymbol: KaClassSymbol) :
     KSClassDeclaration,
-    AbstractKSDeclarationImpl(ktClassOrObjectSymbol),
+    AbstractKSDeclarationImpl(),
     KSExpectActual by KSExpectActualImpl(ktClassOrObjectSymbol) {
+    private val usePsiStrategy: Boolean by lazy {
+        runCatching { ResolverAAImpl.instance }.getOrNull()?.kspConfig?.experimentalPsiResolution == true
+    }
```
jaschdoc, line 48: See my other comment in the resolver. IMO, this should just be `KSPConfig.instance.experimentalPsiResolution`.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +192,114 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
     }
 
     override val declarations: Sequence<KSDeclaration> by lazyMemoizedSequence {
-        val decls = ktClassOrObjectSymbol.declarations()
-        if (origin == Origin.JAVA && classKind != ClassKind.ANNOTATION_CLASS) {
+        // FAST PATH: Use PSI directly for Java classes
+        if (usePsiStrategy &&
+            (origin == Origin.JAVA || origin == Origin.JAVA_LIB) &&
+            classKind != ClassKind.ANNOTATION_CLASS &&
+            classKind != ClassKind.ENUM_CLASS
```
jaschdoc: nit: could we name these predicates or group them into two predicates. E.g., `mayUseFastResolution` and `isNormalClass` or something to that effect.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +192,114 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+
+        fun PsiMethod.isSynthetic(): Boolean {
+            if (binaryClassInfo != null && !isConstructor) {
+                val overloadsFlags = binaryClassInfo.methodAccFlags
+                    .filterKeys { it.startsWith("$name(") }
+                    .values
+
+                if (overloadsFlags.isNotEmpty() && overloadsFlags.all { (it and Opcodes.ACC_SYNTHETIC) != 0 }) {
+                    return true
+                }
```
jaschdoc: This predicate could just be returned directly here. Also, it seems like we should check that `any` of the opcodes is equal to the synthetic attribute instead of bitwise `AND` since it is allowed to have multiple access flags. According to [the specification](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.7.8) we should also check the attribute in addition to the `ACC_SYNTHETIC` access flag.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +192,114 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+            }
+            return false
+        }
+
+        fun PsiField.isSynthetic(): Boolean {
+            if (binaryClassInfo != null) {
+                val access = binaryClassInfo.fieldAccFlags[name] ?: 0
+                return (access and Opcodes.ACC_SYNTHETIC) != 0
+            }
```
jaschdoc: Here we could also just ask for equality on `ACC_SYNTHETIC`.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +192,114 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+            (ktClassOrObjectSymbol.psi as? PsiClass)?.getDeclarationsFromPsi()
+                // If the PSI fast path succeeds, return it. Otherwise, gracefully fall back.
+                ?.let { return@lazyMemoizedSequence it }
+        }
+
+        // SLOW PATH: Original AA fallback for Kotlin classes (or if ASM failed)
+        getDeclarationsFromAnalysisApi()
+    }
+
+    private fun PsiClass.getDeclarationsFromPsi(): Sequence<KSDeclaration>? {
```
jaschdoc: It seems like this helper methods (along with the `*.isSynthetic` methods) could be moved to `impl/symbol/util/PsiUtils.kt`

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
```diff
@@ -70,6 +93,80 @@ class KSPropertyDeclarationJavaImpl private constructor(val ktJavaFieldSymbol: K
...
+
+    override val isMutable: Boolean
+        get() = !psiField.hasModifierProperty(PsiModifier.FINAL)
+
+    override val modifiers: Set<Modifier> by lazy {
+        if (origin == Origin.JAVA) {
+            return@lazy (psiField as PsiModifierListOwner).toKSModifiers()
+        }
+
+        val psiModifiers = (psiField as PsiModifierListOwner).toKSModifiers().toMutableSet()
```
jaschdoc: Same comment here about type casts vs. subtyping / inheritance :)

jaschdoc (commented):
Thank you for the PR! Most of my comments are nits and most questions are just for clarification.
However, I'm concerned that we are starting to depend on the configuration mode in too many places and should reconsider how we want to support these two modes of operation. But that is outside the scope of this PR :)

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: No, this change isn't about performance, it's just because Kotlin smart casts don't work now that `ktFunctionSymbol` is `open` so we have to manually cast.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: It's not technically necessary, but one of the tests depends on this implementation detail unfortunately. Specifically, [GetSymbolsFromAnnotationProcessor](https://github.com/google/ksp/blob/8ac7c6324668ab1ffaab66b60dc0ac23a0b10260/test-utils/src/main/kotlin/com/google/devtools/ksp/processor/GetSymbolsFromAnnotationProcessor.kt#L55):

```kotlin
private fun toString(annotated: KSAnnotated): String {
  return "$annotated:${annotated::class.supertypes.first().classifier.toString().substringAfterLast('.')}"
}
```

In particular, it usese the name of the first super type rather than the class name itself. I think we could just update this test to avoid this implementation detail, but I wanted to avoid changing that for this CL.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: There are some slight changes to the super class to avoid eager initialization of the `KaFunctionSymbol`. For example, the super class used to extend the following:

```kotlin
class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbol: KaFunctionSymbol) :
    KSFunctionDeclaration,
    AbstractKSDeclarationImpl(ktFunctionSymbol),
    KSExpectActual by KSExpectActualImpl(ktFunctionSymbol) {
```

But `AbstractKSDeclarationImpl(ktFunctionSymbol)` doesn't work any more because `ktFunctionSymbol` is exactly the expensive thing we're trying to avoid creating eagerly in the PSI implementation. Similarly, for `KSExpectActual by KSExpectActualImpl(ktFunctionSymbol)`. Thus, I had to shift things around to allow those to be created lazily in the super class. The other changes in the super class is just because `ktFunctionSymbol` property is now open so some smart casts no longer work.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: it implicitly depends on the `KaSession`, e.g. calling the `memberScope` is only available within a KaSession. We could alternatively wrap this method body in `analyze { ... }`, but from my reading of KaSession best practices, there's some cost each time you create and tear down an analysis session so it's better to do it this way and rely on the caller to declare the `analyze { ... }` block.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
bcorso: This sounds good. I initially started with the resolver just because it seemed like it was already being accessed globally, but I agree it seems safer to depend directly on `KSPConfig` where possible.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: lol, sorry, it accidentally got removed at some point and I went to add it back but put it in the wrong spot. I'll fix this!

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: sgtm

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: I completely agree. I'll fix this.

bcorso: Actually, I think this will take more design work.

I realized there's two issues here that don't show up with the `ResolverAAImpl` approach:

  1. Unlike `ResolverAAImpl` where's there's only a single implementation, `KSPConfig` has multiple implementations (namely `KSPJvmConfig`, `KSPNativeConfig`, `KSPCommonConfig`, etc.). Thus, if we were to go this route I think we would need to put the `instance` cache into the individual implementations and callers would need to call those, e.g. `KSPJvmConfig.instance` rather than `KSPConfig.instance`. This might not necessarily be an issue, if all callers know which implementation to use, but I'm not sure if that's the case.
  2. Unlike `ResolverAAImpl` where KSP controls the creation of the instances, `KSPConfig` instances are created by the user. Thus there's no way to guarantee that there's only one implementation at a time. 

In general, it seems like this requires more thought about the design so I'm going to hold off on changing this in this CL.

bcorso: Actually, I guess we could work around these two issues by making the setter method internal and have KSP set it manually when `KotlinSymbolProcessing` is initialized. Let me try this approach.

bcorso: Nm, KSPConfig is in a separate build artifact as KotlinSymbolProcessing so the only way to do it would be to make the `instance`'s getter and setter public,

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
jaschdoc: Aha, I understand. Thanks for the explanation. I might change that test so it's not so tightly coupled with the implementation.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
jaschdoc: Sounds good, thank you for the explanation.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: Nope, not necessary. Thanks, I'll remove it.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: Nope, not necessary. Thanks, I'll remove it.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
bcorso: See the related comment above. This will probably need more thought and kind of tangential to this PR, so I think it's best to leave this for a separate PR.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: You can ignore this now as I've reverted it.

I initially changed this because the `parentDeclaration` in the PSI implementation didn't match the AA implementation. In particular, the AA implementation used the property as the parent of the accessor but the PSI implementation always used the class as the parent. I've updated the PSI implementation to match the AA implementation, and reverted this change back to how it was initially.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: Done.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: You can ignore this now as it's been reverted. 

The `isSynthetic` methods are no longer needed. The particular cases that was causing these issues now always use the AA implementation so that we don't need to try to handle them in the PSI implementation.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: You can ignore this now as it's been reverted. 

The `isSynthetic` methods are no longer needed. The particular cases that was causing these issues now always use the AA implementation so that we don't need to try to handle them in the PSI implementation.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: You can ignore this now as it's been reverted. 

The `isSynthetic` methods are no longer needed. The particular cases that was causing these issues now always use the AA implementation so that we don't need to try to handle them in the PSI implementation.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
bcorso: I decided against this because I don't think this particular method is super general. It's really getting the callableSymbol that's relevant for the "overrides` calculation so it seems very specific to this overrides method and might cause confusion if we tried to move it into a more generic location.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: I think you might be right. I've updated the logic to handle it but I still need to try to add a test case for this.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: Done.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: Done.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
bcorso: Done.

jaschdoc:
@bcorso, do you want me to review it again? If so, could you rebase on `main` and push again? We recently had some CI issues which _should_ be fixed for now.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
jaschdoc: Okay, I understand. We could also move the `kspConfig` property to `KotlinSymbolProcessing` since that is the root component that spawns everything else. It seems fine to me at least that all transitive callees of `KotlinSymbolProcessing.execute` gets to read (but importantly not write) some property on it. This can either be modeled as a globally accessible instance or a Kotlin context parameter or just a regular parameter.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +190,93 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+                    return@lazyMemoizedSequence psi.getDeclarationsFromPsi()
+                }
+            }
+        }
+
+        // SLOW PATH: Original AA fallback for Kotlin classes (or if ASM failed)
+        getDeclarationsFromAnalysisApi()
+    }
+
+    fun canUsePsiResolution(psiClass: PsiClass): Boolean {
```
jaschdoc: Should this function be private?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +190,93 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+
+        // 3. Constructors
+        // IntelliJ's PSI lacks implicit constructors. If none are explicitly declared,
+        // we must fetch the synthesized constructor from the Analysis API.
+        val psiConstructors = if (constructors.isEmpty() && !isInterface) {
+            analyze {
+                ktClassOrObjectSymbol.declaredMemberScope.constructors.map {
+                    KSFunctionDeclarationImpl.getCached(it)
+                }
+            }
```
jaschdoc, line 236: Could we try to construct the name ourselves and somehow obtain a handle to a symbol that way? This is probably we can explore outside of this PR, but my concern is that `declaredMemberSope` triggers the expensive computation we are trying to avoid.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +190,93 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+        // 4. Inner Classes
+        val psiInnerClasses = innerClasses.asSequence()
+            .mapNotNull { psiInnerClass ->
+                analyze {
+                    val name = psiInnerClass.name?.let { Name.identifier(it) } ?: return@analyze null
+                    val instanceSymbols = ktClassOrObjectSymbol.declaredMemberScope.classifiers(name)
+                    val staticSymbols = ktClassOrObjectSymbol.staticDeclaredMemberScope.classifiers(name)
+                    (instanceSymbols + staticSymbols)
+                        .filterIsInstance<KaNamedClassSymbol>()
+                        .find { it.psi == psiInnerClass || it.psi?.isEquivalentTo(psiInnerClass) == true }
```
jaschdoc, line 251: This seems like it could be expensive when obtaining a list based on a name and then `find`ing a given element again (on first look it looks like like a `O(n^2)` operation). Intuitively, it might be cheaper to ask for all nested declarations and store it in a map. The predicate here suggests we need two maps. However, generally there aren't too many nested / inner classes, so I think this is fine for now (let's not prematurely optimize this).

Do you have any thoughts on this or is this not a concern to you?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +190,93 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+                        ?.let { KSClassDeclarationImpl.getCached(it) }
+                }
+            }
+
+        // 5. Partition members to emulate FIR's scope architecture
+        // We materialize the regular members to a list to avoid evaluating the PSI sequences twice.
```
jaschdoc: > ... to avoid evaluating the PSI sequences twice.

Can we use a memoizing sequence and also use the `.distinct` method on the base sequence, or does that not work?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +190,93 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+        val unpartitioned = (psiFields + psiMethods + psiInnerClasses).toList()
+
+        // Split the list into static members and instance members
+        val (psiStaticMembers, psiInstanceMembers) = unpartitioned.partition {
+            Modifier.JAVA_STATIC in it.modifiers ||
+                // Nested classes in KSP don't get the static modifier.
+                (it is KSClassDeclaration && Modifier.INNER !in it.modifiers)
```
jaschdoc, line 260: This seems confusing to me. Why are we filtering for nested classes that do not have the `INNER` modifier, when the comment says that those classes do not get the `static` modifier?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
```diff
@@ -180,8 +190,93 @@ class KSClassDeclarationImpl private constructor(internal val ktClassOrObjectSym
...
+        return psiInstanceMembers.asSequence() + psiConstructors + psiStaticMembers.asSequence()
+    }
+
+    private fun getDeclarationsFromAnalysisApi(): Sequence<KSDeclaration> {
         val decls = ktClassOrObjectSymbol.declarations()
-        if (origin == Origin.JAVA && classKind != ClassKind.ANNOTATION_CLASS) {
+            // The Analysis API is known to leak certain inherited members, so we filter those cases out here.
+            .filter { !it.isLeakedInheritedMember(this) }
+
+        return if ((origin == Origin.JAVA || origin == Origin.JAVA_LIB) && classKind != ClassKind.ANNOTATION_CLASS) {
```
jaschdoc, line 273: Could you explain why the slow path now also accepts Java libs?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -288,6 +309,172 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+    val parentClass: KSClassDeclarationImpl
+) : KSFunctionDeclaration, KSFunctionDeclarationImpl() {
+    companion object : KSObjectCache<PsiMethod, KSFunctionDeclarationPsiImpl>() {
+        fun getCached(psiMethod: PsiMethod, parentClass: KSClassDeclarationImpl) = cache.getOrPut(psiMethod) {
+            KSFunctionDeclarationPsiImpl(psiMethod, parentClass)
+        }
+    }
+
+    override val ktFunctionSymbol: KaFunctionSymbol by lazy {
+        requireNotNull(analyze { findKaFunctionSymbol(psiMethod, parentClass) }) {
```
jaschdoc: Could I ask you to throw `InternalKSPException` here instead?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
```diff
@@ -288,6 +309,172 @@ class KSFunctionDeclarationImpl private constructor(internal val ktFunctionSymbo
...
+
+    override val simpleName: KSName by lazy {
+        if (psiMethod.isConstructor) KSNameImpl.getCached("<init>") else KSNameImpl.getCached(psiMethod.name)
+    }
+
+    override val origin: Origin by lazy {
+        when (parent) {
+            // If our lazy parent resolution determined that this method belongs to a Kotlin property,
+            // it means FIR considers it a synthetic property accessor.
+            is KSPropertyDeclaration -> Origin.SYNTHETIC
```
jaschdoc, line 349: Are explicitly declare property accessor methods also considered synthetic? E.g.,
```kotlin
val prop: Int get() = 42
```

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
```diff
@@ -70,6 +92,80 @@ class KSPropertyDeclarationJavaImpl private constructor(val ktJavaFieldSymbol: K
...
+        fun getCached(symbol: KaJavaFieldSymbol) = cache.getOrPut(symbol) {
+            KSPropertyDeclarationJavaAAImpl(symbol)
+        }
+    }
+}
+
+class KSPropertyDeclarationJavaPsiImpl private constructor(
+    val psiField: PsiField,
+    override val parent: KSClassDeclarationImpl,
+) : KSPropertyDeclarationJavaImpl() {
```
jaschdoc, line 109: Why is one of these classes private while the other is public but with a private constructor?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
```diff
@@ -70,6 +92,80 @@ class KSPropertyDeclarationJavaImpl private constructor(val ktJavaFieldSymbol: K
...
+            val targetName = Name.identifier(psiField.name)
+            // Note: Technically, declaredMemberScope and staticDeclaredMemberScope could be used here since the psi
+            // field is declared in the parent. However, those scopes trigger the performance issue mentioned in
+            // https://youtrack.jetbrains.com/issue/KT-85692, so stick with memberScope/staticMemberScope instead.
+            val instanceSymbols = parent.ktClassOrObjectSymbol.memberScope.callables(targetName)
+            val staticSymbols = parent.ktClassOrObjectSymbol.staticMemberScope.callables(targetName)
+            requireNotNull(
+                (instanceSymbols + staticSymbols)
+                    .find { it.psi == psiField || it.psi?.isEquivalentTo(psiField) == true } as? KaJavaFieldSymbol
+            ) { "Failed to resolve KaJavaFieldSymbol for field ${psiField.name}" }
```
jaschdoc: Would also be great if you used `InternalKSPException` here too! :)

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
```diff
@@ -1196,6 +1197,116 @@ internal val KaDeclarationSymbol.internalSuffix: String
...
+ * maps methods from specific types to properties or renamed identifiers (e.g., mapping
+ * `length()` to a property for `CharSequence`, or `remove` to `removeAt`). KSP must
+ * fall back to the Analysis API slow path for these types to maintain behavioral consistency.
+ */
+fun isKotlinAlteredType(fqName: String) =
+    fqName.startsWith("kotlin.collections.") ||
+        fqName == "kotlin.CharSequence" ||
+        fqName == "kotlin.Throwable" ||
+        fqName == "kotlin.Number"
+
```
jaschdoc, line 1321: This seems a bit fragile. Maybe this list is something that would be nice to have in the Analysis API?

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/DeclarationsInClassProcessor.kt
```diff
@@ -0,0 +1,89 @@
...
+import com.google.devtools.ksp.processing.Resolver
+import com.google.devtools.ksp.symbol.KSAnnotated
+import com.google.devtools.ksp.symbol.KSClassDeclaration
+import com.google.devtools.ksp.symbol.KSDeclaration
+import com.google.devtools.ksp.symbol.KSFunctionDeclaration
+import com.google.devtools.ksp.symbol.KSPropertyDeclaration
+
+class DeclarationsInClassProcessor : AbstractTestProcessor() {
+    val result = mutableListOf<String>()
+    override fun toResult() = result
```
jaschdoc: Let's sort this so results are a bit more stable.

test-utils/testData/api/declarationsInClass.kt
```diff
@@ -0,0 +1,234 @@
+/*
+ * Copyright 2023 Google LLC
+ * Copyright 2010-2023 JetBrains s.r.o. and Kotlin Programming Language contributors.
```
jaschdoc: 2026?

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/DeclarationsInClassProcessor.kt
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2020 Google LLC
+ * Copyright 2010-2020 JetBrains s.r.o. and Kotlin Programming Language contributors.
```
jaschdoc: 2026?

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/DeclarationsInClassProcessor.kt
```diff
@@ -0,0 +1,89 @@
...
+        processDeclaration(resolver.getClassDeclarationByName("lib.InterfaceWithNonObjectMethodOverrides")!!)
+        processDeclaration(resolver.getClassDeclarationByName("lib.AbstractClassWithObjectMethodOverrides")!!)
+        processDeclaration(resolver.getClassDeclarationByName("lib.ConcreteClassWithObjectMethodOverrides")!!)
+        processDeclaration(resolver.getClassDeclarationByName("lib.BaseWithProperties")!!)
+        processDeclaration(resolver.getClassDeclarationByName("lib.OverridesBaseWithProperties")!!)
+        processDeclaration(resolver.getClassDeclarationByName("lib.BaseWithIsProperties")!!)
+        // TODO: Enable this once we've sorted out the ordering between the AA and PSI implementations.
+        //  This should be a pretty rare edge case where a property is named isFoo and is overridden by a Java method
+        //  named isFoo().
+        // processDeclaration(resolver.getClassDeclarationByName("lib.OverridesBaseWithIsProperties")!!)
```
jaschdoc: Could we add an issue for these TODOs?

test-utils/testData/api/declarationsInClass.kt
```diff
@@ -0,0 +1,234 @@
...
+    public String getFoo() {
+        return "foo";
+    }
+
+    // Overrides val getBar
+    @Override
+    public String getGetBar() {
+        return "getBar";
+    }
+}
```
jaschdoc, line 189: Does this also test that writing a Java method `getFoo` that does not override a Kotlin `val foo` also works and is a valid method?

jaschdoc (dismissed):
This looks good! Thank you. My comments are mostly just nits and questions :)

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: Yup! Done.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: Luckily `declaredMemberScope.constructors` (and `declaredMemberScope.classifiers`) doesn't trigger the expensive computation. Only `declaredMemberScope.declarations` is the issue.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: Yeah, I'm not too concerned about the performance of this.

The `classifiers(name)` call should be O(1) because AA should already have these symbols cached by name, so it's just a key lookup in a map. Furthermore, the `find` call is only iterating over classes with the exact same name, so in general it should be size == 1, so I don't expect that to cause an issue either.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: Actually, I can just remove the comment and `.toList()` call. That was from an older implementation of the code, and it shouldn't be an issue anymore. In particular, calling `.partition` will only iterate over the sequence once and results in `Pair<List<T>, List<T>>` so calling `toList()` isn't necessary.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: In this case we're trying to find static "nested" (a.k.a. non-"inner") classes because those are grouped with the static members in AA ordering. 

The issue is that these classes aren't directly marked with the JAVA_STATIC modifier, instead we look the lack of an INNER modifier to know that it's a static nested class.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: Just to clarify, it's not that the slow path now accepts Java libs (it's always accepted Java libs), it's that the unwrapping of property back to a java getter/setter now happens for java libs. Previously, we were only doing the unwrapping for java sources, but that's inconsistent and actually a bug in KSP since that behavior shouldn't change just because we're processing as a Java source instead of a Java lib. FWIW, I needed to fix the bug in this PR because otherwise the PSI (fast path) output would not match the AA (slow path) output and it would make testing more difficult. I could split this into a separate PR if you want and just have two separate implementations of the test for PSI and AA like you did in https://github.com/google/ksp/pull/2921/changes, but I wanted to avoid that pain if possible

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: Done. I've added a test with TODOs because there's a couple issues related to this edge case.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: Done.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
bcorso: Done. I've made both classes private.

No good reason why they're different. Maybe I accidentally removed private while testing or something?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
bcorso: Done. Found that there actually already exists `JavaToKotlinClassMap.mapKotlinToJava` in AA to handle this.

test-utils/testData/api/declarationsInClass.kt
bcorso: Done.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/DeclarationsInClassProcessor.kt
bcorso: Done.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
bcorso: Nope, it shouldn't matter.

In particular, the "synthetic" here isn't referring to the property accessor itself (i.e. whether it's implicit vs explicit). Instead it represents whether a method in the java source/lib is overriding a property accessor in one of its super types. If it is overriding a property accessor in a super type then AA labels the origin of the method as "synthetic" so we do the same in the PSI implementation to match AA.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
bcorso: Done.

test-utils/testData/api/declarationsInClass.kt
bcorso: Good idea. I've added this to the test (works fine).

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/DeclarationsInClassProcessor.kt
bcorso: I was thinking it's best to keep them in their natural order so that we can also test that the natural declaration order of PSI matches the order given by AA. 

WDYT?

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/DeclarationsInClassProcessor.kt
bcorso: Done.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/DeclarationsInClassProcessor.kt
jaschdoc: Is the declaration order important for symbol / annotation processing unless explicitly calling a function that promises to return it in source order?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSFunctionDeclarationImpl.kt
jaschdoc: Okay, that makes sense. I'm used to synthetic meaning it was generated by the compiler. Thank you for the explanation.

test-utils/src/main/kotlin/com/google/devtools/ksp/processor/DeclarationsInClassProcessor.kt
bcorso: I'll sort the results for now.

I'm probably biased because with Javac/KAPT everything was just in source order by default so you didn't have to call a special function to order them.

OTOH, if we wanted to give up on making AA and PSI match, it would actually simplify the logic in `KSClassDeclarationImpl` where I'm jumping through a few hoops to get them to match. I'll save that for a separate CL though.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
jaschdoc: Thank you for the explanation. That makes sense. If I'm reading this correctly, then I suppose we already have a test for that case in this PR, right?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
jaschdoc: Yeah I was also thinking that the size of the list is quite small so it shouldn't really be a problem. Thanks for the comment.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSClassDeclarationImpl.kt
bcorso: Yes, that's correct. The new test I added in this PR will break if you remove the JAVA_LIB now.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
```diff
@@ -70,6 +93,82 @@ class KSPropertyDeclarationJavaImpl private constructor(val ktJavaFieldSymbol: K
...
+    }
+
+    override val origin: Origin
+        get() = parentDeclaration?.origin ?: Origin.JAVA
+
+    override val containingFile: KSFile?
+        get() = parentDeclaration?.containingFile
+
+    override val packageName: KSName
+        get() = parentDeclaration?.packageName ?: KSNameImpl.getCached("")
```
jaschdoc: Nit: I believe we can just use `parent` here instead of `parentDeclaration` since we have a handle on a non-null parent via the constructor.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
```diff
@@ -253,6 +253,12 @@ abstract class KSPUnitTestSuite(
         runTest("$UTIL_PATH/declarationsInAccessor.kt")
     }
 
+    @TestMetadata("declarationsInClass.kt")
+    @Test
+    fun testDeclarationsInClass() {
+        runTest("../test-utils/testData/api/declarationsInClass.kt")
```
jaschdoc: Nit: Use `UTIL_PATH` instead of `../test-utils/testData/api`. See test above :)

jaschdoc (dismissed):
Looks good. I left two minor comments, we can change in another PR.

jaschdoc:
@troelsbjerre could you take a quick look as well? Thanks!

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
bcorso: Done.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSPropertyDeclarationJavaImpl.kt
bcorso: Done. Thanks!

jaschdoc (approved):

troelsbjerre (approved):

jaschdoc:

PR #2879, opened 2026-04-21
Add fast type alias heuristic for `PsiResolutionStrategy`

jaschdoc:
This PR adds a heuristic to `PsiResolutionStrategy` to quickly determine whether a name is a type alias or not. It works by stitching together a short name with import qualifiers and asks the Analysis API whether this name is a type alias or not. If the situation is ambiguous, such as when a name might be shadowed, it falls back to expensive type resolution. @bcorso 

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/impl/PsiResolutionStrategy.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/impl/PsiResolutionStrategy.kt
@@ -17,9 +17,9 @@
 
 package com.google.devtools.ksp.common.impl
 
-import com.google.devtools.ksp.common.mergeMapNotNullKeys
 import com.google.devtools.ksp.common.visitor.CollectAnnotatedSymbolsPsiVisitor
 import com.google.devtools.ksp.containingFile
+import com.google.devtools.ksp.impl.FileCache
 import com.google.devtools.ksp.impl.symbol.kotlin.KSClassDeclarationImpl
 import com.google.devtools.ksp.impl.symbol.kotlin.KSFileImpl
 import com.google.devtools.ksp.impl.symbol.kotlin.KSFunctionDeclarationImpl
@@ -57,6 +57,12 @@ import com.intellij.psi.PsiTypeParameter
 import com.intellij.psi.PsiTypeParameterList
 import com.intellij.util.containers.addIfNotNull
 import org.jetbrains.kotlin.analysis.api.symbols.KaCallableSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaClassSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaTypeAliasSymbol
+import org.jetbrains.kotlin.analysis.api.types.symbol
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.KtAnnotated
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtDeclaration
@@ -64,7 +70,7 @@ import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtTypeReference
 import org.jetbrains.kotlin.psi.psiUtil.parameterIndex
 import org.jetbrains.kotlin.utils.addToStdlib.flatGroupBy
-import kotlin.collections.buildList
+import kotlin.collections.mapValues
 
 /**
  * An [AnnotationResolutionStrategy] that uses a combination of Psi and Kotlin's Analysis API to resolve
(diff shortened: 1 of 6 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/DeclarationNameVisitor.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.jetbrains.kotlin.psi.KtDeclaration
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtImportList
+import org.jetbrains.kotlin.psi.KtPackageDirective
+import org.jetbrains.kotlin.psi.KtTypeAlias
+
+/**
+ * A [PsiRecursiveElementWalkingVisitor] that collects the names of classes, objects and type aliases.
+ */
+class DeclarationNameVisitor : PsiRecursiveElementWalkingVisitor() {
```
bcorso: nit: `DeclarationNameVisitor` is a little ambiguous because there's many declarations other than class, object, and type aliases. 

Maybe`CollectClassifierNameVisitor` since in Koltin compiler terminology classes, interfaces, objects, and type aliases are collectively referred to as "classifiers".

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/visitor/DeclarationNameVisitor.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    private val internalResult = mutableSetOf<String>()
+    val result: Set<String> = internalResult
```
bcorso: style-nit: A common pattern to avoid exposing this property is to expose a companion object method, e.g.:

```kotlin
class DeclarationNameVisitor : PsiRecursiveElementWalkingVisitor() {
  private val result = mutableSetOf<String>()
  // ...
  companion object {
    fun get(file: KtFile) {
        val visitor = DeclarationNameVisitor()
        file.accept(visitor)
        return visitor.result
    }
  }
}
``` 

Another option (which might be more idiomatic for Kotlin) is to make the class itself private and expose a public top-level extension function to access the results, e.g.

```kotlin
private class DeclarationNameVisitor : PsiRecursiveElementWalkingVisitor() {
  val result = mutableSetOf<String>()
  // ...
}

fun KtFile.collectClassifierNames() {
  val visitor = DeclarationNameVisitor()
  accept(visitor)
  return visitor.result
}
```

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -247,15 +252,100 @@ class PsiResolutionStrategy(
...
-            .flatGroupBy { element -> element.annotationEntries }
-            .mapValues { entry ->
-                lazy { entry.value.flatMap { it.resolveToKSAnnotated(entry.key) } }
+    private val annotatedKotlinElementsByFullyQualifiedName: Map<String, List<Lazy<Collection<KSAnnotated>>>> by lazy {
+        buildMap<String, MutableList<Lazy<Collection<KSAnnotated>>>> {
+            val classIdCache = mutableMapOf<Pair<KtFile, String>, ClassId?>()
+            val fileCaches = mutableMapOf<KtFile, FileCache>()
+            annotatedKotlinElements.forEach { annotated ->
+                annotated.annotationEntries.forEach { annotationEntry ->
+                    val file = annotationEntry.containingKtFile
```
bcorso: Since all of the annotations are in the same containing file as the annotated element, we can pull `val file` declaration outside of this loop as `val file = annotated. containingKtFile` and just check it once per annotated element.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -247,15 +252,100 @@ class PsiResolutionStrategy(
...
+                    val text = annotationEntry.typeReference?.text ?: return@forEach
```
bcorso: The `text` can be a little error-prone. Let's use `annotationEntry.shortName` instead.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -247,15 +252,100 @@ class PsiResolutionStrategy(
...
+                    val fileCache = fileCaches.getOrPut(file) { FileCache(file) }
+                    val classId = classIdCache.getOrPut(file to text) {
+                        resolveAnnotationEntryClassId(annotationEntry, fileCache)
+                    }
+                    // N.B.: Skip nullable qualified names without error. This is what the AA-based implementation does.
+                    //   For now, this is the intended behavior. If it is changed in the future, then it is an API change.
+                    val fqn = classId?.asFqNameString() ?: return@forEach
+                    getOrPut(fqn, ::mutableListOf)
+                        .add(lazy { annotated.resolveToKSAnnotated(annotationEntry) })
```
bcorso: Let's collect the annotation entries here without resolving first. Then you can use the same `mapValues` transformation you were doing before on the entire list of annotation entries, i.e.:

```kotlin
.mapValues { entry ->
    lazy { entry.value.flatMap { it.resolveToKSAnnotated(entry.key) } }
}
```

This way the return type can stay `Map<String, Lazy<Collection<KSAnnotated>>>` rather than `Map<String, List<Lazy<Collection<KSAnnotated>>>>`

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -247,15 +252,100 @@ class PsiResolutionStrategy(
...
      * ```
      */
-    private val annotatedKotlinElementsByFullyQualifiedName: Map<String, Lazy<Collection<KSAnnotated>>> by lazy {
-        annotatedKotlinElements
-            .flatGroupBy { element -> element.annotationEntries }
-            .mapValues { entry ->
-                lazy { entry.value.flatMap { it.resolveToKSAnnotated(entry.key) } }
+    private val annotatedKotlinElementsByFullyQualifiedName: Map<String, List<Lazy<Collection<KSAnnotated>>>> by lazy {
+        buildMap<String, MutableList<Lazy<Collection<KSAnnotated>>>> {
+            val classIdCache = mutableMapOf<Pair<KtFile, String>, ClassId?>()
```
bcorso: In general, this cache is error-prone because it doesn't handle shadowing properly. In particular, it assumes that there is only ever one ClassId per file+text pair, but this isn't the case with shadowing, e.g. the following two annotations will have the same text but different ClassIds:

```kotlin
annotation class Foo

@Foo // Resolves to the top-level Foo
class Outer {
    annotation class Foo 

    @Foo // Resolves to the nested Outer.Foo
    fun doSomething() {}
}
```

We could start by just not caching this.

Once this is submitted, we could investigate the performance difference with caching enabled for cases where we know the short name isn't shadowed (using the FileCache.localDeclarations) and try caching those cases. However, I think most of those cases will just take the fast path so my guess is that caching these won't provide a huge benefit.

bcorso: (And sorry, I know this bug was also in my initial prototype).

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -247,15 +252,100 @@ class PsiResolutionStrategy(
...
+    private fun fastResolveClassId(annotationEntry: KtAnnotationEntry, fileCache: FileCache): ClassId? {
+        val shortName = annotationEntry.shortName?.asString() ?: return null
+        val unresolvedTypeName = annotationEntry.typeReference?.text ?: return null
+        val isPotentiallyShadowed = shortName in fileCache.localDeclarations
+
+        if (isPotentiallyShadowed) {
+            return null
+        }
+
+        if (shortName != unresolvedTypeName) {
```
bcorso: nit: The ` if (shortName != unresolvedTypeName)` check seems a bit out of order. Can we move this check before we even define `isPotentiallyShadowed`?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/PsiResolutionStrategy.kt
```diff
@@ -247,15 +252,100 @@ class PsiResolutionStrategy(
...
+            return null
+        }
+
+        val candidateClassIds = buildList {
```
bcorso: style-nit: Splitting the explicit import check might make it slightly more straight-forward, e.g.

```kotlin
val candidateClassIds = fileCache.explicitImports[shortName]?.let { explicitFqn ->
    // Direct name import found. Do not consider other imports.
    listOf(ClassId.topLevel(explicitFqn))
} ?: buildList {
    // No direct name import found. Consider package imports, star imports, and default imports.
    add(mkClassId(fileCache.packageName, shortName))
    fileCache.starImports.forEach { add(mkClassId(it, shortName)) }
    FileCache.DEFAULT_IMPORTS.forEach {
        add(mkClassId(FqName(it), shortName))
    }
}
```

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/impl/PsiResolutionStrategy.kt
```diff
@@ -247,15 +252,100 @@ class PsiResolutionStrategy(
...
+            }
+        }
+
+        candidateClassIds.forEach { classId ->
+            // Check whether it is a type alias
+            classId.toTypeAlias()?.let { typeAlias ->
+                // Fall back to expensive resolution
+                return analyze { typeAlias.expandedType.fullyExpandedType.symbol?.classId }
+            }
+            analyze { findClass(classId)?.classId }?.let {
```
bcorso: Since we need to call `analyze` no matter what, let's wrap the entire for-loop in the analyze block instead. I'm not sure how much it actually matters, but it seems better to open as few sessions as possible

bcorso (approved):
Thanks!

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/impl/PsiResolutionStrategy.kt
jaschdoc: I see what you mean. However, it seems to me that these two `analyze` blocks are mutually exclusive, so the current code keeps the `analyze` scope as small as possible.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/impl/PsiResolutionStrategy.kt
```diff
@@ -247,15 +252,106 @@ class PsiResolutionStrategy(
...
+        } ?: buildList {
+            // No direct name import found. Consider package imports, star imports, and default imports.
+            add(mkClassId(fileCache.packageName, shortName))
+            fileCache.starImports.forEach { add(mkClassId(it, shortName)) }
+            FileCache.DEFAULT_IMPORTS.forEach {
+                add(mkClassId(FqName(it), shortName))
+            }
+        }
+
+        var result: ClassId? = null
```
bcorso: Looks like there's actually three related APIs:

  * `findClass`: Returns non-type alias class, object, interface with given classId
  * `findTypeAlias`: Returns type alias with given classId
  * `findClassLike`: Returns either of the above with the given classId

Thus, I think we could just do something like:

```kotlin
return analyze {
    candidateClassIds.firstNotNullOfOrNull {
        when (val symbol = findClassLike(classId)) {
            // Type alias: fall back to expensive expansion
            is KaTypeAliasSymbol -> symbol.expandedType.fullyExpandedType.symbol?.classId

            // Normal class: there's no need to expand.
            is KaClassSymbol -> symbol.classId

            // No class with this class id.
            null -> null
        }
    }
}
```

WDYT?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/impl/PsiResolutionStrategy.kt
bcorso: Yeah, I meant to discuss this in our sync today.

I definitely see what you're saying, and I think there's some trade-off in either case. I'm not sure what's the best approach yet. Moving the `analyze` blocks to a smaller scope is definitely better if we can avoid calling them altogether. However, we could also get into a case where if we move them into too small of a scope we accidentally start creating and tearing down analysis sessions in a hot-loop rather than starting a single session for the entire loop. I think we should look more into this, and maybe check with JetBrains on how they intended this to be used.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/impl/PsiResolutionStrategy.kt
jaschdoc: That's great, thanks for the comment!

bcorso (approved):
LGTM. Thanks!

troelsbjerre (dismissed):

jaschdoc:
@troelsbjerre, rebased on main. Only change since last review :)

troelsbjerre (approved):

PR #2881, opened 2026-04-22
feat: Update Gradle and AGP versions

hfmehmed:
* This commit updates the Gradle wrapper and Android Gradle Plugin (AGP) versions across the project.
* Sets the min runtime version of AGP allowed to 8.10.0 (which was released one 1 year ago)
* Cleans up code that depends on AGP version < 8.10.0 which can no longer be true


troelsbjerre (dismissed):

jaschdoc (approved):
Thanks, looks good!

PR #2882, opened 2026-04-22
Enable project ksp project isolation when org.gradle.unsafe.isolated-projects=true

### LINKED ISSUE #2866: Enable ksp.project.isolation.enabled by default if org.gradle.unsafe.isolated-projects=true
Enable ksp.project.isolation.enabled by default if org.gradle.unsafe.isolated-projects=true.

We should enable this flag automatically if user already expressed the intent to use this.

hfmehmed:
Fixes #2866

### DIFF
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/ProjectIsolationIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/ProjectIsolationIT.kt
@@ -1,0 +1,40 @@
+package com.google.devtools.ksp.test
+
+import com.google.devtools.ksp.test.fixtures.TemporaryTestProject
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+import java.io.File
+import java.util.jar.JarFile
+
+class ProjectIsolationIT {
+    @Rule
+    @JvmField
+    val project: TemporaryTestProject = TemporaryTestProject(
+        "playground",
+        experimentalPsiResolution = true
+    )
+
+    @Test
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/ProjectIsolationIT.kt
```diff
@@ -0,0 +1,40 @@
...
+import java.io.File
+import java.util.jar.JarFile
+
+class ProjectIsolationIT {
+    @Rule
+    @JvmField
+    val project: TemporaryTestProject = TemporaryTestProject(
+        "playground",
+        experimentalPsiResolution = true
+    )
```
jaschdoc, line 18: Shouldn't this class be parameterized as well?

```kt
@RunWith(Parameterized::class)
class ProjectIsolationIT(experimentalPsiResolution: Boolean) {
    @Rule
    @JvmField
    val project: TemporaryTestProject = TemporaryTestProject(
        "playground",
        experimentalPsiResolution = experimentalPsiResolution
    )
```

jaschdoc (commented):
Thanks, looks good! Just one minor comment :)

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/ProjectIsolationIT.kt
hfmehmed: I am not sure to be honest. These integration tests add non trivial amount of time to the total CI time. Is there any benefit to running all our integration tests with this experimental resolution enabled and disabled. I think only couple of tests should run with this experiment. If we parameterize it in all our tests i think we should remove that in fact.

mohamedkabaka922-lang (approved):

PR #2896, opened 2026-04-27
Memoize KSTypeImpl.hashCode()

### LINKED ISSUE #2576: [KSP2][Performance]: Memoize `KSTypeImpl.fullyExpand()` or `KSTypeImpl.hashCode()`?
In Dagger we found a particularly slow part of our build was creating our [`KspArrayType$Factory`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspArrayType.kt;l=115-131;bpv=1;bpt=0;drc=e6d33dd5d0a60001a5784d84123b05308d35f410)

Image

The pprof above shows that this issue is due to creating the `reverseBuiltInArrayLookup` map (shown below for convience):

```kotlin
private val builtInArrays =
    mapOf(
        "kotlin.BooleanArray" to KspPrimitiveType(env, env.resolver.builtIns.booleanType),
        "kotlin.ByteArray" to KspPrimitiveType(env, env.resolver.builtIns.byteType),
        "kotlin.CharArray" to KspPrimitiveType(env, env.resolver.builtIns.charType),
        "kotlin.DoubleArray" to KspPrimitiveType(env, env.resolver.builtIns.doubleType),
        "kotlin.FloatArray" to KspPrimitiveType(env, env.resolver.builtIns.floatType),
        "kotlin.IntArray" to KspPrimitiveType(env, env.resolver.builtIns.intType),
        "kotlin.LongArray" to KspPrimitiveType(env, env.resolver.builtIns.longType),
        "kotlin.ShortArray" to KspPrimitiveType(env, env.resolver.builtIns.shortType),
    )

// map from the primitive to its array
private val reverseBuiltInArrayLookup =
    builtInArrays.entries.associateBy { it.value.ksType }
```

While we ended up fixing this performance issue by caching the factory instance (which removes nearly the entire 63s), it also would have helped to have more caching at the KSP level.

For example, given that the majority of time is spent calculating hashcodes for the same 8 builtin types above, memoizing the result of `KSTypeImpl.hashcode()` or `KSTypeImpl.fullyExpand()` would have helped here, assuming these types are indeed immutable (i.e. there hashes don't change over time).

shmuelr:
Cache `KSTypeImpl.hashCode()` to avoid reopening an Analysis API session on every call.

  `hashCode()` previously called `type.fullyExpand().hashCode()` which opens a fresh `analyze { }` session per invocation. Because `KSTypeImpl` is used as a HashMap key in hot paths
  (`propertyAsMemberOfCache`, `functionAsMemberOfCache`), this fires hundreds of thousands of times per KSP task.

  **Fix:** `private val cachedHashCode: Int by lazy { type.fullyExpand().hashCode() }`

  ## Benchmark

  Integration bench: full `KotlinSymbolProcessing.execute()` over 40 test classes (generics, type aliases, inheritance chains), with the processor calling `.hashCode()` 1.2M times per run in a tight loop. 5 warmup iterations, 8 stable measurement iterations, fresh JVM per variant.

  | Variant | Median | Delta |
  |---------|--------|-------|
  | No cache (baseline) | ~400 ms | — |
  | `by lazy { }` | ~130 ms |  -68% |

  Consistent across 8 independent runs in varied execution order. The three cached strategies tested (two-field primitive, `lazy(NONE)`, `lazy {}`) were indistinguishable at wall-clock, within ±15ms of each other across all runs.

Fixes https://github.com/google/ksp/issues/2576

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
@@ -36,6 +36,16 @@ import org.jetbrains.kotlin.analysis.api.symbols.*
 import org.jetbrains.kotlin.analysis.api.types.*
 
 class KSTypeImpl private constructor(internal val type: KaType) : KSType {
+
+    // Per github.com/google/ksp/issues/2576: fullyExpand() opens a fresh
+    // Analysis API session every call, and hashCode() is called by every
+    // HashMap/HashSet containing this KSTypeImpl.
+    //
+    // Idempotent computation makes any lost write safe - 
+    // a racing thread just recomputes the same value.
+    private var cachedHash: Int = 0
+    private var cachedHashIsZero: Boolean = false
+
     companion object : KSObjectCache<IdKey<KaType>, KSTypeImpl>() {
         fun getCached(type: KaType): KSTypeImpl = cache.getOrPut(IdKey(type)) {
             KSTypeImpl(type)
@@ -162,7 +172,16 @@ class KSTypeImpl private constructor(internal val type: KaType) : KSType {
         get() = type is KaFunctionType && type.isSuspend
 
     override fun hashCode(): Int {
-        return type.fullyExpand().hashCode()
+        var h = cachedHash
+        if (h == 0 && !cachedHashIsZero) {
+            h = type.fullyExpand().hashCode()
+            if (h == 0) {
+                cachedHashIsZero = true
+            } else {
+                cachedHash = h
+            }
+        }
+        return h
     }
 
     override fun equals(other: Any?): Boolean {
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSTypeImpl.kt
@@ -161,9 +161,11 @@ class KSTypeImpl private constructor(internal val type: KaType) : KSType {
     override val isSuspendFunctionType: Boolean
         get() = type is KaFunctionType && type.isSuspend
 
-    override fun hashCode(): Int {
-        return type.fullyExpand().hashCode()
+    private val cachedHashCode: Int by lazy {
+        type.fullyExpand().hashCode()
     }
+
+    override fun hashCode(): Int = cachedHashCode
 
     override fun equals(other: Any?): Boolean {
         if (other !is KSTypeImpl) {
--- a/test-utils/src/main/kotlin/com/google/devtools/ksp/processor/TypeAliasProcessor.kt
+++ b/test-utils/src/main/kotlin/com/google/devtools/ksp/processor/TypeAliasProcessor.kt
@@ -57,6 +57,19 @@ open class TypeAliasProcessor : AbstractTestProcessor() {
                 assert(sameTypeAliases.map { it.hashCode() }.distinct().size == 1) {
                     "different hashcodes for members of $signature"
                 }
+                // Hash stability: repeated hashCode() calls on the same instance
+                // must return the same value. Guards the KSTypeImpl hashCode
+                // cache (github.com/google/ksp/issues/2576) against regressions
+                // where a second call recomputes a different value or fails to
+                // use the cached result.
+                for (t in sameTypeAliases) {
+                    val first = t.hashCode()
+                    val second = t.hashCode()
+                    val third = t.hashCode()
+                    assert(first == second && second == third) {
+                        "hashCode not stable for $t: $first, $second, $third"
+                    }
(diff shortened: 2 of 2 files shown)


jaschdoc:
Hi @shmuelr, thank you for the pull request! Do you have any data that shows this is implementation is preferable to just wrapping it in lazy val? If not, let's start with that, since it's a bit simpler and then we can always move to a in-depth implementation if the numbers show that. Cheers! :)

shmuelr:
> Hi @shmuelr, thank you for the pull request! Do you have any data that shows this is implementation is preferable to just wrapping it in lazy val? If not, let's start with that, since it's a bit simpler and then we can always move to a in-depth implementation if the numbers show that. Cheers! :)

Thanks for the review @jaschdoc!

I mirrored the Java String.hashcode and avoided Kotlin's Lazy to avoid the Volatile field & Synchronized block in LazyJVM.kt, given that the hashcode is idempotent it felt overkill to have thread safety here. 

I'm happy to benchmark this, what's the best way to go about that? (I've been testing this in a primary project I'm working on, but it's a huge project and susceptible to noise)

jaschdoc:
Aha, I think `lazy` is the best way to go if we don't know the impact yet. In terms of benchmarking, there's not really a good answer there. This is also something we are evaluating, so I cannot help you with that. But any performance numbers you can provide in a reproducible way would be a great help. I know it might be a big ask, but performance improvements are always tricky if they cannot be measured.

I will leave the PR open and come back if I figure out a way to measure the impact of this in a stable way.

shmuelr:
I took a pass at benchmarking this, I ran the KSP test suite (along with an added 1000 iteration loop) and measuring the timing of the calls (3 warmup passes, then 10 measurement passes)

The raw measurements of the hashcode showed the string.hashcode pattern (with the int field) to be 5x faster then lazy(NONE), but this was nanoseconds difference on my machine Macbook M2 pro (0.39 vs 2.07 ns/call). (this makes sense to me, the direct int usage vs the calls to UnsafeLazyImpl.value + Int? boxing). 

But, in the end this only accounted for 1% difference in the hashcode bechmarking speed. The string pattern was 38% faster than the current hashcode calculations, and lazy(None) was 37% faster.

I'm open to changing this to Lazy(None) { } if you want, personally I'd rather get the extra performance but I do understand there is more cognitive load & maintenance for this code now.

jaschdoc:
Thanks for the measurements! Could you clarify with code snippets exactly what each implementation consists of? It's also unclear to me if you benchmarked the baseline / current implementation.

shmuelr:
I ran a test locally that sets up these types and calls hashcode in a loop. I pulled the latest codebase as the baseline and compared to 3 variants
1. Baseline
2. The String.hashcode pattern (this pr) which uses the int field to store the hash
3. A Lazy.None approach, https://github.com/google/ksp/commit/47f192986c86a4ba8aac6082e514de4099f72ad3 (this is what I assumed you meant by using a lazy val)

Hammering it now a bunch, 2 & 3 seem to be neck & neck (~1-3% variance) with the winner alternating. 
I think I'm just gonna go with the lazy val, it doesn't seem like theres a material difference in practice

jaschdoc:
Cool. What is the performance of the default constructor for `lazy`? I.e.,
```kotlin
private val cachedHashCode: Int by lazy {
  // fully expanded type etc...
}
```

shmuelr:
It's not super clear to me, the initial runs showed a ~10% slowdown with lazy {} vs lazy(None) {}, but the more I run this the more the gap closes (feels like JVM hotspot optimizations are helping here in the test loops)

jaschdoc:
I see. Let's go with the `lazy` block with the default constructor for now. It can be really tricky to benchmark the performance of JVM programs without a proper setup exactly due to JIT compilation. This is still something we can improve in the future when there's better metrics for performance. Did you have numbers for how it compares to the baseline? A table with each implementation, it's min, mean, and max timings (after warmup) as well as a relative improvement over the baseline (three columns total) would be really helpful!

shmuelr:
> I see. Let's go with the `lazy` block with the default constructor for now. It can be really tricky to benchmark the performance of JVM programs without a proper setup exactly due to JIT compilation. This is still something we can improve in the future when there's better metrics for performance. Did you have numbers for how it compares to the baseline? A table with each implementation, it's min, mean, and max timings (after warmup) as well as a relative improvement over the baseline (three columns total) would be really helpful!

Sgtm, updated code & pr description, thank you for the review!

jaschdoc (approved):
Thank you for the patch! Could I ask you to squash the changes into a single commit?

jaschdoc (dismissed):

jaschdoc:
@shmuelr, could you please fix linting errors and squash it into the same commit? You can run all checks locally with `./gradlew check`. It might take a while but then you can be sure everything is good to go.

shmuelr:
ah, that's what I missed, all good now 👍 , thanks @jaschdoc !

jaschdoc (approved):

jaschdoc:
Congrats on landing the PR! 🎉

PR #2910, opened 2026-05-01
Fix results for non-repeatable repeated annotations

### LINKED ISSUE #2919: Only return symbols once when annotated with repeated non-repeatable annotations
If an annotation is repeated more than once on a declaration but the annotation itself is not annotated with `@Repeatable`, we should only return the declaration once. The Analysis API already does this, but the Psi implementation does not.

Bug shown in https://github.com/google/ksp/pull/2917

jaschdoc:
Depends on https://github.com/google/ksp/pull/2908

- [x] Use test setup for failures introduced in https://github.com/google/ksp/pull/2911
- [x] Add issue for this bug

Fixes https://github.com/google/ksp/issues/2919


jaschdoc:
See https://github.com/google/ksp/pull/2917

troelsbjerre (approved):

PR #2911, opened 2026-05-01
Add negative tests for `all` meta-use-site target

jaschdoc:
Depends on https://github.com/google/ksp/pull/2915

This PR adds the following:
- Adds a *failing* test for invalid use of the `all` use-site target on grouped annotations (see [KEEP 402](https://github.com/Kotlin/KEEP/blob/74a4b4a4937a0a9a70afff71276044c731c0e20b/proposals/KEEP-0402-annotation-target-in-properties.md?plain=1#L107))
- Adds a `runFailingTest` function to `AbstractKSPTest` that asserts the expected output does **not** match the actual output
- Adds a `runThrowingTest` function to `AbstractKSPTest` that asserts the test throws an `InternalKSPException`. This can be expanded to assert other types of exceptions, but for now the idea is that it forces throwing expressions to use `InternalKSPException` which provides an error message with a bit more information than what one might provide with `error` or `require`.

- [x] Refactor `runTest`, `runFailingTest`, `runThrowingTest` to reuse test setup code.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/visitor/CollectAnnotatedSymbolsPsiVisitor.kt
@@ -1,5 +1,6 @@
 package com.google.devtools.ksp.common.visitor
 
+import com.google.devtools.ksp.InternalKSPException
 import com.google.devtools.ksp.impl.symbol.kotlin.toLocation
 import com.intellij.psi.PsiAnnotation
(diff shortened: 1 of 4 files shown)


jaschdoc:
Btw, since this PR was opened, a code/execution path that previously threw an exception has been fixed with a proper implementation on `main`. That leaves the `runThrowingTest` unused, but I have seen the implementation of `runThrowingTest` succeed by inserting exceptions (both being green by using InternalKSPException and being red under no exception and just `kotlin.Exception`).

troelsbjerre (dismissed):

jaschdoc:
Rebased on main

troelsbjerre (approved):

PR #2915, opened 2026-05-04
Extract const vals for test paths

jaschdoc:
Introduce `const val`s `AA_PATH` and `UTIL_PATH` for path prefixes used in tests. Makes it less error-prone to add new tests.


troelsbjerre (dismissed):
UTIL_PATH could even include `/api` here.

jaschdoc:
> UTIL_PATH could even include `/api` here.

Yes, thanks for spotting that!

troelsbjerre (approved):

Issue #2913, opened 2026-05-01
experimentalPsiResolution fails with "no element found"


bcorso:
Consider a case where the user's code has a slight bug where it puts an incorrect use-site `@field:` on a constructor parameter that doesn't use `val` or have a backing field:

```kotlin
annotation class MyAnnotation

class MyClass(
    @field:MyAnnotation param: String // <- no backing field
)
```

Normally in this case, the annotation is just skipped/dropped, but in `experimentalPsiResolution` mode it currently fails compilation with the following error:

```
ERROR: unexpected element of type: no element found

Attachments:
info.txt
- ktElement:
    Class: ksp.org.jetbrains.kotlin.psi.KtTypeReference
Value:
    <File name: MyClass.kt, Physical: true>
constructor(
  @field:<ELEMENT>MyAnnotation</ELEMENT> param: String
)

========

ksp.org.jetbrains.kotlin.analysis.low.level.api.fir.api.InvalidFirElementTypeException: unexpected element of type: no element found
	at ksp2_impl.jar//ksp.org.jetbrains.kotlin.analysis.api.fir.components.KaFirTypeProvider.handleUnexpectedFirElementError(KaFirTypeProvider.kt:208)
	at ksp2_impl.jar//ksp.org.jetbrains.kotlin.analysis.api.fir.components.KaFirTypeProvider.getType(KaFirTypeProvider.kt:197)
	at ksp2_impl.jar//ksp.org.jetbrains.kotlin.analysis.api.impl.base.KaBaseSession.getType(KaBaseSession.kt)
	at ksp2_impl.jar//ksp.com.google.devtools.ksp.common.impl.PsiResolutionStrategy.getQualifiedName(PsiResolutionStrategy.kt:537)
	at ksp2_impl.jar//ksp.com.google.devtools.ksp.common.impl.PsiResolutionStrategy.annotatedKotlinElementsByFullyQualifiedName_delegate$lambda$0$2(PsiResolutionStrategy.kt:258)
	at ksp2_impl.jar//ksp.com.google.devtools.ksp.common.MapUtilsKt.mergeMapNotNullKeys(MapUtils.kt:36)
	at ksp2_impl.jar//ksp.com.google.devtools.ksp.common.impl.PsiResolutionStrategy.annotatedKotlinElementsByFullyQualifiedName_delegate$lambda$0(PsiResolutionStrategy.kt:258)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:86)
[...]

PR #2921, opened 2026-05-05
Add test for field and property use-sites on constructor parameters

jaschdoc:
Adds a test that reproduces #2913


troelsbjerre (dismissed):

jaschdoc:
Rebased on main

troelsbjerre (approved):

PR #2920, opened 2026-05-04
Split tests into smaller suites for CI

jaschdoc:
This PR splits integration tests into two tasks 'primary' and 'secondary'. The primary test suite blocks merges whereas the 'secondary' test suite is allowed to fail without blocking merges.
Additionally, the CI runs the following in parallel now:
- API check
- Unit tests
- Primary integration tests
- Secondary integration tests (now includes agp compatibility tests)

This should hopefully reduce CI time by a bit.

PR #2923, opened 2026-05-05
Separate AGP tests from secondary integration tests

jaschdoc:
Splits AGP compatibility tests into its own package since it's slow on Windows targets. This change should hopefully reduce max time spent in CI.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -11,9 +11,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


jaschdoc:
Longest CI job in this PR is 43 minutes. On `main` it is 70 minutes. A 27 minute improvement. If you count this PR and https://github.com/google/ksp/pull/2920, then it's a full hour faster than CI at commit 43181767c13c2c1c69e66b7a508cc8b834726b13

troelsbjerre (approved):

Issue #2271, opened 2024-12-26
[ksp2] Unexpected symbol type class org.jetbrains.kotlin.analysis.api.fir.symbols.KaFirEnumEntrySymbol


acejingbo:
For a Enum in library, like 
```
enum class SomeEnumInLib {
    ONE,
    TWO,
    THREE,
}
```
if you query `SomeEnumInLib.ONE` and ask for its `modifiers` it will crash with `IllegalStateException: Unexpected symbol type class KaFirEnumEntrySymbol`

# Repro

https://github.com/acejingbo/ksp/commit/f370e4c7bbb2d60d088ad635422b9b3da33431b7

See `resolver.getClassDeclarationByName("SomeEnumInSrc.ONE")!!.modifiers}` did not crash
but `resolver.getClassDeclarationByName("SomeEnumInLib.ONE")!!.modifiers}` crash with
```
java.lang.IllegalStateException: Unexpected symbol type class org.jetbrains.kotlin.analysis.api.fir.symbols.KaFirEnumEntrySymbol
	at com.google.devtools.ksp.impl.symbol.kotlin.AbstractKSDeclarationImpl$modifiers$2.invoke(AbstractKSDeclarationImpl.kt:67)
	at com.google.devtools.ksp.impl.symbol.kotlin.AbstractKSDeclarationImpl$modifiers$2.invoke(AbstractKSDeclarationImpl.kt:59)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:83)
	at com.google.devtools.ksp.impl.symbol.kotlin.AbstractKSDeclarationImpl.getModifiers(AbstractKSDeclarationImpl.kt:59)
	at com.google.devtools.ksp.processor.JavaModifierProcessor.process(JavaModifierProcessor.kt:61)
	at com.google.devtools.ksp.impl.KotlinSymbolProcessing$execute$1$1.invoke(KotlinSymbolProcessing.kt:549)
	at com.google.devtools.ksp.impl.KotlinSymbolProcessing$execute$1$1.invoke(KotlinSymbolProcessing.kt:547)
	at com.google.devtools.ksp.common.IncrementalContextBase.closeFilesOnException(IncrementalContextBase.kt:409)
	at com.google.devtools.ksp.impl.KotlinSymbolProcessing.execute(KotlinSymbolProcessing.kt:547)
	at com.google.devtools.ksp.test.AbstractKSPAATest.runTest(AbstractKSPAATest.kt:142)
[...]

PR #2929, opened 2026-05-06
Add test to reproduce issue 2271

jaschdoc:
Reproduces the issue in #2271


kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/processor/EnumModifierProcessor.kt
jaschdoc, line 1: Don't we need a copyright header here? None of the processors seem to have that.

troelsbjerre: Yes, we should.

troelsbjerre (dismissed):

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/processor/EnumModifierProcessor.kt
jaschdoc: Fixed!

troelsbjerre (approved):

PR #2935, opened 2026-05-07
Override `toString` in `InternalKSPException`

jaschdoc:
As suggested by the comment, this prevents the big error message from being printed in the stack trace and ruining the usability of it.

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/errors/InternalKSPException.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/errors/InternalKSPException.kt
@@ -46,6 +46,15 @@ internal class InternalKSPException(
         appendLine("   |")
     }
 ) {
+    override fun toString(): String {
+        // N.B.: Override the toString method to prevent the big error message being printed in the stack trace.
+        return buildString {
+            append(javaClass.name)
+            append(": ")
+            append("Internal KSP Error")
+        }
+    }
+
     internal companion object {
         private fun Location.render(): String {
             return when (this) {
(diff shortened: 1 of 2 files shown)


api/src/main/kotlin/com/google/devtools/ksp/errors/InternalKSPException.kt
```diff
@@ -30,7 +30,7 @@ internal class InternalKSPException(
     message: String,
     val location: Location,
     val originatingClass: Class<*>
-) : Exception(
+) : Throwable(
```
troelsbjerre: Should not extend Throwable directly

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/InternalKSPException.kt
```diff
@@ -33,7 +33,7 @@ internal class InternalKSPException(
     message: String,
     val location: Location,
     val originatingClass: Class<*>
-) : Exception(
+) : Throwable(
```
troelsbjerre: Should not extend Throwable directly

troelsbjerre (changes requested):

api/src/main/kotlin/com/google/devtools/ksp/errors/InternalKSPException.kt
jaschdoc: Fixed

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/InternalKSPException.kt
jaschdoc: Fixed

troelsbjerre (approved):

Issue #2873, opened 2026-04-20
Add support for explicit backing fields


jaschdoc:
See https://youtrack.jetbrains.com/projects/KT/issues/KT-82993/Support-explicit-backing-fields-in-the-Analysis-API
And [KEEP 430](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0430-explicit-backing-fields.md)

PR #2941, opened 2026-05-11
Add negative test for explicit backing fields

jaschdoc:
This PR adds a test for #2873, asserting that it is not yet implemented. An implementation will be added in a subsequent PR.

See KEEP 430 for further details of this feature (also linked in #2873)

### DIFF
--- a/kotlin-analysis-api/testData/getSymbolsWithAnnotation/explicitBackingFields.kt
+++ b/kotlin-analysis-api/testData/getSymbolsWithAnnotation/explicitBackingFields.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2026 Google LLC
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TEST PROCESSOR: GetSymbolsWithAnnotationProcessor
+// PROCESSOR INPUT: Anno
+// EXPECTED:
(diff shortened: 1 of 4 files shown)


jaschdoc:
@troelsbjerre ready for review now, apologies! :)

kotlin-analysis-api/testData/getSymbolsWithAnnotation/explicitBackingFields.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    val propertyWhereFieldIsDirectlyAnnotated: List<String>
+        @Anno field = MutableList<String>
+
+    @field:Anno
+    val annotatedPropertyWithFieldTarget: List<String>
+        field = MutableList<String>
+
+    @field:Anno
+    @property:Anno
```
troelsbjerre, line 48: This still only shows up once...

troelsbjerre (dismissed):

kotlin-analysis-api/testData/getSymbolsWithAnnotation/explicitBackingFields.kt
jaschdoc: That's correct. In the ideal world, it should be returned twice or once as `annotatedPropertyAndFieldViaUseSiteTargets` and once as `annotatedPropertyAndFieldViaUseSiteTargets.field`. The current behavior is due to the fact that KSP uses the property as a stand-in for the field and deduplicates symbols. It's a limitation, but I'm working a proper solution. This test only asserts that it works with the current implementation.

jaschdoc:
Rebased on main to resolve conflicts.

kotlin-analysis-api/testData/getSymbolsWithAnnotation/explicitBackingFields.kt
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+// TEST PROCESSOR: GetSymbolsWithAnnotationProcessor
+// PROCESSOR INPUT: Anno
+// EXPECTED:
+// Anno: MyClass.annotatedProperty
+// Anno: MyClass.annotatedPropertyAndFieldViaUseSiteTargets
+// Anno: MyClass.annotatedPropertyAndFieldViaUseSiteTargets.field
+// Anno: MyClass.annotatedPropertyWithFieldTarget
+// Anno: MyClass.propertyWhereFieldIsDirectlyAnnotated
```
troelsbjerre: .field, right?

troelsbjerre (changes requested):

troelsbjerre (approved):

PR #2953, opened 2026-05-18
Clean up how KSP checks for klib cross-compilation support

omarismail94:
Querying KGP's `kotlinCompilation.crossCompilationSupported` eagerly writes a metadata JSON file inside the build directory during task graph serialization. This dynamic file creation invalidates the Gradle Configuration Cache on subsequent runs.

To fix:
1. Consolidated native cross-compilation support checking on KGP's standard kotlin.native.enableKlibsCrossCompilation property as the single source of truth.
2. Swapped the operands in the task `onlyIf` check so host native compiles (like linuxX64 on Linux) short-circuit immediately and skip querying the cross-compilation property entirely.
3. Added two integration tests to verify that the configuration cache now successfully reuses on native compilations, and that subproject DSL overrides are correctly captured and respected.

CC: @eygraber @liutikas

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
@@ -77,6 +77,7 @@ import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeCompilation
 import org.jetbrains.kotlin.gradle.tasks.AbstractKotlinCompileTool
 import org.jetbrains.kotlin.gradle.tasks.KotlinNativeCompile
 import org.jetbrains.kotlin.konan.target.HostManager
+import org.jetbrains.kotlin.tooling.core.KotlinToolingVersion
 import java.io.ByteArrayOutputStream
 import java.io.File
 import java.io.ObjectOutputStream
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -446,25 +443,23 @@ abstract class KspAATask @Inject constructor(
...
-
-            return if (isSupportCrossCompilationPropertyAvailable) {
-                kotlinCompilation.crossCompilationSupported
-            } else {
-                // Fallback for Kotlin versions before 2.3.20-Beta2
-                project.providers.gradleProperty(
-                    "kotlin.native.enableKlibsCrossCompilation"
-                ).orElse("false").map { it.toBoolean() }
-            }
+            val projectValue = project.findProperty("kotlin.native.enableKlibsCrossCompilation")?.toString()
```
hfmehmed: are the changes here necessary? and are the equivalent to each other. e.g if i set `kotlinCompilation.crossCompilationSupported = true` is this going to return a true provider?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -420,14 +419,12 @@ abstract class KspAATask @Inject constructor(
...
-                            kotlinCompilation
-                        )
+                        val enableKlibsCrossCompilation = getKlibCrossCompilationSupport(project, kotlinCompilation)
+
                         kspAATask.onlyIf {
-                            isKlibCrossCompilationEnabled.get() || HostManager().enabled.any {
+                            HostManager().enabled.any {
                                 it.name == konanTargetName
-                            }
+                            } || enableKlibsCrossCompilation.getOrElse(false)
```
hfmehmed: out of curiosity was the switch between the 2 conditions needed or is this purely aestetics change?

omarismail94: I got rid of it, but i had it check host manager first so that if that fails, it can just short-circut the code without having to evaluate `enableKlibsCrossCompilation`

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
omarismail94: Reverted

hfmehmed (approved):

Issue #2940, opened 2026-05-11
Remove `test-utils` directory


jaschdoc:
Figure out if each test in `test-utils` should be removed or added to `kotlin-analysis-api/testData`. This is related to https://github.com/google/ksp/issues/2916.

PR #2954, opened 2026-05-19
Add unused test `allFunctions_kotlin`

jaschdoc:
This PR adds a test that was unused from the `test-utils` directory at moves it to the `kotlin-analysis-api/testData` directory.. It seems like there was a unit test that was meant to test it but specified a different file path.

Related to https://github.com/google/ksp/issues/2940

### DIFF
--- a/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
+++ b/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
@@ -64,7 +64,7 @@ abstract class KSPUnitTestSuite(
     @TestMetadata("allFunctions_kotlin.kt")
     @Test
     fun testAllFunctions_kotlin() {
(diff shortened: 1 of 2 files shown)


kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt
```diff
@@ -64,7 +64,7 @@ abstract class KSPUnitTestSuite(
     @TestMetadata("allFunctions_kotlin.kt")
     @Test
     fun testAllFunctions_kotlin() {
-        runTest("$AA_PATH/allFunctions_java_inherits_kt.kt")
```
jaschdoc, line 67: This test is actually tested in the test above this one `testAllFunctions_java_inherits_kt`: https://github.com/google/ksp/blob/b19ad1c2d97d42e1439c1b8199c1fb374c7247d9/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt#L61

hfmehmed (approved):

PR #2955, opened 2026-05-20
Support suggested fixes for error and warning reporting (#2551)

### LINKED ISSUE #2551: Support suggested fixes
Along with supporting more specific node locations (#2550) it would be very useful to have the ability to suggest replacement code for nodes. This is something that can be done with Android Lint, but for KSP processors that need to do validation of user code and report errors it would be extremely valuable to not have to rely on an external tool when the processor itself has the information it needs to suggest a fix.

KSP doesn't model the whole syntax tree so I would't expect it to be able to do every kind of suggested fix one might want, but I believe there are still many things that _could_ be done through KSP.

I have a use-case, for example, where I need to prevent use of specific annotations, but for each I could suggest a different annotation to replace it with. Right now I can report an error with a message that suggests the replacement, but that isn't something tools could provide a way of automatically applying.

HMH-3080:
### Summary

Adds `KSPSuggestedFix` to KSP so processors can suggest replacement code alongside error/warning messages. This enables IDE tools (IntelliJ/Android Studio) to offer Quick Fixes directly from KSP diagnostics, similar to how Android Lint works.

### Changes

**New API (`:api`)**
- `KSPSuggestedFix` — new `data class` carrying `replacementText` and optional `description`
- `KSPLogger` — added default `warn(message, symbol, fix)` and `error(message, symbol, fix)` overloads (backward compatible)

**Logger implementations**
- `KspGradleLogger` — overrides `warn`/`error` with fix, prints as:
  `w: [ksp] <message> -> Suggested Fix (<desc>): [<replacement>]`
- `CommandLineKSPLogger` — same, writes to `System.err` with `Warning:` / `Error:` prefix

**Bug fix (`:kotlin-analysis-api`)**
- `KotlinSymbolProcessing` — logger wrapper was missing overrides for the new `fix` overloads, so `hasError` was never set when calling `logger.error(..., fix)`. Fixed by adding the missing `error`/`warn` overrides with proper `hasError` tracking.

**API signature (`:api`)**
- Updated `api.base` with `component1()`, `component2()`, `copy()` for `KSPSuggestedFix`

**Tests**
- Unit tests: `KspGradleLoggerTest` (4 tests) + `CommandLineKSPLoggerTest` (4 tests)
- End-to-end test: `SuggestedFixProcessor` + `suggestedFix.kt` test data — validates the full pipeline (processor → logger → exit code)
- All tests pass for both AA and Psi configurations

### Usage example

```kotlin
env.logger.error(
    "This annotation is deprecated",
    declaration,
    KSPSuggestedFix("@NewAnnotation", "Replace with @NewAnnotation")
)

### DIFF
--- a/api/api.base
+++ b/api/api.base
@@ -141,10 +141,23 @@ package com.google.devtools.ksp.processing {
 
   public interface KSPLogger {
     method public void error(@NonNull String message, @Nullable com.google.devtools.ksp.symbol.KSNode symbol);
+    method public default void error(@NonNull String message, @Nullable com.google.devtools.ksp.symbol.KSNode symbol, @NonNull com.google.devtools.ksp.processing.KSPSuggestedFix fix);
     method public void exception(@NonNull Throwable e);
     method public void info(@NonNull String message, @Nullable com.google.devtools.ksp.symbol.KSNode symbol);
     method public void logging(@NonNull String message, @Nullable com.google.devtools.ksp.symbol.KSNode symbol);
     method public void warn(@NonNull String message, @Nullable com.google.devtools.ksp.symbol.KSNode symbol);
+    method public default void warn(@NonNull String message, @Nullable com.google.devtools.ksp.symbol.KSNode symbol, @NonNull com.google.devtools.ksp.processing.KSPSuggestedFix fix);
+  }
+
+  public final class KSPSuggestedFix {
+    ctor public KSPSuggestedFix(@NonNull String replacementText, @Nullable String description);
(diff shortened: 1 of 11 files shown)


jaschdoc:
Hi thank you for the PR! Your contribution is much appreciated, however, supporting suggested fixes is not a top priority at the moment and I'm not sure this design is quite what we need even though there are some good ideas here. I'm going to close this PR, but feel free to propose a design in the linked issue that we can iterate on.

PR #2969, opened 2026-06-04
Add backing fields to KSP API

### LINKED ISSUE #2873: Add support for explicit backing fields

jaschdoc:
- [x] Add test for type of backing field, e.g., should be `MutableList<String>`

Closes #2873


jaschdoc:
I have ignored the integration tests that include Room. I'm working on submitting a fix for Room, but the problem is circular, since Room depends on KSP, and our tests depend on Room. Thus, the tests should be disabled until Room has the proper fix and then we can update our tests to use the fixed version of Room and re-enable the tests.

PR #2975, opened 2026-06-10
Log resolution trace and lookups

jaschdoc:
This PR adds a new log file to the logs for incremental builds including which symbols were looked up by which files. Additionally, the lookup graph is now both written as a log and as a graphviz file.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/IncrementalContextBase.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/IncrementalContextBase.kt
@@ -28,6 +28,7 @@ import com.intellij.psi.PsiImportStatement
 import com.intellij.psi.PsiJavaFile
 import com.intellij.psi.PsiPackage
 import com.intellij.util.containers.MultiMap
+import java.io.Closeable
 import java.io.File
 import java.util.Date
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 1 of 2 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/IncrementalContextBase.kt
```diff
@@ -531,6 +529,24 @@ abstract class IncrementalContextBase(
...
+
+    protected fun mkFileInLogDir(fileName: String): File =
+        File(logsDir, fileName)
+            .also {
+                if (it.exists()) {
+                    it.delete()
+                    it.createNewFile()
+                } else {
+                    it.createNewFile()
+                }
```
troelsbjerre: `File.delete()` just returns false when the file doesn't exist, so the if-else can be replaced by the first branch.

troelsbjerre:
All this logging is based on naked calls to `File.appendText`. Wouldn't it be preferable to buffer these writes, and perhaps wrap the logger destination in a PrintStream?

troelsbjerre (changes requested):
All this logging is based on naked calls to File.appendText. Wouldn't it be preferable to buffer these writes, and perhaps wrap the logger destination in a PrintStream?

jaschdoc:
Thanks for the review. I have updated the PR.

troelsbjerre (approved):

PR #2981, opened 2026-06-12
fix: sanitize ':' in internal-name module suffix so KSP 2.3.9 works with Kotlin 2.4.0 default module names

### LINKED ISSUE #2964: KSP 2.3.9 with Kotlin 2.4.0 fails to generate code
This can be triggered while using Dagger/Hilt, on 

```kotlin
// my-project/features/Foo/impl/src/kotlin/.../FooModule.kt
@InstallIn(SingletonComponent::class)
internal object FooModule {
    @Provides
    internal fun providesMyDependency(...) = ...
}
```

Here is the corresponding stacktrace it will generate.

```
e: [ksp] java.lang.IllegalArgumentException: Can't escape identifier `FooModule_ProvidesMyDependency$my_project_feature_Foo:implFactory` because it contains illegal characters: :
```

The error goes away when I remove the `internal` modifier on `providesMyDependency`.

This appears to be linked with this recent change in Kotlin 2.4.0:  
> https://kotlinlang.org/docs/whatsnew24.html#consistent-module-names-across-platforms
> Prior to Kotlin 2.4.0, default module names differed across platforms. This inconsistency could cause naming conflicts and resolution issues. Kotlin 2.4.0 standardizes the default names to `{group}:{project_name}` across all platforms.

Manually overriding the `moduleName` fixes the issue, but I suppose KSP (or dagger, I'm not sure whose responsibility it is) should handle these invalid chars that are now standardized by Kotlin 2.4.0 before using them through kotlinpoet.

mvanhorn:
## Summary

KSP's `internal`-member name mangling now sanitizes the module-name suffix the same way the Kotlin compiler does, instead of replacing only `.` and `-`. Kotlin 2.4.0's consistent `${project.group}:${project.name}` default module names inject a literal `:` into the mangled JVM name, and downstream processors that build generated class names from KSP's reported JVM name fail with `Can't escape identifier ... because it contains illegal characters: :` (the Dagger/Hilt report in #2964).

## Why this matters

KSP's reported JVM names need to agree byte-for-byte with what kotlinc actually emits. The compiler already solved the analogous problem for `.kotlin_module` files (KT-69701) with `NameUtils.sanitizeAsJavaIdentifier`; `toSuffix` in `kotlin-analysis-api/.../util.kt` now applies the same sanitization, covering both the `KaSourceModule` and `KaLibraryModule` branches. The workarounds in the issue thread (drop `internal`, override `moduleName`) stop being necessary.

## Testing

`mangledNames.kt` test data now covers a module name carrying `:` and other non-identifier characters. The colon-bearing compiler module name is passed through a test-harness hook (`AbstractKSPAATest`) rather than the test module directory name, so the Windows CI leg, where `:` is illegal in path components, is unaffected.

Fixes #2964

### DIFF
--- a/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AbstractKSPAATest.kt
+++ b/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AbstractKSPAATest.kt
@@ -43,6 +43,11 @@ import java.io.PrintStream
 import java.net.URLClassLoader
 
 abstract class AbstractKSPAATest(val experimentalPsiResolution: Boolean) : AbstractKSPTest(FrontendKinds.FIR) {
+    private companion object {
+        const val MODULE = "// MODULE:"
+        const val COMPILER_MODULE_NAME = "// COMPILER MODULE NAME:"
+    }
+
     val TestModule.kotlinSrc
         get() = File(testRoot, "kotlinSrc")
 
@@ -54,6 +59,26 @@ abstract class AbstractKSPAATest(val experimentalPsiResolution: Boolean) : Abstr
                 it.writeText(file.originalContent)
             }
         }
+    }
+
+    private fun TestModule.compilerModuleName(): String {
+        val originalFile = files.firstOrNull()?.originalFile ?: return name
+        var currentModule: String? = null
+        var compilerModuleName: String? = null
+        originalFile.forEachLine { line ->
+            when {
+                line.startsWith(MODULE) -> {
+                    currentModule = line.substringAfter(MODULE).substringBefore("(").trim()
+                }
+
+                currentModule == name && line.startsWith(COMPILER_MODULE_NAME) -> {
+                    if (compilerModuleName == null) {
+                        compilerModuleName = line.substringAfter(COMPILER_MODULE_NAME).trim()
+                    }
+                }
+            }
+        }
+        return compilerModuleName ?: name
     }
 
     private fun compileKotlin(
@@ -101,7 +126,14 @@ abstract class AbstractKSPAATest(val experimentalPsiResolution: Boolean) : Abstr
         val jvmTarget = testServices.compilerConfigurationProvider
             .getCompilerConfiguration(module, CompilationStage.FIRST)
             .get(JVMConfigurationKeys.JVM_TARGET)
-        compileKotlin(dependencies, module.kotlinSrc.path, module.javaDir.path, module.outDir, module.name, jvmTarget)
+        compileKotlin(
+            dependencies,
+            module.kotlinSrc.path,
+            module.javaDir.path,
+            module.outDir,
+            module.compilerModuleName(),
+            jvmTarget
+        )
         val classpath = (dependencies + KtTestUtil.getAnnotationsJar() + module.outDir)
             .joinToString(File.pathSeparator) { it.absolutePath }
         val options = listOf(
@@ -133,7 +165,7 @@ abstract class AbstractKSPAATest(val experimentalPsiResolution: Boolean) : Abstr
         val testRoot = mainModule.testRoot
 
         val kspConfig = KSPJvmConfig.Builder().apply {
-            moduleName = mainModule.name
+            moduleName = mainModule.compilerModuleName()
             sourceRoots = listOf(mainModule.kotlinSrc)
             javaSourceRoots = compilerConfiguration.javaSourceRoots.map { File(it) }.toList()
             jdkHome = compilerConfiguration.get(JVMConfigurationKeys.JDK_HOME)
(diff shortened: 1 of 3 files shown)


jaschdoc:
Hi thank you for the PR! I will take a look and get back to you :)

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AbstractKSPAATest.kt
```diff
@@ -55,6 +60,26 @@ abstract class AbstractKSPAATest(val experimentalPsiResolution: Boolean) : Abstr
         }
     }
 
+    private fun TestModule.compilerModuleName(): String {
```
jaschdoc, line 64: Please add a doc comment clarifying the purpose of the function and the reason behind it :)

jaschdoc (commented):
Thank you for submitting a fix for this! Just left one minor comment so we know why the extension function is there in 6 months.

jaschdoc:
Oh and one more thing: Could you also add the `@Bug` annotation to the test function in `KSPUnitTestSuite` linking it to https://github.com/google/ksp/issues/2964 so we know that it tests for that issue? See snippet: https://github.com/google/ksp/blob/543112bcd864c5237ac10d38c366c4f9a375e5c1/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt#L536

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AbstractKSPAATest.kt
SimonMarquis: What about this? (disclaimer, I generated it with AI)

```suggestion
   /**
    * Returns the compiler module name for this [TestModule].
    *
    * Test module names in `.kt` test files may contain characters (e.g. `.`, `-`)
    * that are not valid in Kotlin compiler module names. To allow test files to declare
    * a distinct compiler module name via a `// COMPILER MODULE NAME:` directive, this
    * function reads that directive from the first source file.
    *
    * @return the value of the `// COMPILER MODULE NAME:` directive if present, or [name] as fallback
    * @see [issue #2964](https://github.com/google/ksp/issues/2964)
    */
    private fun TestModule.compilerModuleName(): String {
```

SimonMarquis:
> Oh and one more thing: Could you also add the `@Bug` annotation to the test function in `KSPUnitTestSuite` linking it to #2964 so we know that it tests for that issue? See snippet:
> 
> https://github.com/google/ksp/blob/543112bcd864c5237ac10d38c366c4f9a375e5c1/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt#L536

What about this line? 

```kotlin
@Bug("https://github.com/google/ksp/issues/2964", BugState.FIXED)
```

👋 @mvanhorn

mvanhorn:
Rebased onto latest `main`. The conflict was in `AbstractKSPAATest.kt`: `main` added a `jvmTarget` parameter to the `compileKotlin` call while this PR changed the module-name argument from `module.name` to `module.compilerModuleName()`. Combined both - the call now passes `module.compilerModuleName()` and `jvmTarget`.

That also fixed the `:kotlin-analysis-api:ktlint` failure: the merged single-line call exceeded the 120-char limit, so I wrapped it across lines matching the `compileKotlin` signature's existing style. Ran ktlint 0.40.0 locally against the file and it passes clean.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AbstractKSPAATest.kt
```diff
@@ -56,6 +61,26 @@ abstract class AbstractKSPAATest(val experimentalPsiResolution: Boolean) : Abstr
...
+                }
+
+                currentModule == name && line.startsWith(COMPILER_MODULE_NAME) -> {
+                    if (compilerModuleName == null) {
+                        compilerModuleName = line.substringAfter(COMPILER_MODULE_NAME).trim()
+                    }
+                }
+            }
+        }
+        return compilerModuleName ?: name
```
jaschdoc, line 81: IIRC, in a test file you can declare a module that depends on another module defined later in the file, so I think we need to scan over the file twice, i.e., once to get the `MODULE` and once to get the `COMPILER_MODULE_NAME`.

jaschdoc (commented):
Thank you for rebasing. Could you also add the `@Bug` annotation and document the `TestModule.compilerModuleName` extension function? @SimonMarquis has provided some good suggestions for both :)

jaschdoc:
@mvanhorn, do you still want contribute this PR?

jaschdoc (approved):

mvanhorn:
Thank you @jaschdoc for reviewing and landing the module-suffix sanitization.

mvanhorn:
Appreciate it, @jaschdoc! Sanitizing the colon in the internal-name suffix keeps KSP 2.3.9 working.

PR #2983, opened 2026-06-16
Move ksp task cache to a gradle build service

hfmehmed:


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/IsolatedClassLoaderCacheBuildService.kt
```diff
@@ -38,7 +39,7 @@ object IsolatedClassLoaderCache {
 
 abstract class IsolatedClassLoaderCacheBuildService : BuildService<BuildServiceParameters.None>, AutoCloseable {
     companion object {
-        const val KEY = "IsolatedClassLoaderCacheBuildService"
+        val KEY = "IsolatedClassLoaderCacheBuildService_" + UUID.randomUUID().toString()
```
jaschdoc, line 42: Why does this need a random UUID?

hfmehmed: I am using random uuid to prevent ClassCastException in complex builds where the ksp plugin is loaded by multiple different classloaders (e.g bia buildSrc or composite builds). This ensure a unique key is generated per classloader, without naming collisions. AGP uses this practice across all its build services.

jaschdoc (approved):

Issue #2968, opened 2026-06-04
Use Gradle version catalog


jaschdoc:
Dependencies should no longer be declared directly in build files or Gradle property files.

PatriM83:
> Dependencies should no longer be declared directly in build files or Gradle property files.

jaschdoc:
Hey @KlyneChrysler, are you still interested in continuing with this issue?

KlyneChrysler:
> Hey [@KlyneChrysler](https://github.com/KlyneChrysler), are you still interested in continuing with this issue?

will work on it today

KlyneChrysler:
> > Hey [@KlyneChrysler](https://github.com/KlyneChrysler), are you still interested in continuing with this issue?
> 
> will work on it today
got busy from work ^^

jaschdoc:
No worries, there's no pressure! Just wanted to know if you were still interested :)

KlyneChrysler:
This is complete as of #3102. All eight slices are merged: #3026, #3031, #3033, #3040, #3041, #3089, #3095, #3102. Every dependency and plugin version now lives in gradle/libs.versions.toml, and gradle.properties keeps only the two documented values that TestConfig reads at runtime. Thank you @jaschdoc and @hfmehmed for the reviews!

hfmehmed:
Thank you for all the work.

jaschdoc:
Thank you again, @KlyneChrysler. However, as noted in [my comment here](https://github.com/google/ksp/pull/3102#issuecomment-5129922055), this issue is not completed. See for instance: https://github.com/google/ksp/blob/main/kotlin-analysis-api/build.gradle.kts#L117-L226

KlyneChrysler:
You are right, my apologies for the premature completion note. The merged slices covered the versions that lived in gradle.properties, and the inline versions declared directly in build files remain, like the kotlin-analysis-api block you linked. I will continue with the same cadence: small slices moving the remaining inline versions to the catalog, a few dependencies at a time, until none are declared outside it.

PR #2989, opened 2026-06-18
Use a Gradle version catalog for dependency versions #2968

IRus:
1. Moved all versions from gradle.properties into version catalog
2. Updated build files and TestConfig to use version catalog

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -14,10 +14,10 @@ tasks.withType<KotlinCompile> {
 }
 
 plugins {
-    kotlin("jvm")
+    alias(libs.plugins.kotlin.jvm)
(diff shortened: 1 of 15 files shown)


jaschdoc:
Hey @IRus, thank you for the PR! This is really helpful. However, it's a little error-prone to do in one go, so do you think you would be able to split the PR into smaller PRs where we move at most 5 dependencies into the catalog at a time? That way it's much easier to confirm that it's correct.

IRus:
@jaschdoc Sure thing, maybe do migration module-by-module? Starting with less critical

jaschdoc:
Sounds good!

jaschdoc:
I will close this PR for now :)

PR #2995, opened 2026-06-22
Fix R-class resolution in KSP when AGP 9 built-in Kotlin is enabled

hfmehmed:
This change resolves a KSP regression where generated Android R classes fail to resolve when `android.builtInKotlin=true` is enabled. 

The fix is targeted based on whether KAPT is enabled or not.

If KAPT is not applied, KSP now safely wires the kotlinCompile libraries directly. If KAPT (specifically com.android.legacy-kapt) is applied, we fall back to a restricted classpath but wire the compiledRClassArtifact using AGP 9's ComponentCreationConfig API to restore R class visibility. Internal AGP API usage warnings have been suppressed at the file level in AndroidPluginIntegration.kt to satisfy lint checks. 

Fixes #2857

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
@@ -22,6 +22,7 @@ import com.google.devtools.ksp.gradle.utils.allKotlinSourceSetsObservable
 import com.google.devtools.ksp.gradle.utils.canUseGeneratedKotlinApi
 import com.google.devtools.ksp.gradle.utils.enableProjectIsolationCompatibleCodepath
 import com.google.devtools.ksp.gradle.utils.isAgpBuiltInKotlinUsed
+import com.google.devtools.ksp.gradle.utils.isLegacyKaptPluginApplied
 import com.google.devtools.ksp.impl.KotlinSymbolProcessing
 import com.google.devtools.ksp.processing.ExitCode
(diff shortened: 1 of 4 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -233,22 +233,29 @@ abstract class KspAATask @Inject constructor(
...
-                                project.provider {
-                                    val configuration = project.configurations.getByName(
-                                        kotlinCompilation.compileDependencyConfigurationName
-                                    )
-                                    configuration.incoming.artifactView { config ->
-                                        config.attributes.attribute(
-                                            Attribute.of("artifactType", String::class.java), "android-classes-jar"
+                            val androidComponents =
+                                project.extensions.getByType(AndroidComponentsExtension::class.java)
+                            val isKaptApplied = project.plugins.hasPlugin("com.android.legacy-kapt")
```
jaschdoc: This predicate is constructed twice. I think we should investigate if we can move such patterns into a `GradleConfig` object (or something to that effect). Asking for a string property in distinct locations seems a little error-prone.

jaschdoc (dismissed):

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
hfmehmed: Good point. Done

jaschdoc (approved):

PR #2999, opened 2026-06-23
Allow test failures in non-primary tests

jaschdoc:
The `continue-on-error` that we have applied to entire job causes the matrix build to continue if one of the matrix instances fails. It does not apply to the job. This PR fixes that, so if the matrix instance with the secondary tests fails, then the workflow does not fail.

See https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepscontinue-on-error and https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idcontinue-on-error

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,10 +84,12 @@ jobs:
 
     - name: Run secondary integration tests
       if: matrix.test-suite == 'integration-test-secondary'
+      continue-on-error: true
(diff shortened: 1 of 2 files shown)


jaschdoc:
As [this commit](https://github.com/google/ksp/commit/fc7a76457f1049000f5118a6ac8fe2e9c537aa86) shows, the tests are now green even with failures. I'm not sure if we want this behavior, though, as it hides failures in the secondary suites. What do you think @hfmehmed?

hfmehmed:
I agree, we want checks to fail even if something in secondary suite or other suites fail.

PR #3001, opened 2026-06-23
Consume all Kotlin sources in KSP to support other code generators

### LINKED ISSUE #2494: Support AGP kotlin variant sources
AGP has two apis for registering generated code:

```kotlin
variant.sources.java?.addGeneratedSourceDirectory
```

and 

```kotlin
variant.sources.kotlin?.addGeneratedSourceDirectory
```

If we use the latter KSP does not pick up sources generated. 

I had a quick look and I think its relates to this line in particular https://github.com/google/ksp/blob/ab9274e3479182c0045b949a9652031d8e915148/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt#L79

hfmehmed:
This change updates KSP to consume sources.kotlin.all instead of sources.kotlin.static (these are both AGP APIs). This ensures KSP can process Kotlin sources generated by other plugins or tasks in the build that are registered with AGP itself.

Since KSP's own generated outputs (both Java and Kotlin) are registered to sources.java, they do not appear in sources.kotlin.all. This avoids introducing circular dependencies while still allowing KSP to consume other generated Kotlin sources. 

Fixes #2494

### DIFF
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/secondary/AndroidBuiltInKotlinIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/secondary/AndroidBuiltInKotlinIT.kt
@@ -272,4 +272,59 @@ class AndroidBuiltInKotlinIT(experimentalPsiResolution: Boolean) {
             }
         }
     }
+
+    @Test
+    fun `test Kotlin generated sources are picked up by KSP with built-in Kotlin`() {
+        val gradleRunner = GradleRunner.create().withProjectDir(project.root).forwardOutput()
+
+        File(project.root, "workload/build.gradle.kts").appendText(
+            """
+                android {
+                    androidComponents.onVariants { variant ->
+                        val name = variant.name
+                
+                        val task = project.tasks.register("generate${'$'}{name.replaceFirstChar(Char::uppercase)}KotlinSource", KotlinSourceGenerationTask::class) {
+                            getOutputDirectory().set(project.layout.buildDirectory.dir("generated/custom_kotlin/${'$'}{name}"))
+                        }
+                
+                        variant.sources.kotlin?.addGeneratedSourceDirectory(task, KotlinSourceGenerationTask::getOutputDirectory)
+                    }
+                }
+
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/kotlin/com/google/devtools/ksp/test/secondary/AndroidBuiltInKotlinIT.kt
```diff
@@ -272,4 +272,57 @@ class AndroidBuiltInKotlinIT(experimentalPsiResolution: Boolean) {
...
+        File(project.root, "gradle.properties").appendText("\nagpVersion=9.0.0-beta05")
+
+        gradleRunner.withArguments(":workload:assembleDebug", "--stacktrace").build().let { result ->
+            Assert.assertEquals(TaskOutcome.SUCCESS, result.task(":workload:kspDebugKotlin")?.outcome)
+
+            val logFile = File(
+                project.root, "workload/build/generated/ksp/debug/resources/TestProcessor.log"
+            )
+            Assert.assertTrue(
+                "Log file expected to exist but it does not exist: ${'$'}{logFile.path}", logFile.exists()
```
jaschdoc: Is `${'$'}{logFile.path}` intentional? It seems like you want expansion of the path to happen?

jaschdoc (commented):
Before I approve, could you provide some more context as to what the issue is (in addition to the fact that KSP doesn't pick up sources generated by in other ways than KSP)? :)

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/secondary/AndroidBuiltInKotlinIT.kt
hfmehmed: Oops, yes, that looks wrong - updated it

hfmehmed:
> Before I approve, could you provide some more context as to what the issue is (in addition to the fact that KSP doesn't pick up sources generated by in other ways than KSP)? :)

Updated the description of the PR.

cmwarrington (approved):

jaschdoc (approved):

PR #3013, opened 2026-07-01
Record class reference dependencies for incremental compilation

### LINKED ISSUE #3011: KSP incremental compilation does not consider library class references in annotations
Consider the following program:
```kt
// Inside a library / class file
@MyAnno(MyClass::class)
class Annotated
```
If a source file has a reference to `Annotated` and a processor looks it up, then if the processor follows the reference to `MyClass`, then the dependency graph should look like `Source -> Annotated -> MyClass`, but KSP does not record the edge `Annotated -> MyClass`, so if any changes are applied to `MyClass`, then `Annotated` will not be marked dirty and transitively neither will `Source`.
The defect is in the lack of recording the lookup in `KSValueArgument.value`.

jaschdoc:
Adds missing `MyClass::class` references to the dependency graph. If such references occur in class files then it stores a synthetic file handle for it and restores the symbol representing the type computing dirtiness in a subsequent compilation.

Depends on https://github.com/google/ksp/pull/3012

Fixes #3011

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/IncrementalContextBase.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/IncrementalContextBase.kt
@@ -18,6 +18,7 @@
 package com.google.devtools.ksp.common
 
 import com.google.devtools.ksp.IncrementalContextLoggingOptions
+import com.google.devtools.ksp.impl.symbol.kotlin.separateQualifierAndName
 import com.google.devtools.ksp.symbol.KSClassDeclaration
 import com.google.devtools.ksp.symbol.KSDeclaration
 import com.google.devtools.ksp.symbol.KSDeclarationContainer
@@ -253,8 +254,7 @@ abstract class IncrementalContextBase(
 
         // Calculate dirty files by dirty classes in CP.
         val dirtyFilesByCP = changedClasses.flatMap { fqn ->
-            val name = fqn.substringAfterLast('.')
-            val scope = fqn.substringBeforeLast('.', "<anonymous>")
+            val (scope, name) = separateQualifierAndName(fqn)
             classLookupCache[LookupSymbolWrapper(name, scope)].map { it.toRelativeFile() } +
                 symbolLookupCache[LookupSymbolWrapper(name, scope)].map { it.toRelativeFile() }
         }.toSet()
@@ -521,8 +521,7 @@ abstract class IncrementalContextBase(
             return
 
         val path = psiFile.virtualFile.path
-        val name = fqn.substringAfterLast('.')
-        val scope = fqn.substringBeforeLast('.', "<anonymous>")
+        val (scope, name) = separateQualifierAndName(fqn)
 
         // Java types are classes. Therefore lookups only happen in packages.
         fun record(scope: String, name: String) =
@@ -615,8 +614,14 @@ internal class DirtinessPropagator(
         visitedFiles.add(file)
 
         // Propagate by dependencies
-        symbolsMap[file]?.forEach {
-            visit(it)
+        if (NoSourceFile.isSyntheticFile(file.name)) {
+            val fqn = stripSyntheticFileNameModifiers(file.name)
+            val (scope, name) = separateQualifierAndName(fqn)
+            visit(LookupSymbolWrapper(name, scope))
+        } else {
+            symbolsMap[file]?.forEach {
+                visit(it)
+            }
         }
 
         // Propagate by input-output relations
@@ -634,5 +639,18 @@ internal class DirtinessPropagator(
     fun propagate(initialSet: Collection<File>): Set<File> {
         initialSet.forEach { visit(it) }
         return visitedFiles
+    }
+
+    private val noSourceFilePrefix = "<NoSourceFile for "
+    private val noSourceFileSuffix = " is a virtual file; DO NOT USE.>"
+
+    /**
+     * Returns `true` if the file represents is a class literal reference in an annotation argument.
+     */
+    private fun isClassLiteralReferenceInAnnotationArgument(file: File): Boolean {
+        return file != anyChangesWildcard &&
+            file != removedOutputsKey &&
+            file.path.startsWith(noSourceFilePrefix) &&
+            file.path.endsWith(noSourceFileSuffix)
     }
 }
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
@@ -18,12 +18,12 @@
 package com.google.devtools.ksp.impl
 
 import com.google.devtools.ksp.IncrementalContextLoggingOptions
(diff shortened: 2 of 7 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/IncrementalContextBase.kt
```diff
@@ -615,8 +615,15 @@ internal class DirtinessPropagator(
         visitedFiles.add(file)
 
         // Propagate by dependencies
-        symbolsMap[file]?.forEach {
-            visit(it)
+        if (isClassLiteralReferenceInAnnotationArgument(file)) {
```
hfmehmed: is this any different to NoSourceFile.isSyntheticFile ?

jaschdoc: No, I don't think so except the functions were created at different times.

hfmehmed: can we re-use the existing one if it works

jaschdoc: Yes, I will try it out

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/IncrementalContextBase.kt
```diff
@@ -615,8 +615,15 @@ internal class DirtinessPropagator(
...
+            val fqn = file.path.removePrefix(noSourceFilePrefix).removeSuffix(noSourceFileSuffix)
+            val name = fqn.substringAfterLast('.')
+            val scope = fqn.substringBeforeLast('.', "<anonymous>")
```
hfmehmed: nit: these 2 lines already exist somewhere else in this file. could we move to a function? or is it pointless.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/java/KSAnnotationJavaImpl.kt
```diff
@@ -207,6 +218,13 @@ fun calcValue(value: PsiAnnotationMemberValue?): Any? {
                 else -> {
                     ResolverAAImpl.instance
                         .getClassDeclarationByName(component.canonicalText)?.asStarProjectedType()
+                        ?.also { type ->
+                            parent?.let { ctx ->
+                                type.toKaType()?.let { tpe ->
+                                    recordClassReferenceLookup(tpe, ctx)
```
hfmehmed, line 224: is there an exmaple of what "tpe" will be here. im just trying to follow the chain

hfmehmed: if tpe is the FQDN of the type, isnt that equivalent to component.canonicalText ?

jaschdoc: Yeah, it's the type of the annotation reference, e.g., `MyClass` in `@MyAnnotation(MyClass::class)`.

jaschdoc: No, it's the resolved type.

hfmehmed (approved):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/java/KSAnnotationJavaImpl.kt
```diff
@@ -175,8 +181,13 @@ class KSAnnotationJavaImpl private constructor(private val psi: PsiAnnotation, o
     }
 }
 
-fun calcValue(value: PsiAnnotationMemberValue?): Any? {
+fun calcValue(value: PsiAnnotationMemberValue?, parent: KSNode? = null): Any? {
```
bcorso, line 184: Looks like this commit updates all usages of `calcValue` to pass in a non-null parent except for 2 usages in [KSAnnotationImpl.kt](https://github.com/google/ksp/blob/4c2c36d1ea7277e3205ef8cc11e11aa94d685f21/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSAnnotationImpl.kt#L24)

Does it make sense to just update those usages too to pass in the parent and make this input non-null?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/java/KSAnnotationJavaImpl.kt
```diff
@@ -207,6 +218,13 @@ fun calcValue(value: PsiAnnotationMemberValue?): Any? {
                 else -> {
                     ResolverAAImpl.instance
                         .getClassDeclarationByName(component.canonicalText)?.asStarProjectedType()
+                        ?.also { type ->
+                            parent?.let { ctx ->
+                                type.toKaType()?.let { tpe ->
```
bcorso, line 223: `tpe` seems weird for a name. Should we call this `kaType`?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
```diff
@@ -182,6 +185,75 @@ class IncrementalContextAA(
         recordWithArgs(type, file)
     }
 
+    /**
+     * Records a reference to `MyClass::class`.
+     * 
+     * @param type the referenced type, e.g., `MyClass`
+     * @param context the parent of [type], i.e., where the reference occurs.
+     */
+    fun recordClassReferenceLookup(type: KaType, context: KSNode) {
```
bcorso, line 194: Not related to this issue, but would it be worth creating a `IncrementalContextNoop` implementation for when `isIncremental` is false, rather than including this at the top of every function?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/IncrementalContextBase.kt
```diff
@@ -635,4 +640,17 @@ internal class DirtinessPropagator(
         initialSet.forEach { visit(it) }
         return visitedFiles
     }
+
+    private val noSourceFilePrefix = "<NoSourceFile for "
```
bcorso, line 644: This is unused now, right?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/java/KSAnnotationJavaImpl.kt
jaschdoc: Yes, good point. It's just an old habit of mine from Scala.

PR #3014, opened 2026-07-03
Fix configuration cache invalidation in klib cross-compilation check

omarismail94:
KotlinNativeCompilation.crossCompilationSupported resolves the crossCompilationMetadata JSONs exported by project dependencies and probes their existence. Capturing it in onlyIf makes the configuration cache evaluate it at store time, before the exporting tasks have run. When the build subsequently creates these files, the next build discards the cache entry with: "configuration cache cannot be reused because the file system entry ... has been created".

To fix this, we mirror KGP's own compileKotlin<Target> task behavior:

* Keep the target-level condition (cross-compilation property enabled and no cinterops) in a stored provider
* Declare the metadata JSONs as task inputs so that the exporting tasks run first.
* Read and parse these metadata files in onlyIf at execution time.

TESTED: Update the regression test (testNativeConfigurationCacheReuse) which could not catch this bug previously. The old test used a host-supported target that never took the cross-compilation target. The updated test targets iosArm64 to force the cross-compilation path, and builds the iosArm64 klib and asserts configuration cache reuse.

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
@@ -36,8 +36,10 @@ import org.gradle.api.JavaVersion
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.attributes.Attribute
+import org.gradle.api.attributes.Usage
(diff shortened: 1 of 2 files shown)


jaschdoc:
I think this PR requires a review from @hfmehmed :)

hfmehmed (dismissed):

hfmehmed (approved):

PR #3018, opened 2026-07-06
Add logging for dependency graph

### LINKED ISSUE #3015: Visualize dependency graph for incremental compilation
Given an origin symbol, output a `.dot` file that visualizes what other symbols are reachable from that node. Such a graph makes it easier to spot which edges are missing when debugging or error reporting.

jaschdoc:
Adds a new Gradle property: `ksp.incremental.log.graph.origin=<String>` that takes a fully qualified name and if `ksp.incremental.log` is enabled/set to `true`, then it will output a Graphviz (`.dot`) file in the logs directory that contains the dependencies of the fully qualified name.

Closes https://github.com/google/ksp/issues/3015

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
@@ -17,10 +17,14 @@
 
 package com.google.devtools.ksp.impl
 
+import com.google.devtools.ksp.IncrementalContextLoggingOptions
 import com.google.devtools.ksp.common.IncrementalContextBase
 import com.google.devtools.ksp.common.LookupStorageWrapper
 import com.google.devtools.ksp.common.LookupSymbolWrapper
 import com.google.devtools.ksp.common.LookupTrackerWrapper
+import com.google.devtools.ksp.common.NoSourceFile
+import com.google.devtools.ksp.common.isSyntheticFileName
+import com.google.devtools.ksp.common.stripSyntheticFileNameModifiers
 import com.google.devtools.ksp.containingFile
 import com.google.devtools.ksp.impl.symbol.kotlin.KSFileJavaImpl
 import com.google.devtools.ksp.impl.symbol.kotlin.KSFunctionDeclarationImpl
@@ -66,7 +70,7 @@ class IncrementalContextAA(
     override val isIncremental: Boolean,
     lookupTracker: LookupTracker,
     anyChangesWildcard: File,
-    incrementalLog: Boolean,
+    loggingOptions: IncrementalContextLoggingOptions,
     baseDir: File,
     cachesDir: File,
     kspOutputDir: File,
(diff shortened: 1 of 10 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
```diff
@@ -274,17 +277,91 @@ class IncrementalContextAA(
...
+                    }
+                }
+            }
+        }
+
+        // BFS
+        val visited = mutableSetOf<String>()
+        val queue = ArrayDeque<String>()
+        val bfsEdges = mutableSetOf<Pair<String, String>>()
+        edges.keys.find { it == origin }?.let { startVertex ->
```
hfmehmed: i think this may be sub optimal O(N) lookup ?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
```diff
@@ -274,17 +277,91 @@ class IncrementalContextAA(
...
-        if (!incrementalLog) {
+    /**
+     * Computes the dependency graph from [origin] using a normal BFS and logs it as a Graphviz `.dot` file.
+     */
+    private fun logDependencyGraphFrom(origin: String?, lookupRecords: Map<String, List<String>>) {
+        if (!loggingOptions.incrementalLoggingEnabled || origin == null) {
+            return
+        }
+
+        if (origin.isBlank() || origin.isEmpty()) {
```
hfmehmed: isBlank is enough here

jaschdoc: Yeah good point.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
```diff
@@ -274,17 +277,91 @@ class IncrementalContextAA(
...
+        updatedSymbols.keySet().forEach { file ->
+            allDefinedSyms.getOrPut(file, ::mutableSetOf).addAll(updatedSymbols[file])
+        }
+
+        // Prepare edges
+        val edges = mutableMapOf<String, MutableSet<String>>()
+        lookupRecords.forEach { (fqn, paths) ->
+            paths.forEach { path ->
+                if (isSyntheticFileName(path)) {
+                    val fqnSource = stripSyntheticFileNameModifiers(path)
```
hfmehmed, line 316: this will throw exception is the type of the path is AnyChanges?

jaschdoc:
Thank you for the review. Fixed it.

hfmehmed:
Can you confirm that none of this code will be executed if ksp.incremental.log.graph.origin gradle property is not set. Just to ensure there is no side effect to users who dont need this feature

jaschdoc:
> Can you confirm that none of this code will be executed if ksp.incremental.log.graph.origin gradle property is not set. Just to ensure there is no side effect to users who dont need this feature

I will double check and ping you again after.

jaschdoc:
@hfmehmed, I have double checked on examples that the code does not run if disabled and added additional tests.

hfmehmed (approved):

Issue #3008, opened 2026-06-29
`KSValueArgument.value` may return `List<*>`


jaschdoc:
The API specifies that `KSValueArgument.value` returns an `Array` of the above listed types, however, if an annotation is given in a Java file as `@Anno(args = [arg1, arg2, ...])`, then `[arg1, arg2, ...]` is returned as a `java.util.ArrayList<*>`. Thus, if you match on the result of `KSValueArgument.value` and match on any or all of the types declared in the documentation, there is a case where the match is non-exhaustive or rather where you never hit a valid branch.

https://github.com/google/ksp/blob/98e422e6b32ba2f060d4e2ff20f7552321c2b8c2/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSValueArgument.kt#L55

The fix is to check for the `ArrayList<*>` type and convert it to a `kotlin.Array<*>` type.

Yong-yuan-X:
Hi, I’d like to work on this and add a regression test for the `ArrayList<*>` to `Array<*>` conversion

jaschdoc:
Hi @Yong-yuan-X, thank you for the interest. The best way to start with this would be to open a PR with a new test in [`kotlin-analysis-api/testData`](https://github.com/google/ksp/tree/main/kotlin-analysis-api/testData) directory that shows this defect. Then add the test to [`KSPUnitTestSuite`](https://github.com/google/ksp/blob/main/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/KSPUnitTestSuite.kt) and link it to this issue with the [`@Bug` annotation](https://github.com/google/ksp/blob/main/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/annotations/Bug.kt). Note that you can use `runFailingTest` to add a test that gives the wrong output. Once that PR is submitted, we can continue to implement a fix.

jaschdoc:
@danysantiago, this might impact Room, since you already have a fix for it in place https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room3/room3-compiler-processing/src/main/java/androidx/room3/compiler/processing/ksp/KspAnnotationValue.kt;l=164?q=%22KSValueArgument%22

jaschdoc:
Related to https://github.com/google/ksp/issues/214

jaschdoc:
[...]

PR #3023, opened 2026-07-08
Allow collection-backed annotation array values in regression test

Yong-yuan-X:
Refs #3008

## Summary

* Narrow the `annotationArrayValueType` regression test to document the currently supported value shapes for array-valued annotation arguments.
* Update the test to accept annotation array values represented as either `Array<*>` or `Collection<*>`.
* Make the test fail if the value is not one of the supported documented shapes.
* Keep the regression test as a regular passing `runTest`.

## Tests

* Ran `.\gradlew.bat :kotlin-analysis-api:test --tests "com.google.devtools.ksp.test.AAConfiguredUnitTestSuite.testAnnotationArrayValueType" --tests "com.google.devtools.ksp.test.PsiConfiguredUnitTestSuite.testAnnotationArrayValueType"`
* `AAConfiguredUnitTestSuite > testAnnotationArrayValueType() PASSED`
* `PsiConfiguredUnitTestSuite > testAnnotationArrayValueType() PASSED`
* `BUILD SUCCESSFUL`

### DIFF
--- a/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AAConfiguredUnitTestSuite.kt
+++ b/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AAConfiguredUnitTestSuite.kt
@@ -31,7 +31,7 @@ class AAConfiguredUnitTestSuite : KSPUnitTestSuite(experimentalPsiResolution = f
     @TestMetadata("annotationArrayValueType.kt")
     @Test
     override fun testAnnotationArrayValueType() {
-        runFailingTest("$AA_PATH/annotationArrayValueType.kt")
+        runTest("$AA_PATH/annotationArrayValueType.kt")
     }
 
     @TestMetadata("allUseSiteTargetAppliedToAnnotationList.kt")
--- a/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AAConfiguredUnitTestSuite.kt
+++ b/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AAConfiguredUnitTestSuite.kt
@@ -28,12 +28,6 @@ class AAConfiguredUnitTestSuite : KSPUnitTestSuite(experimentalPsiResolution = f
         runTest("$AA_PATH/getSymbolsWithAnnotation/aliasedAnnotation.kt")
     }
 
-    @TestMetadata("annotationArrayValueType.kt")
-    @Test
-    override fun testAnnotationArrayValueType() {
-        runTest("$AA_PATH/annotationArrayValueType.kt")
-    }
-
     @TestMetadata("allUseSiteTargetAppliedToAnnotationList.kt")
     @Test
     override fun testAllUseSiteTargetAppliedToAnnotationList() {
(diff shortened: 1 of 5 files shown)


jaschdoc:
Hey @Yong-yuan-X, thank you for another PR. We may have to rethink this a bit. It occurred to me that it's possibly better to treat this defect as a documentation bug. So instead of fixing all occurrences of `List<T>`, it's better to document that the value can be a list. The real design problem here is that the `KSValueArgument.value` property has type `Any?`. Instead, we should have a sum type, e.g., `KSValueArgumentElement`, that describes the possible values of `KSValueArgument.value`. However, to avoid breaking changes, we should add this type outside of the API and introduce an extension function or property, `KSValueArgument.typedValue : KSValueArgumentElement` that users can use in the future.

Thus, I think this PR should update the test you made earlier to allow for the `Collection<T>` / `List<T>` case and throw errors if it's not one of the documented types. That change probably also requires changing the expected results, but that's okay. At least we can change it to a passing test.

In the meantime, I will write up a better design in #3008 and if you still want to implement it we can discuss further there.

Again, thank you for the contribution and apologies for changing the goal midway through the work. I hope you understand! :)

Yong-yuan-X:
@jaschdoc Thank you for the clarification, and no worries about the change in direction. I appreciate the explanation and will adjust the PR accordingly.

I understand the concern that this is more about documenting the currently supported value shapes for `KSValueArgument.value`, rather than forcing annotation array values to be represented as `Array<*>`.

I'll update this PR to narrow the scope:
* revert the production changes that convert list-backed annotation values to arrays;
* update the `annotationArrayValueType` regression test to allow `Collection`/`List` for array-valued annotation arguments;
* keep the test as a regular passing `runTest`;
* make it fail only when the value is not one of the documented supported types.

I'll also follow the design update in #3008, and I'm happy to help with the future `typedValue` direction once it is settled.

jaschdoc (approved):
Thank you for the PR! Feel free to open another one that fixes the documentation. Cheers :)

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/AAConfiguredUnitTestSuite.kt
```diff
@@ -31,7 +31,7 @@ class AAConfiguredUnitTestSuite : KSPUnitTestSuite(experimentalPsiResolution = f
     @TestMetadata("annotationArrayValueType.kt")
     @Test
     override fun testAnnotationArrayValueType() {
```
jaschdoc: Could you move this function into `KSPUnitTestSuite`, i.e., remove the `abstract` modifier from `testAnnotationArrayValueType` in `KSPUnitTestSuite`. Thanks! :)

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/test/PsiConfiguredUnitTestSuite.kt
```diff
@@ -31,7 +31,7 @@ class PsiConfiguredUnitTestSuite : KSPUnitTestSuite(experimentalPsiResolution =
     @TestMetadata("annotationArrayValueType.kt")
     @Test
     override fun testAnnotationArrayValueType() {
```
jaschdoc: Same as the other comment.

jaschdoc (changes requested):
Ah apologies for changing my mind. Just a nitpick that I wanted to catch first.

Yong-yuan-X:
I moved `testAnnotationArrayValueType` into `KSPUnitTestSuite` and removed the duplicate overrides from the AA and PSI suites. Both targeted tests still pass. Thanks again for the detailed reviews and guidance on both PRs — I really appreciate it！

jaschdoc (approved):
Great, thank you for the changes.

PR #3025, opened 2026-07-08
Use Gradle version catalog for dependency declarations

### LINKED ISSUE #2968: Use Gradle version catalog

KlyneChrysler:
Fixes #2968

Moves dependency and plugin versions into `gradle/libs.versions.toml` and replaces interpolated dependency strings with catalog accessors across the module build files (`api` and other modules without version properties are untouched).

## Scope decisions

- **Test-project fixtures under `integration-tests/src/test/resources/` and `examples/` are intentionally NOT migrated** — they are spawned as separate Gradle builds by the test harness and receive versions via `-P` properties, which a root catalog cannot reach.
- **`kotlinBaseVersion` and `agpBaseVersion` remain in `gradle.properties`** (with a comment explaining why): `gradle-plugin`'s `TestConfig` loads the root `gradle.properties` file directly at test runtime. Build files no longer read them. A follow-up could route these through the existing `prepareTestConfiguration` task to eliminate the duplication — happy to do that here if preferred.
- In `kotlin-analysis-api`, `aaKotlinBaseVersion`/`aaIntellijVersion` stay as local `val`s sourced from `libs.versions.*` since they version loop-generated coordinates, `kotlin(...)` helper calls, and jar-rename regexes that catalog accessors can't express.
- CI cache keys in `main.yml`/`auto-merge.yml` now also hash `**/libs.versions.toml`.

## Verification

- `./gradlew compileKotlin compileTestKotlin`: BUILD SUCCESSFUL across all modules.
- `./gradlew :gradle-plugin:test`: 41 pass; the 12 failing tests are the Android TestKit tests, which fail identically on a clean checkout of main on this machine (no Android SDK) — unrelated to this change.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -78,7 +78,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ~/.gradle/caches
-          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle.kts') }}-${{ hashFiles('**/gradle.properties') }}
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle.kts') }}-${{ hashFiles('**/gradle.properties') }}-${{ hashFiles('**/libs.versions.toml') }}
           # An ordered list of keys to use for restoring the cache if no cache hit occurred for key
           restore-keys: |
             ${{ runner.os }}-gradle-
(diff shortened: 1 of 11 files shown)


jaschdoc:
Hi @KlyneChrysler, thank you for the PR! As mentioned in https://github.com/google/ksp/pull/2989, it'd be better if we split this into smaller PRs and only move one or two dependencies at a time. Thus, I will close this PR for now, but feel free to break this into smaller PRs that can be reviewed separately. Additionally, you will have to sign the CLA.

PR #3026, opened 2026-07-08
Move common-util and common-deps dependencies to a version catalog

KlyneChrysler:
First slice of #2968, following the guidance in #3025 and #2989 to migrate module by module with only a couple of dependencies at a time.

This introduces gradle/libs.versions.toml with kotlin-stdlib and junit, and switches common-util and common-deps to catalog accessors. Nothing is removed from gradle.properties yet because other modules still read those properties.

Verified locally with ./gradlew :common-util:compileTestKotlin :common-deps:compileTestKotlin.

The remaining modules are prepared as equally small follow-up slices and I will submit them one at a time as each PR lands, so every review stays small.

### DIFF
--- a/common-deps/build.gradle.kts
+++ b/common-deps/build.gradle.kts
@@ -2,7 +2,6 @@ import org.jetbrains.kotlin.gradle.dsl.JvmDefaultMode
 
 description = "Kotlin Symbol Processor"
 
-val junitVersion: String by project
 val signingKey: String? by project
 val signingPassword: String? by project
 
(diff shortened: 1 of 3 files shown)


jaschdoc (commented):
Thank you for the PR! @hfmehmed, could you double check this to make sure everything works exactly as before?

KlyneChrysler:
To help with the double check: I compared the resolved dependency graphs between this commit and its base commit by running ./gradlew :common-util:dependencies and :common-deps:dependencies for the compileClasspath, runtimeClasspath, testCompileClasspath, and testRuntimeClasspath configurations. The output is byte for byte identical, so resolution is unchanged. Both modules' test suites also pass on this branch, 20 tests, 0 failures.

hfmehmed (approved):
looks good to me. thanks for doing this

KlyneChrysler:
All checks passed, thank you for reviewing. @hfmehmed @jaschdoc

PR #3031, opened 2026-07-10
Move integration-tests dependencies to the version catalog

KlyneChrysler:
Second slice of #2968, now that #3026 has landed. Thank you @hfmehmed and @jaschdoc for the quick review and merge.

The integration-tests module's junit, kotlinx-serialization, and coroutines dependencies move to catalog accessors, and the kotlinVersion and agpVersion test system properties now read from the catalog. Their gradle.properties entries stay because other modules still read them.

Verified that resolution is unchanged: ./gradlew :integration-tests:dependencies output for the compile, runtime, testCompile, and testRuntime classpaths is identical before and after this commit. :integration-tests:compileTestKotlin passes.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,12 @@
 [versions]
 kotlin-base = "2.3.20"
+kotlinx-serialization = "1.6.3"
+agp-base = "9.3.0-alpha01"
 junit = "4.13.1"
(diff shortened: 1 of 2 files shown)


hfmehmed (approved):

KlyneChrysler:
All checks passed, thank you for reviewing.

PR #3033, opened 2026-07-10
Move gradle-plugin dependencies to the version catalog

KlyneChrysler:
Third slice of #2968, following #3026 and #3031.

The Kotlin Gradle plugin artifacts, kotlin-compiler-embeddable, AGP, and Truth move to catalog accessors, and the generateKSPVersions task reads the coroutines version from the catalog. googleTruthVersion is removed from gradle.properties since gradle-plugin was its last consumer. kotlinBaseVersion stays because TestConfig loads gradle.properties directly at test runtime.

Verified that resolution is unchanged: ./gradlew :gradle-plugin:dependencies output for the compile, runtime, testCompile, and testRuntime classpaths is identical before and after this commit. :gradle-plugin:compileTestKotlin passes.

### DIFF
--- a/gradle-plugin/build.gradle.kts
+++ b/gradle-plugin/build.gradle.kts
@@ -16,8 +16,8 @@ plugins {
 }
 
 dependencies {
-    compileOnly(libs.kotlin.gradle.plugin.api)
-    compileOnly(libs.kotlin.gradle.plugin.asProvider())
+    compileOnly(libs.kotlin.gradlePluginApi)
+    compileOnly(libs.kotlin.gradlePlugin)
     compileOnly(libs.kotlin.compiler.embeddable)
     // replace AGP dependency w/ gradle-api when we have source registering API available.
     compileOnly(libs.agp)
--- a/gradle-plugin/build.gradle.kts
+++ b/gradle-plugin/build.gradle.kts
@@ -3,11 +3,6 @@ import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 
 description = "Kotlin Symbol Processor"
 
-val kotlinBaseVersion: String by project
-val junitVersion: String by project
-val googleTruthVersion: String by project
-val agpBaseVersion: String by project
-val aaCoroutinesVersion: String? by project
 val signingKey: String? by project
 val signingPassword: String? by project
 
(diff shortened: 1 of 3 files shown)


gradle-plugin/build.gradle.kts
```diff
@@ -21,21 +16,21 @@ plugins {
 }
 
 dependencies {
-    compileOnly("org.jetbrains.kotlin:kotlin-gradle-plugin-api:$kotlinBaseVersion")
-    compileOnly("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinBaseVersion")
-    compileOnly("org.jetbrains.kotlin:kotlin-compiler-embeddable:$kotlinBaseVersion")
+    compileOnly(libs.kotlin.gradle.plugin.api)
+    compileOnly(libs.kotlin.gradle.plugin.asProvider())
```
jaschdoc: Could you explain why this is required `.asProvider()`? :)

jaschdoc (commented):
Thank you for another contribution. I just have one question before proceeding.

gradle-plugin/build.gradle.kts
KlyneChrysler: The catalog has both kotlin-gradle-plugin and kotlin-gradle-plugin-api, so kotlin-gradle-plugin is simultaneously a leaf alias and a prefix of another alias. 

In that case Gradle's generated type safe accessor libs.kotlin.gradle.plugin is an accessor group rather than a dependency provider, and the leaf has to be selected explicitly with .asProvider(). Without it the build script does not compile. 

If you prefer cleaner call sites I am happy to rename the aliases instead, for example kotlinGradlePlugin and kotlinGradlePluginApi, which avoids the prefix collision entirely.

jaschdoc: Sounds good, thank you for the explanation. Please rename such that the use site is `libs.kotlin.gradlePluginApi` and `libs.kotlin.gradlePlugin`.

jaschdoc (changes requested):
Just a minor renaming around the gradle plugin api. Looks good otherwise :)

gradle-plugin/build.gradle.kts
KlyneChrysler: Done, renamed to libs.kotlin.gradlePlugin and libs.kotlin.gradlePluginApi. The asProvider call is gone since the aliases no longer share a prefix. Verified :gradle-plugin:compileTestKotlin still passes and resolution is unchanged.

hfmehmed: Did you push the changes?

KlyneChrysler: Apologies, the push had silently failed on my side, thank you for catching that.

jaschdoc (approved):
Thank you for the changes. Looks good :)

PR #3037, opened 2026-07-15
Fix KSP crash for missing parent declaration in incremental compilation

### LINKED ISSUE #3030: Unexpected missing parent declaration for KSNode '@NotNull' with KSP 2.3.10
Micronaut project build worked with 2.3.9 but failed with 2.3.10:

e: [ksp] Originating element: options
e: [ksp] >>> Internal KSP Error
[ksp] Originating element: options

   | *** THIS IS A BUG IN KSP ***
   |
   | Unexpected missing parent declaration for KSNode '@NotNull'
   |
   | Location           : <unknown location>
   | Class at occurrence: class com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSAnnotationResolvedImpl
   |
   | You can report it at https://github.com/google/ksp/issues/new
   |

It is a large project so I was not able to make a reproducible case.

stacktrace:
[scratch_41.txt](https://github.com/user-attachments/files/29870263/scratch_41.txt)

jaschdoc:
This PR removes the internal KSP exception that can happen when recording class reference dependencies in incremental compilation, and moves the responsibility of obtain the parent declaration to call sites. That way, a dependency is safely recorded if there is a parent declaration available (the crash could happen for `KSTypeImpl` which does not have a parent declaration for its annotations, since `KSType` is not a `KSNode` or a `KSDeclaration`).

Depends on https://github.com/google/ksp/pull/3034
Fixes https://github.com/google/ksp/issues/3030

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/IncrementalContextAA.kt
@@ -18,7 +18,6 @@
 package com.google.devtools.ksp.impl
 
 import com.google.devtools.ksp.IncrementalContextLoggingOptions
-import com.google.devtools.ksp.InternalKSPException
 import com.google.devtools.ksp.common.IncrementalContextBase
 import com.google.devtools.ksp.common.LookupStorageWrapper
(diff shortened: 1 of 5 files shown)


jaschdoc:
@troelsbjerre, I'm not sure if you have seen this PR or not, so this is just a friendly ping! :)

troelsbjerre (approved):

PR #3040, opened 2026-07-16
Move symbol-processing-aa-embeddable dependencies to the version catalog

KlyneChrysler:
Fourth slice of #2968, following #3026, #3031, and #3033.

The packed coroutines jar moves to a catalog accessor and the aa Kotlin base version becomes a catalog sourced value used by the packed jar renaming and the generated version metadata. aaKotlinBaseVersion and aaCoroutinesVersion stay in gradle.properties while kotlin-analysis-api still reads them.

Verified that resolution is unchanged: ./gradlew :symbol-processing-aa-embeddable:dependencies output for the compile, runtime, and packedJars configurations is identical before and after this commit. :symbol-processing-aa-embeddable:compileKotlin passes.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,6 +4,7 @@ kotlinx-serialization = "1.6.3"
 agp-base = "9.3.0-alpha01"
 junit = "4.13.1"
 google-truth = "1.4.5"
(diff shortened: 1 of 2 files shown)


jaschdoc (approved):
Thanks for another PR. Looks good! :)

PR #3041, opened 2026-07-16
Move kotlin-analysis-api test dependencies to the version catalog

KlyneChrysler:
Fifth slice of #2968 and the first of three for kotlin-analysis-api, following #3026, #3031, #3033, and #3040.

The JUnit 5 and JUnit Platform test artifacts move to catalog accessors. junitVersion, junit5Version, and junitPlatformVersion are removed from gradle.properties since this module was their last reader, which I verified by grepping every build file including the integration test fixtures.

Verified that resolution is unchanged: ./gradlew :kotlin-analysis-api:dependencies output for the testCompile and testRuntime classpaths is identical before and after this commit, 254 lines compared. :kotlin-analysis-api:compileTestKotlin passes.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,10 +5,6 @@ org.gradle.parallel=true
 kotlinBaseVersion=2.3.20
 kotlinxSerializationVersion=1.6.3
 agpBaseVersion=9.3.0-alpha01
-junitVersion=4.13.1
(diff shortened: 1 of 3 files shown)


jaschdoc (approved):
Thank you again, looks good :)

PR #3045, opened 2026-07-17
Experiment with allowing global ksp config for single platforms

hfmehmed:


jaschdoc:
We are updating the documentation to reflect that one should use `kspTest` for test sources. See https://github.com/JetBrains/kotlin-web-site/pull/5687. Should we still support this? I think having explicit source configurations is clearer and also gives a clear mental model for KMP. What do you think?

jaschdoc (commented):
See https://github.com/google/ksp/pull/3045#issuecomment-5023481897

jaschdoc:
@hfmehmed where did we land on this?

Issue #2498, opened 2025-06-19
[KSP2] Incorrect parentDeclaration for Java method inheriting a Kotlin property


bacecek:
Reproducer:
```
// FILE: K.kt
interface MyInterface {
    val property: Int
}

// FILE: MyInterfaceImplJava.java
class MyInterfaceImplJava implements MyInterface {
    @Override
    public int getProperty() { return 0; }
}
```

`KSDeclaration.parent` and `KSDeclaration.parentDeclaration` returns `property` instead of `MyInterfaceImplJava`. KSP1 works correctly.

I've got it in #2497.

bacecek:
Issue is still actual

PR #3048, opened 2026-07-17
Reproduce incorrect parent of Java method overriding a Kotlin property

bacecek:
Reproduces #2498

I kept it in separate file instead of `kotlin-analysis-api/testData/parent.kt` to isolate the problem and use `runFailingTest` only for that exact problem so that all other cases in `kotlin-analysis-api/testData/parent.kt` will keep running normally.

### DIFF
--- a/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/processor/ParentOfJavaMethodOverridingKotlinPropertyProcessor.kt
+++ b/kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/processor/ParentOfJavaMethodOverridingKotlinPropertyProcessor.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2026 Google LLC
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/processor/ParentOfJavaMethodOverridingKotlinPropertyProcessor.kt
jaschdoc, line 1: Missing copyright header :)

jaschdoc, line 1: Please also move this file to `kotlin-analysis-api/src/test/.../processor`.

jaschdoc (commented):
Hi @bacecek, thank you for the PR! Just two minor comments, but it looks good otherwise.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/processor/ParentOfJavaMethodOverridingKotlinPropertyProcessor.kt
bacecek: Thanks, fixed

bacecek: Thanks, fixed (I guess)

jaschdoc (approved):
Thanks for the changes @bacecek!

PR #3063, opened 2026-07-21
Fix build cache misses by marking KspGradleConfig.logLevel as @Internal

### LINKED ISSUE #3039: logLevel is tracked as a task @Input, causing needless cache misses when the Gradle log level changes
The KSP Gradle task exposes KspGradleConfig.logLevel (Property<LogLevel>), and it's annotated @org.gradle.api.tasks.Input. Since the value is the current Gradle log level, running the same build at a different log level changes the task's cache key and re-runs all KSP tasks:

> Task :some-module:kspKotlinJs
  Value of input property 'kspConfig.logLevel' has changed for task ':some-module:kspKotlinJs'

This bites in practice whenever the log level isn't constant - e.g. a plain build vs. -q/--warn, or a CI pipeline that retries with --info. Every such build throws away otherwise-valid KSP cache entries.

Versions: KSP 2.3.9, Gradle 9.6.1.

hfmehmed:
Fixes #3039


jaschdoc (approved):
Thanks! :)

PR #3064, opened 2026-07-21
Update `KSVisitor` deprecation message

jaschdoc:
This commit changes wording from "next KSP version" to "an upcoming KSP version" and provides a bit more context as to why the interface is deprecated this way.
Related to #2873


jaschdoc:
Also a ping on this one @troelsbjerre :)

troelsbjerre (approved):

PR #3066, opened 2026-07-22
Update error message for ksp.useKSP2 property

jaschdoc:
Adds a sentence noting that the property will be removed in the future so users know they shouldn't rely on it.

What do you think about this @hfmehmed?

Related to https://github.com/google/ksp/issues/3067

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -145,7 +145,8 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
         if (useKsp2.not()) {
             throw RuntimeException(
                 "KSP1 is no longer available. Please use KSP2 instead and do not explicitly set ksp.useKsp2 to false " +
-                    "via the DSL or the Gradle property please"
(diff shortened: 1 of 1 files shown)


hfmehmed:
sounds good to me. lets also update README.md to stop mentioning this property as well as anything in the docs/ folder that might reference it. 

we should just assume KSP1 does not exist in our docs from now on

jaschdoc:
Sounds good. I will create another PR.

jaschdoc:
Could I ask you to also submit a review? Thanks! :)

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
```diff
@@ -145,7 +145,8 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
         if (useKsp2.not()) {
             throw RuntimeException(
                 "KSP1 is no longer available. Please use KSP2 instead and do not explicitly set ksp.useKsp2 to false " +
-                    "via the DSL or the Gradle property please"
+                    "via the DSL or the Gradle property please. " +
```
hfmehmed, line 148: can you remove the last "please" too in your next PR. it seems like a repeat since the sentence already starts with please

hfmehmed (approved):

PR #3068, opened 2026-07-22
Remove `use.KSP2` from documentation

### LINKED ISSUE #3067: Remove documentation for `use.KSP2`
KSP1 no longer exists and there is no way to run in the current version. Setting the `use.KSP2` property to `false` results in an error, and the documentation should thus avoid mentioning the property.

jaschdoc:
Closes https://github.com/google/ksp/issues/3067

### DIFF
--- a/docs/ksp2.md
+++ b/docs/ksp2.md
@@ -1,4 +1,4 @@
-# Introduction to KSP2 (Beta)
+# Introduction to KSP2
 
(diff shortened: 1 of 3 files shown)


docs/ksp2.md
jaschdoc, line 1: I have not removed this file because it seems okay to have. I have, however, tried to de-emphasize the KSP1 parts.

troelsbjerre (approved):

PR #3069, opened 2026-07-22
Document collection values in KSValueArgument.value

Yong-yuan-X:
## Summary

* Document that `KSValueArgument.value` may be represented as a `Collection` in addition to an `Array`.
* Align the API documentation with the behavior covered by #3023.

## Testing

Not run (documentation-only change).

Refs #3008

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSValueArgument.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSValueArgument.kt
@@ -50,7 +50,7 @@ interface KSValueArgument : KSAnnotated {
      * * [KSClassDeclaration] for annotation arguments of type [Enum] (in this
      *   case[KSClassDeclaration.classKind] equals to [ClassKind.ENUM_CLASS]);
      * * [KSAnnotation] for embedded annotation arguments;
-     * * [Array] of a possible type listed above.
+     * * [Array] or [List] of a possible type listed above.
      */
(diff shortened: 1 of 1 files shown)


api/src/main/kotlin/com/google/devtools/ksp/symbol/KSValueArgument.kt
```diff
@@ -50,7 +50,7 @@ interface KSValueArgument : KSAnnotated {
      * * [KSClassDeclaration] for annotation arguments of type [Enum] (in this
      *   case[KSClassDeclaration.classKind] equals to [ClassKind.ENUM_CLASS]);
      * * [KSAnnotation] for embedded annotation arguments;
-     * * [Array] of a possible type listed above.
+     * * [Array] or [List] of a possible type listed above.
```
jaschdoc: I think it's safer to document it as `Collection` instead of `List`.

jaschdoc (commented):
Thank you for the PR, @Yong-yuan-X! Just one minor comment.

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSValueArgument.kt
Yong-yuan-X: Done, updated to `Collection`

jaschdoc (approved):
Thank you for the changes! :)

PR #3077, opened 2026-07-24
Fix throwing on asMemberOf on Intersection Override functions

vadymsavchenko-lyft:
Fix asMemberOf throwing on intersection-override members
  
  When an interface inherits the same member from two supertypes nested in different classes, KSP2 attributed that member to the receiver's enclosing class instead of the real super-interface.
  asMemberOf's subtype check then failed and threw.

```
  class A { interface M { fun f(): String } }
  class B { interface M { fun f(): String } }
  class Box {
      interface Both : A.M, B.M   // intersection override of f()
  }
  // resolver.asMemberOf(Both.f, Both)  ->  used to throw IllegalArgumentException
```

  Fix: resolve the declaring class via intersectionOverriddenSymbols/fakeOverrideOriginal before the subtype check. Adds a regression case to asMemberOf golden tests.

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/ResolverAAImpl.kt
@@ -891,9 +891,28 @@ class ResolverAAImpl(
         }
     }
 
+    // Returns the class/interface that declares this member. For a fake or intersection override,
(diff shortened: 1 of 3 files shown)


jaschdoc:
Hi @vadymsavchenko-lyft, thank you for the PR! Please ping me when you have signed the CLA. Also, please split this PR in two, one with the test (in a new test file and processor) and one with a fix. See https://github.com/google/ksp/blob/main/CONTRIBUTING.md for more information. I couldn't find an issue for this, so unless there already is an existing one, please also open a new issue describing the bug. :)

PR #3089, opened 2026-07-28
Move kotlin-analysis-api single-artifact aa dependencies to the version catalog

KlyneChrysler:
Sixth slice of #2968 and the second of three for kotlin-analysis-api, following #3041.

The five single-artifact aa dependencies move to catalog accessors: guava, streamex, asm-all, stax2-api, and aalto-xml. Their five version properties are removed from gradle.properties since this module was their last reader.

Verified that resolution is unchanged: ./gradlew :kotlin-analysis-api:dependencies output for the compile and runtime classpaths is identical before and after this commit, 156 lines compared. :kotlin-analysis-api:compileKotlin passes.


jaschdoc (approved):
Thanks again @KlyneChrysler. I will also let @hfmehmed take a look.

PR #3090, opened 2026-07-28
Enable Gradle parallel tooling sync

jaschdoc:
What do you think about this @hfmehmed?


hfmehmed:
we set org.gradle.parallel for task parallelism which i think enables this by default?

jaschdoc:
> we set org.gradle.parallel for task parallelism which i think enables this by default?

Yes, I think you're right. I just re-read https://docs.gradle.org/current/userguide/performance.html#sec:configure_tooling_api_actions_parallelism

PR #3091, opened 2026-07-28
Update gradle to 9.5.0 stable

hfmehmed:


jaschdoc (approved):
Thanks! The sha256 sum matches the one from Gradle's website :)

PR #3092, opened 2026-07-28
fix: expose raw typed annotation lookup

### LINKED ISSUE #1038: Bug: ClassNotFoundException, but it does exist in the project
Hello there, I am trying to read my annotation using
```
classDeclaration.getAnnotationsByType(ActivityReference::class).first()
```

and getting this error

```
Caused by: com.google.devtools.ksp.KSTypeNotPresentException: java.lang.ClassNotFoundException: com.veepee.vpcore.route.BarImpl

```
This is a bit odd since the full qualified name in the exception seems to correct.

I tried [this](https://github.com/google/ksp/issues/888), but I still got the same error.

A project to reproduce it is attached [here](https://github.com/google/ksp/files/9077420/ksp-problem.zip)

To my specific needs I don't need to instantiate the annotation, but to get the String with the qualified name for `ActivityReference.activityName`, would it be possible with KSP ?

mvanhorn:
Add an experimental `getKSAnnotationsByType()` utility that reuses the existing short-name and qualified-name matching logic while returning `Sequence<KSAnnotation>` without constructing annotation proxies. `KSAnnotated.getAnnotationsByType()` still converts every matching `KSAnnotation` into a runtime annotation proxy.

A source annotation with one `KClass` member and an array of `KClass` members is found through `getKSAnnotationsByType()`, and the processor reads the expected qualified names from the raw `KSType` argument values without throwing; Built-in, source-declared, array, and generic array class literals remain distinguishable through the raw values, while the existing typed lookup continues to match the same annotation and retain its established proxy behavior.

Fixes #1038


jaschdoc:
Hi @mvanhorn, thanks for the PR! Could you add a test for this first in a separate PR? See https://github.com/google/ksp/blob/main/CONTRIBUTING.md for more information :)

mvanhorn:
Done, split out as #3108.

That PR is tests only, no production code, and the `EXPECTED` block records what KSP does today rather than what I want it to do, so the behaviour change shows up as a diff in that file when this PR lands on top.

It covers the two things around the existing case that were untested:

- `isAnnotationPresent` on the annotation whose `KClass` values are not loadable
- a second annotation nested in a private object that shares the simple name `ClassValueAnnotation` but not the qualified name, asserted through both `getAnnotationsByType(...).count()` and `isAnnotationPresent(...)`, which pins the lookup to qualified-name matching

Current behaviour, now recorded:

```
User
String, Company, IntArray, Array<User>
true
0
false
```

Both `AAConfiguredUnitTestSuite` and `PsiConfiguredUnitTestSuite` pass on unmodified upstream code.

I'll rebase this PR on #3108 once it lands so the diff here is just the behaviour change plus the two new expected lines.

PR #3095, opened 2026-07-29
Move remaining kotlin-analysis-api versions to the version catalog

KlyneChrysler:
Seventh slice of #2968 and the last of three for kotlin-analysis-api, following #3089.

fastutil, kotlinx-serialization-json, and the coroutines artifacts move to catalog accessors. The aa Kotlin and IntelliJ versions become catalog sourced local values since they version loop generated coordinates, kotlin() helper calls, and jar rename regexes that accessors cannot express. Five version properties with no remaining readers are removed from gradle.properties. kotlinBaseVersion and agpBaseVersion stay with a comment explaining the duplication: gradle-plugin's TestConfig loads gradle.properties directly at test runtime.

Verified that resolution is unchanged: ./gradlew :kotlin-analysis-api:dependencies output for the compile, runtime, and testRuntime classpaths is identical before and after this commit, 301 lines compared. :kotlin-analysis-api:compileTestKotlin and :symbol-processing-aa-embeddable:compileKotlin pass.

PR #3102, opened 2026-07-30
Move root build plugin versions to the version catalog

KlyneChrysler:
Final slice of #2968, following #3095. Related to #2968.

Two commits. The first moves the nexus-publish, shadow, dokka, and android-lint plugin versions to a [plugins] section and extends the CI Gradle cache keys to also hash libs.versions.toml so catalog only version bumps invalidate the cache. The second is the rename requested in the #3095 review: kotlinx-coroutines-coreJvm, with the asProvider disambiguation removed at all use sites.

With this slice every dependency and plugin version lives in the catalog. gradle.properties keeps only kotlinBaseVersion and agpBaseVersion, documented, because gradle-plugin's TestConfig loads the file directly at test runtime.

Verified: compileKotlin passes for kotlin-analysis-api, gradle-plugin, integration-tests, and symbol-processing-aa-embeddable with the renamed accessors, and the root plugins resolve. Alias renames cannot change resolution since the coordinates are unchanged.


jaschdoc (approved):
Thanks! Looks good to me. I will also let @hfmehmed take a look at it.

hfmehmed (approved):

jaschdoc:
Thank you for the PR again, @KlyneChrysler. Note, however, that there are still lots of dependencies that have yet to be moved to the version catalog, hence why I edited your PR description to not close the linked issue :)

PR #3104, opened 2026-07-30
Remove konanHome

### LINKED ISSUE #3098: Remove use of `konanHome` in KSP's Gradle plugin
The `konanHome` variable is deprecated as an error in Kotlin 2.4.0. This means that we cannot upgrade `kotlinBaseVersion`, which is used to interact with Kotlin's Gradle API, until `konanHome` is removed. This issue further blocks other upgrade issues due to Kotlin Metadata being outdated.

hfmehmed:
Remove deprecated `(it as KotlinNativeCompile).konanHome` property and manual Native klib resolution from KspAATask.

Since Kotlin 2.0 (`KT-61559`), KGP automatically includes the Kotlin/Native stdlib and platform klibs in kotlinCompilation.compileDependencyFiles.
Because KSP populates cfg.libraries directly from kotlinCompilation.compileDependencyFiles for Native targets, manually reading konanHome is no longer required.

### DIFF
--- a/gradle-plugin/lint-baseline.xml
+++ b/gradle-plugin/lint-baseline.xml
@@ -8,30 +8,30 @@
         errorLine2="                                                             ~~~">
         <location
             file="src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt"
-            line="109"
+            line="127"
             column="62"/>
     </issue>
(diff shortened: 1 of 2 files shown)


jaschdoc (approved):
Thank you! :)

PR #3105, opened 2026-07-30
Move kotlin-analysis-api IntelliJ platform dependencies to the version catalog

KlyneChrysler:
First slice of the inline build file versions remaining for #2968, continuing from #3102.

Five org.jetbrains.intellij.deps artifacts move from hardcoded versions to catalog accessors: jna, jna-platform, trove4j, log4j in the originalLog4j configuration, and jdom. The two jna artifacts share one version entry, and the jnaPlatform alias uses the camelCase leaf so no asProvider is needed, per the earlier reviews.

Verified that resolution is unchanged: ./gradlew :kotlin-analysis-api:dependencies output for the compile, runtime, and originalLog4j configurations is identical before and after this commit, 165 lines compared. :kotlin-analysis-api:compileKotlin passes.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,6 +14,10 @@ aa-stax2 = "4.2.1"
 aa-aalto-xml = "1.3.0"
 aa-streamex = "0.7.2"
 aa-fastutil = "8.5.14-jb1"
+aa-jna = "5.9.0.26"
+aa-trove4j = "1.0.20200330"
+aa-log4j = "1.2.17.2"
+aa-jdom = "2.0.6"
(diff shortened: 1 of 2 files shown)


jaschdoc (approved):
Thanks again! As per usual, I will let @hfmehmed take a look as well.

hfmehmed (approved):

jaschdoc:
@KlyneChrysler, if you can avoid rebasing or merging `main` unless there are merge conflicts, that would be great. In the future, it makes it easier and faster for us to merge your PR :)

jaschdoc (approved):

PR #3107, opened 2026-07-31
Bump build Kotlin version to 2.4.0

### LINKED ISSUE #2965: Upgrade to Kotlin 2.4.0
Quick experiment with this showed there is a build issue with the gradle plugin. It might also require bumping other dependencies.

jaschdoc:
Closes https://github.com/google/ksp/issues/2965

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,5 +10,5 @@ agpBaseVersion=9.3.0-alpha01
 kotlin.jvm.target.validation.mode=warning
 
 # Build or runtime dependencies of this project
-buildKotlinVersion=2.3.20
+buildKotlinVersion=2.4.0
 buildKspVersion=2.3.5
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/agp/AGPVersionBuiltInKotlinIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/agp/AGPVersionBuiltInKotlinIT.kt
@@ -30,10 +30,10 @@ class AGPVersionBuiltInKotlinIT(
         @Parameterized.Parameters(name = "AGP: {0}, KGP: {1}, Gradle: {2}, experimentalPsiResolution: {3}")
         fun data(): Collection<Array<String>> {
             return listOf(
-                arrayOf("9.0.0-alpha14", "2.2.10", "9.1.0"),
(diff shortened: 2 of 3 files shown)


jaschdoc:
@hfmehmed, this PR breaks all integration tests that use Kotlin `2.2.x` since the meta data is no longer readable by the `2.2.x` Kotlin compiler. Is that acceptable or what are your thoughts on it?

Note that this PR bumps the version of Kotlin used to build KSP to `2.4.0`. We have some integration tests that use `2.2.x` to build the workload / sample projects in our integration tests. They are now incompatible due to the meta data.

hfmehmed:
Acceptable or not that's the policy Jetbrains have i think. The compile binary compatbility rule states that a kotlin compiler of version N can read metadata from compiled libraries up to version N+1. So compiler 2.2.x can ready metadata from 2.3.x at most. 

We should drop tests that use 2.2.x and lower in this case i guess. Checkout https://github.com/google/ksp/pull/3109

jaschdoc:
I see, thanks! This seems to be another breaking change for `2.2.x` users, so is this a good idea to include it now?

hfmehmed:
I think so, the symbol-processing-api jar is too coupled to the version of kotlin we compile things with so stdlib gets bundled in the final JAR which essentially forces anyone using 2.2.x to stick with current version of KSP. 

Now that KSP releases are uncoupled from Kotlin we have the problem to keep some support levels (e.g promises to support Kotlin releases at least 1 year old and maybe at most 1.5 years old?). Or we need to find some smart way to be able to compile KSP with newer Kotlin version but still allow users to use old KGP/Kotlin versions but i guess that could lead to tricky classloader situations

jaschdoc:
We already upgraded from 2.2.10 to 2.3.20 this year alone without causing much trouble so I suppose it's okay-ish to upgrade now.

jaschdoc:
I agree we should have a clear policy for when to upgrade Kotlin so users know exactly which versions work with their build when upgrading.

troelsbjerre (dismissed):

jaschdoc:
@troelsbjerre @hfmehmed I think we should merge this. Merging it is more aligned with the current latest version of the Kotlin compiler. Please review and approve if you agree (only one approval is necessary).

troelsbjerre (approved):

PR #3109, opened 2026-08-03
Update kotlin version to 2.4.0

hfmehmed:
Drop any tests that are using 2.2.x or lower


jaschdoc:
Closing since we have https://github.com/google/ksp/pull/3107

Issue #2472, opened 2025-05-25
Support getting context parameters


lsafer-meemer:
Context parameters is now in Kotlin 2.2.0-RC as a preview feature and getting extension parameters is currently not possible in [2.2.0-RC-2.0.1](https://github.com/google/ksp/releases/tag/2.2.0-RC-2.0.1)

vdshb:
Since context parameters are stable, it would be nice to support them in KSP. Are there any rough estimations/plans on this?

jaschdoc:
Yes, it's hard to say when exactly the feature is ready, but it is actively being worked on in KSP. The first milestone is supporting backing fields. See https://github.com/google/ksp/pull/2969

PR #3110, opened 2026-08-03
Allow processors to opt in to new features in Resolver

jaschdoc:
This PR addresses the situation where a unchanged processor changes its behavior because it relies on `Resolver`. The idea is that the processor gets a lambda that registers it for new features in the `Resolver`. The new lambda is introduced in `SymbolProcessorEnvironment` but does not break binary compatibility. The reason is that `Resolver.getSymbolsWithAnnotation` could start returning backing fields and context parameters, however, this flag allows the Resolver to adapt its behavior (at runtime) depending on whether or not the current processor has registered itself for new features. In this design, a processor that has chosen *not* to opt in can coexist with a processor that *has* opted in.

- [x] Extends the `SymbolProcessorEnvironment` with a lambda that allows processors to register for new features in `Resolver`. This alleviates the problem that `Resolver.getSymbolsWithAnnotation` changes its behavior when `KSBackingField` and `KSContextParameter` is properly introduced.
- [x] Update `ResolverAAImpl` to adapt according to the current processor's opt-in choice.
- [x] Update `AnnotationResolutionStrategy` implementations with feature flag. See https://github.com/google/ksp/pull/3131
- [ ] Update test suites to be configured with this new mode of operation. See https://github.com/google/ksp/pull/3132

Related to https://github.com/google/ksp/pull/2969
Related to https://github.com/google/ksp/issues/2472
Related to https://github.com/google/ksp/issues/2873

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorEnvironment.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorEnvironment.kt
@@ -45,7 +45,39 @@ class SymbolProcessorEnvironment(
 
     /** KSP version */
     val kspVersion: KotlinVersion,
+
+    /**
+     * Processors may call this lambda to opt in to upcoming features:
+     * backing fields and context parameters.
+     */
+    val registerProcessorForNewFeatures: (SymbolProcessor) -> Unit
 ) {
+
+    /**
+     * Secondary constructor for binary compatibility that assigns
+     * [registerProcessorForNewFeatures] a lambda without effect.
+     */
+    constructor(
+        options: Map<String, String>,
+        kotlinVersion: KotlinVersion,
+        codeGenerator: CodeGenerator,
+        logger: KSPLogger,
+        apiVersion: KotlinVersion,
+        compilerVersion: KotlinVersion,
+        platforms: List<PlatformInfo>,
+        kspVersion: KotlinVersion,
+    ) : this(
+        options,
+        kotlinVersion,
+        codeGenerator,
+        logger,
+        apiVersion,
+        compilerVersion,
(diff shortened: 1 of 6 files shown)


kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/impl/PsiResolutionNext.kt
jaschdoc, line 1: This class is almost a copy of `PsiResolutionStrategy`. It removes support for `inDepth = true` and adds support for backing fields. This is added to avoid duplicating a lot of functionality in `PsiResolutionStrategy`.

api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorEnvironment.kt
```diff
@@ -45,7 +45,39 @@ class SymbolProcessorEnvironment(
...
+    ) : this(
+        options,
+        kotlinVersion,
+        codeGenerator,
+        logger,
+        apiVersion,
+        compilerVersion,
+        platforms,
+        kspVersion,
+        registerProcessorForNewFeatures = { _ -> }
```
troelsbjerre: registerProcessorForNewFeatures = {}

troelsbjerre (changes requested):

api/src/main/kotlin/com/google/devtools/ksp/processing/SymbolProcessorEnvironment.kt
jaschdoc: Done

troelsbjerre (approved):

PR #3117, opened 2026-08-04
Add typed annotation values for KSValueArgument

Yong-yuan-X:
Refs #3008

This PR implements the initial `KSAnnotationValue` and `KSPrimitiveType` ADT proposed in #3008, with a `KSValueArgument.typedValue()` extension. It provides a typed view of annotation values without changing the existing `KSValueArgument.value` API or its `Array`/`Collection` behavior.

It also updates `api/api.base` and extends the existing annotation value test to exercise `typedValue()` with arrays, nested annotations, representative primitive values, an enum, and a class reference across the AA and PSI test suites.

This follows the proposed starting design. Feedback on the API shape and conversion semantics is welcome.

Tested locally:
- `:api:checkApi`
- `AAConfiguredUnitTestSuite.testAnnotationArrayValueType`
- `PsiConfiguredUnitTestSuite.testAnnotationArrayValueType`

### DIFF
--- a/api/src/main/kotlin/com/google/devtools/ksp/KSValueArgumentExtensions.kt
+++ b/api/src/main/kotlin/com/google/devtools/ksp/KSValueArgumentExtensions.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2026 Google LLC
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.devtools.ksp
+
+import com.google.devtools.ksp.symbol.AnnotationClass
+import com.google.devtools.ksp.symbol.ArrayValue
+import com.google.devtools.ksp.symbol.EnumClass
+import com.google.devtools.ksp.symbol.ErrorValue
+import com.google.devtools.ksp.symbol.KSAnnotation
+import com.google.devtools.ksp.symbol.KSAnnotationValue
+import com.google.devtools.ksp.symbol.KSBool
+import com.google.devtools.ksp.symbol.KSByte
+import com.google.devtools.ksp.symbol.KSChar
+import com.google.devtools.ksp.symbol.KSClassDeclaration
+import com.google.devtools.ksp.symbol.KSDouble
+import com.google.devtools.ksp.symbol.KSFloat
+import com.google.devtools.ksp.symbol.KSInt
+import com.google.devtools.ksp.symbol.KSLong
+import com.google.devtools.ksp.symbol.KSShort
+import com.google.devtools.ksp.symbol.KSString
+import com.google.devtools.ksp.symbol.KSType
+import com.google.devtools.ksp.symbol.KSUByte
+import com.google.devtools.ksp.symbol.KSUInt
+import com.google.devtools.ksp.symbol.KSULong
+import com.google.devtools.ksp.symbol.KSUShort
+import com.google.devtools.ksp.symbol.KSValueArgument
+import com.google.devtools.ksp.symbol.Primitive
+import com.google.devtools.ksp.symbol.ReflectionClassReference
+
+fun KSValueArgument.typedValue(): KSAnnotationValue = value.toKSAnnotationValue()
+
+private fun Any?.toKSAnnotationValue(): KSAnnotationValue =
+    when (this) {
+        is Boolean -> Primitive(KSBool(this))
+        is Byte -> Primitive(KSByte(this))
+        is Short -> Primitive(KSShort(this))
+        is Int -> Primitive(KSInt(this))
+        is Long -> Primitive(KSLong(this))
+        is Float -> Primitive(KSFloat(this))
+        is Double -> Primitive(KSDouble(this))
+        is Char -> Primitive(KSChar(this))
+        is String -> Primitive(KSString(this))
+        is UByte -> Primitive(KSUByte(this))
+        is UShort -> Primitive(KSUShort(this))
+        is UInt -> Primitive(KSUInt(this))
+        is ULong -> Primitive(KSULong(this))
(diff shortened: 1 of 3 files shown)


jaschdoc:
Hey @Yong-yuan-X, are you still interested in contributing this PR?

Yong-yuan-X:
Yes, still interested! Sorry for the delay — I’ve been quite busy with my job search recently. I’ve picked this back up and will continue working on the PR over the next couple of days.

api/api.base
jaschdoc, line 1: Please revert this file. We are specifically doing it this way to avoid updating the API surface.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/processor/AnnotationArrayValueTypeProcessor.kt
```diff
@@ -46,7 +55,38 @@ class AnnotationArrayValueTypeProcessor(override val enableNewFeatures: Boolean)
...
+
+        val mode = arguments.getValue("mode").typedValue() as? EnumClass
+        check(mode?.clazz?.simpleName?.asString() == "FIRST") { "Unexpected typed enum: $mode" }
+        results.add("KotlinAnnotated TypedAnnotation mode: EnumClass(FIRST)")
+
+        val klass = arguments.getValue("klass").typedValue() as? ReflectionClassReference
+        check(klass?.type?.declaration?.simpleName?.asString() == "String") {
+            "Unexpected typed class reference: $klass"
         }
+        results.add("KotlinAnnotated TypedAnnotation klass: ReflectionClassReference(String)")
```
jaschdoc: Let's add all of this in a new test processor. Then we can discuss what to put in the test processor. Ideally, we should cover every single branch.

kotlin-analysis-api/testData/annotationArrayValueType.kt
jaschdoc, line 1: Let's make a copy of this file and add the new expected results there. Also make sure to add a comment in the new file that it is a copy of `annotationArrayValueType.kt` but with a different expected output. The comment should also mention that the two files should be kept in sync.

api/src/main/kotlin/com/google/devtools/ksp/KSValueArgumentExtensions.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.google.devtools.ksp.symbol.KSUInt
+import com.google.devtools.ksp.symbol.KSULong
+import com.google.devtools.ksp.symbol.KSUShort
+import com.google.devtools.ksp.symbol.KSValueArgument
+import com.google.devtools.ksp.symbol.Primitive
+import com.google.devtools.ksp.symbol.ReflectionClassReference
+
+fun KSValueArgument.typedValue(): KSAnnotationValue = value.toKSAnnotationValue()
+
+private fun Any?.toKSAnnotationValue(): KSAnnotationValue =
```
jaschdoc, line 57: Let's document this function by explaining how types are converted and wrapped.

api/src/main/kotlin/com/google/devtools/ksp/KSValueArgumentExtensions.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.google.devtools.ksp.symbol.KSType
+import com.google.devtools.ksp.symbol.KSUByte
+import com.google.devtools.ksp.symbol.KSUInt
+import com.google.devtools.ksp.symbol.KSULong
+import com.google.devtools.ksp.symbol.KSUShort
+import com.google.devtools.ksp.symbol.KSValueArgument
+import com.google.devtools.ksp.symbol.Primitive
+import com.google.devtools.ksp.symbol.ReflectionClassReference
+
+fun KSValueArgument.typedValue(): KSAnnotationValue = value.toKSAnnotationValue()
```
jaschdoc, line 48: Let's also document this function and reference that it should be preferred over the `value` property on `KSValueArgument`.

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSPrimitiveType.kt
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.devtools.ksp.symbol
+
+sealed interface KSPrimitiveType
```
jaschdoc, line 20: Please also add some brief documentation of this interface. I do not think its subtypes require the same treatment though.

api/src/main/kotlin/com/google/devtools/ksp/symbol/KSAnnotationValue.kt
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.devtools.ksp.symbol
+
+sealed interface KSAnnotationValue
```
jaschdoc, line 20: Please add brief documentation for this interface and its subtypes.

jaschdoc (commented):
Thank you for the PR! Looks good but I have a handful of minor changes that'd be great to have before merging it.

Yong-yuan-X:
Thanks for the thoughtful and detailed review! I really appreciate the care you’ve put into reviewing this PR.

Updated in `cbdc8b2` — reverted the `api/api.base` changes, moved the `typedValue()` checks into a separate test processor and fixture, and added the requested documentation.

The API check and the annotation array tests passed locally across both AA and PSI, with new features enabled and disabled. I’ll follow up in the existing review thread about the remaining test coverage.

kotlin-analysis-api/src/test/kotlin/com/google/devtools/ksp/processor/AnnotationArrayValueTypeProcessor.kt
Yong-yuan-X: I’ve moved the typedValue() checks into AnnotationArrayTypedValueProcessor with a separate fixture. The current tests cover array-valued arguments with nested annotations, plus String, Int, enum, and class-reference values.

My proposed next step is to extend this processor with tests for the remaining primitive conversions, checking both the wrapper type and the wrapped value. 

For branches requiring specific inputs, one option would be to use KSValueArgument instances with controlled value inputs to exercise both Array and Collection paths, unsigned values, and the null/unsupported-value fallback. Would you prefer keeping these controlled-input cases in the processor or moving them into a separate unit test?

PR #3127, opened 2026-08-08
Move five kotlin-analysis-api dependencies to the version catalog

KlyneChrysler:
Continuing the inline build file versions for #2968, following #3105.

Five hardcoded dependency versions in kotlin-analysis-api move to catalog accessors: kotlinx-collections-immutable, caffeine, javaslang, javax.inject, and kotlin-reflect.

Verified that resolution is unchanged: ./gradlew :kotlin-analysis-api:dependencies output for the compile and runtime classpaths is identical before and after this commit, 156 lines compared. :kotlin-analysis-api:compileKotlin passes.

### DIFF
--- a/kotlin-analysis-api/build.gradle.kts
+++ b/kotlin-analysis-api/build.gradle.kts
@@ -157,7 +157,7 @@ dependencies {
             depSourceJars("$it:$aaKotlinBaseVersion:sources") { isTransitive = false }
         }
 
-    implementation("org.jetbrains.kotlinx:kotlinx-collections-immutable-jvm:0.3.4")
+    implementation(libs.aa.collectionsImmutable)
     implementation(libs.kotlinx.serialization.json)
     compileOnly(kotlin("stdlib", aaKotlinBaseVersion))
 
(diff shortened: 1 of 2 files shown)


jaschdoc (approved):
Thank you for the PR. @hfmehmed, could you also take a look? Thanks! :)

kotlin-analysis-api/build.gradle.kts
```diff
@@ -166,16 +166,16 @@ dependencies {
...
     implementation(libs.aa.jna) { isTransitive = false }
     implementation(libs.aa.jnaPlatform) { isTransitive = false }
     implementation(libs.aa.trove4j) { isTransitive = false }
     originalLog4j(libs.aa.log4j) { isTransitive = false }
     implementation(files(filteredLog4j))
     implementation(libs.aa.jdom) { isTransitive = false }
-    implementation("io.javaslang:javaslang:2.0.6")
-    implementation("javax.inject:javax.inject:1")
-    implementation("org.jetbrains.kotlin:kotlin-reflect:1.6.10")
+    implementation(libs.aa.javaslang)
```
hfmehmed, line 176: do we use this dependency anywhere at all?

kotlin-analysis-api/build.gradle.kts
```diff
@@ -166,16 +166,16 @@ dependencies {
...
+    implementation(libs.aa.javaxInject)
+    implementation(libs.aa.kotlinReflect)
```
hfmehmed, line 178: similarly do we use this kotlin-reflect? if we use it, can you check if it is for tests only and if so move it to testImplementation.

hfmehmed:
Sorry for the delay on this @KlyneChrysler . Thank you for doing this - can i ask you to also check if the 2 dependency i outlined are used.

KlyneChrysler:
javaslang has zero references anywhere in the repo so it can be dropped, and kotlin-reflect is used only by kotlin.reflect.full.isSubclassOf in the tests so it belongs on testImplementation, both verified by compiling main and tests with javaslang removed and kotlin-reflect moved to test.

jaschdoc (changes requested):
Sounds good. I think removing the `kotlin-reflect` dependency might also solve a few issues users have with conflicting class paths, although I'm not sure about that. In any case, it's good to remove unused dependencies. I'm worried, however, that these dependencies are inherited from Kotlin/AA. Let's try to remove the dependencies and see what happens. :)

EDIT: I cannot find any reference to `javaslang` in the Kotlin repo either.

KlyneChrysler:
Done, removed both entirely rather than cataloging them. You were right that they are inherited from Kotlin/Analysis-API: main and the tests both still compile with the two explicit declarations gone, and the test's kotlin.reflect.full.isSubclassOf resolves transitively. Left it to CI to confirm the full suite.

hfmehmed (approved):

hfmehmed:
@jaschdoc please review again and merge if all good from your end

jaschdoc (approved):
Thank you for updating the PR. Looks good :)

PR #3144, opened 2026-08-19
Fix implicit task dependency error between KSP and Android Lint in Android KMP

### LINKED ISSUE #3128: KSP 2.3.11 causes implicit task dependency failures in Android-KMP host-test lint tasks
Description

 ## Summary
 
 Upgrading KSP from 2.3.10 to 2.3.11 breaks `check` in an Android-KMP library using the standalone lint plugin.
 
 The Android host-test lint tasks consume output directories from `kspAndroidHostTest` without declaring a dependency on that task.
 
 ## Reproducer
 
 https://github.com/alyssoncs/KspBug
 
 The repository contains two independent, source-identical modules. Their build files differ only by the KSP version.
 
 ```shell
 ./gradlew :previous:check # KSP 2.3.10: passes
 ./gradlew :latest:check   # KSP 2.3.11: fails

Actual behavior

 > Task :latest:generateAndroidHostTestLintModel FAILED
 > Task :latest:lintAnalyzeAndroidHostTest FAILED
 
 Task ':latest:lintAnalyzeAndroidHostTest' uses this output of task
 ':latest:kspAndroidHostTest' without declaring an explicit or implicit dependency:
 
 latest/build/generated/ksp/android/androidHostTest/kotlin
 latest/build/generated/ksp/android/androidHostTest/java
 
 Task ':latest:generateAndroidHostTestLintModel' reports the same problem.

Expected behavior

Applying KSP must not produce an invalid Gradle task graph.  check  should pass as it does with KSP 2.3.10.

Environment

 - KSP: 2.3.10 passes; 2.3.11 fails
 - Kotlin: 2.4.10
 - Android Gradle Plugin: 9.3.1
 - Gradle: 9.7.0
 - Plugins:
  -  org.jetbrains.kotlin.multiplatform 
  -  com.android.kotlin.multiplatform.library 
  -  com.android.lint 

Additional findings

 - No symbol processor dependency is required to reproduce the failure. This is intentional:  androidHostTest  has an empty processor classpath, which is a valid target-specific KSP configuration.
 -  com.android.lint  is load-bearing because the Android-KMP library plugin does not add lint verification to  check  by itself.
 - The equivalent project using normal  com.android.library  passes  check  with both KSP 2.3.10 and 2.3.11. The regression appears specific to Android-KMP host-test lint integration.
 - KSP 2.3.11 changed empty-processor tasks to execute cleanup instead of being skipped as part of #3046. That change appears to expose the missing task dependency.

hfmehmed:
For android-kmp projects, KSP is adding generated output directories directly to `defaultSourceSet.kotlin.srcDirs`, causing AGP to classify them as static (non-generated) sources. AGP Lint tasks collect all non-generated sources as inputs, leading Gradle to detect task inputs depending on KSP outputs without task dependencies.

Fixes #3128

### DIFF
--- a/integration-tests/src/test/resources/playground-mpp/workload/src/androidHostTest/kotlin/com/example/AndroidHostTest.kt
+++ b/integration-tests/src/test/resources/playground-mpp/workload/src/androidHostTest/kotlin/com/example/AndroidHostTest.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2026 Google LLC
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+
+class AndroidHostTest {
+    @Test
+    fun testGeneratedBuilder() {
+        val builder = AClassBuilder()
+        builder
+            .withA(1)
+            .withB("foo")
+            .withC(2.3)
+            .withD(hello.HELLO())
+        val aClass = builder.build()
+        assertNotNull(aClass)
+        assertEquals("1, foo, 2.3", aClass.p)
+    }
+}
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt
@@ -207,6 +207,9 @@ class KspGradleSubplugin @Inject internal constructor(private val registry: Tool
             kspExtension,
         )
 
+        val isAndroidKmp = project.isAndroidKmpProject() &&
+            kotlinCompilation is KotlinMultiplatformAndroidCompilation
+
         val generatedSources = arrayOf(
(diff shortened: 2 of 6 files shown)


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
```diff
@@ -266,15 +266,27 @@ object AndroidPluginIntegration {
...
+                androidComponent.sources.java?.addGeneratedSourceDirectory(
+                    taskProvider = kspTaskProvider,
+                    wiredWith = { task -> task.kspConfig.javaOutputDir }
+                )
+            } else {
+                androidComponent.sources.kotlin?.addGeneratedSourceDirectory(
+                    taskProvider = kspTaskProvider,
+                    wiredWith = { task -> task.kspConfig.javaOutputDir }
+                )
+            }
```
jaschdoc: This seems a little puzzling to me. Why are we adding the Java output dir to the Kotlin sources?

integration-tests/src/test/resources/playground-mpp/workload/src/androidHostTest/kotlin/com/example/AndroidHostTest.kt
jaschdoc, line 1: Nit: missing copyright header :)

hfmehmed: Ack

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/AndroidPluginIntegration.kt
hfmehmed: Ignore this, i am experimenting with something - its likely not needed

jaschdoc: Sounds good. Ping me when you need a review :)

hfmehmed: ready now

jaschdoc (approved):
Looks good. Please rebase on main as I have updated some required status checks, and the workflow won't succeed without the latest changes from main.

jaschdoc (approved):
Rebased on main

PR #3145, opened 2026-08-20
remove link to "why ksp"

budindepunk:
"Why KSP" page was deleted from the docs as per Kotlin docs issue [KEDOC-323](https://youtrack.jetbrains.com/issue/KEDOC-323/Delete-Why-KSP-from-KSP-docs)


jaschdoc (approved):
Thanks for updating our docs as well! :)

PR #3146, opened 2026-08-20
fix: resolve annotation default values from KMP metadata

### LINKED ISSUE #2356: KSP2 version 2.1.20-RC-1.0.30 crashes with NoSuchElementException: Collection contains no element matching the predicate
After migration from KSP1 to KSP2 I have a problem when I try to read a parameter of an annotation on a class, which has a default value. The annotation is defined like this:
```
@Target(AnnotationTarget.CLASS)
public annotation class SimpleHtmlComponent(val tagName: String, val withText: Boolean = false)
```
and the code of the processor looks like this:
```
val simpleHtmlComponent = classDeclaration.getAnnotationsByType(SimpleHtmlComponent::class).first()
val tagName = simpleHtmlComponent.tagName
val withText = simpleHtmlComponent.withText
```
The last line throws the exception:
```
Caused by: java.util.NoSuchElementException: Collection contains no element matching the predicate.
        at com.google.devtools.ksp.UtilsKt.createInvocationHandler$lambda$8(utils.kt:587)
        at jdk.proxy10/jdk.proxy10.$Proxy212.withText(Unknown Source)
        at dev.kilua.ksp.KiluaProcessor.process(KiluaProcessor.kt:62)
        at com.google.devtools.ksp.impl.KotlinSymbolProcessing$execute$1$1.invoke(KotlinSymbolProcessing.kt:581)
        at com.google.devtools.ksp.impl.KotlinSymbolProcessing$execute$1$1.invoke(KotlinSymbolProcessing.kt:579)
        at ksp.com.google.devtools.ksp.common.IncrementalContextBase.closeFilesOnException(IncrementalContextBase.kt:404)
        at com.google.devtools.ksp.impl.KotlinSymbolProcessing.execute(KotlinSymbolProcessing.kt:579)
        at com.google.devtools.ksp.impl.KSPLoader$Companion.loadAndRunKSP(KSPLoader.kt:37)
        at com.google.devtools.ksp.impl.KSPLoader.loadAndRunKSP(KSPLoader.kt)

```
To reproduce:
1. Clone https://github.com/rjaros/kilua project
2. Checkout `ksp2issue` branch
3. Run `./gradlew allTest --stacktrace`
4. See the crash

The annotations are in the `modules/kilua-annotations` module. The processor in the `plugins/kilua-ksp-processor` module.

The same code works fine with KSP1.

mvanhorn:
Update the existing default-value conversion in `util.kt` to consume a deserialized FIR value parameter's metadata-backed default expression on every target, converting it through the same FIR annotation-value machinery already used for JVM defaults. KSP2 omits default arguments declared by annotation classes in compiled Kotlin libraries when processing a non-JVM KMP target.

On a Native target, read an annotation from a compiled library through `getAnnotationsByType` and verify an omitted Boolean member returns its declared value instead of throwing; Omit a vararg annotation member and verify the proxy exposes an empty array, covering the empty-collection failure reported in the thread.

Fixes #2356

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/util.kt
@@ -126,6 +126,7 @@ import org.jetbrains.kotlin.codegen.state.md5base64
 import org.jetbrains.kotlin.fir.declarations.FirRegularClass
 import org.jetbrains.kotlin.fir.declarations.FirResolvePhase
 import org.jetbrains.kotlin.fir.declarations.utils.moduleName
+import org.jetbrains.kotlin.fir.expressions.impl.FirExpressionStub
 import org.jetbrains.kotlin.fir.java.JavaTypeParameterStack
 import org.jetbrains.kotlin.fir.java.toFirExpression
 import org.jetbrains.kotlin.fir.resolve.getContainingClassSymbol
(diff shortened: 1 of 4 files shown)


jaschdoc:
Fixed by https://github.com/google/ksp/pull/3200

PR #3164, opened 2026-08-25
Reproduce NPE in package name for explicit backing fields

jaschdoc:
Reproduces https://github.com/google/ksp/issues/3165
Related to #2873

PR #3169, opened 2026-08-26
Fix missing `packageName` in `KSBackingFieldImpl`

### LINKED ISSUE #3165: Package name for `KSBackingField` may cause NPE
If an **implicit** backing field is targeted by an annotation, accessing its package name causes an NPE. This only happens when backing fields are enabled.

```kotlin
package com.example

annotation class MyAnnotation

@field:MyAnnotation
val x = 42
```

The reason is that `AbstractKSDeclarationImpl` calls `KaSymbol.toContainingFile` and unsafely casts it to non-null. However, `KaSymbol.toContainingFile` returns `null` if there is no Psi element for it, which is the case for implicit backing fields.

```kotlin
internal fun KaSymbol.toContainingFile(): KSFile? {
    return when (val psi = this.psi) {
        is KtElement -> analyze {
            KSFileImpl.getCached(psi.containingKtFile.symbol)
        }

        is PsiElement -> KSFileJavaImpl.getCached(psi.containingFile as PsiJavaFile)
        else -> null
    }
}

// ...

class AbstractKSDeclarationImpl(...) {
    override val containingFile: KSFile? by lazy {
        ktDeclarationSymbol.toContainingFile()
    }

    override val packageName: KSName by lazy {
        // source
        if (origin == Origin.KOTLIN || origin == Origin.JAVA) {
            containingFile!!.packageName // NPE thrown here
        }
    // ...
    }
}
```

This does not happen for Java fields since the backing field symbol has a Psi element. There is a 1:1 correspondence between properties and backing fields in Java.

The fix is to override `containingFile` in `KSBackingFieldImpl`. The implementation should just default to the owning property's containing file, since they always correspond to the same file. Additionally, using the owner's property also leads to fewer allocations and will cache the computation for the owning property as well, thereby skipping recomputation if the owning property's containing file is accessed.

jaschdoc:
Depends on https://github.com/google/ksp/pull/3164
Fixes https://github.com/google/ksp/issues/3165

### DIFF
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSBackingFieldImpl.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/symbol/kotlin/KSBackingFieldImpl.kt
@@ -25,6 +25,7 @@ import com.google.devtools.ksp.impl.symbol.kotlin.resolved.KSTypeReferenceResolv
 import com.google.devtools.ksp.symbol.KSAnnotation
 import com.google.devtools.ksp.symbol.KSBackingField
 import com.google.devtools.ksp.symbol.KSExpectActual
+import com.google.devtools.ksp.symbol.KSFile
 import com.google.devtools.ksp.symbol.KSName
 import com.google.devtools.ksp.symbol.KSPropertyDeclaration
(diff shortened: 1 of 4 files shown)


jaschdoc:
This PR will likely conflict with https://github.com/google/ksp/pull/3170

troelsbjerre (approved):

PR #3174, opened 2026-09-01
Move remaining kotlin-analysis-api and gradle-plugin inline versions to the version catalog

KlyneChrysler:
Following #3127. This moves the last inline dependency versions in the production modules to the catalog, so no version literal remains in any non-fixture build file.

kotlin-analysis-api: annotations, lz4-java, opentelemetry-api. gradle-plugin: the lint-gradle lintChecks dependency.

Unlike the previous slice, all four are in use, so they are cataloged rather than removed: annotations is used directly in source, and lz4 and opentelemetry are isTransitive false runtime pins for the IntelliJ platform (opentelemetry is also relocated into the embeddable jar), so dropping them would risk a runtime NoClassDefFoundError.

Verified resolution is unchanged: :kotlin-analysis-api:dependencies for compile and runtime and :gradle-plugin:dependencies for lintChecks are identical before and after, 157 lines compared. Both modules compile.

The only versions still declared inline live in benchmark/exhaustive-processor, which I left alone as a benchmark fixture, consistent with excluding the examples and test resource projects. Happy to migrate it too if you would like it in scope for this issue.

### DIFF
--- a/gradle-plugin/build.gradle.kts
+++ b/gradle-plugin/build.gradle.kts
@@ -33,7 +33,7 @@ dependencies {
     testImplementation(libs.google.truth)
     testImplementation(gradleTestKit())
 
-    lintChecks("androidx.lint:lint-gradle:1.0.0-alpha05")
+    lintChecks(libs.androidx.lintGradle)
 }
 
 kotlin {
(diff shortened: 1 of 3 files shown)


jaschdoc (approved):
Thanks, looks good! @hfmehmed please take a look as well :)

hfmehmed (approved):

PR #3189, opened 2026-09-05
Honor effective Gradle Isolated Projects flags in KSP source wiring

### LINKED ISSUE #3188: Command-line Isolated Projects flags are not reflected in KSP source wiring
KSP's `enableProjectIsolationCompatibleCodepath()` reads Gradle/system properties directly, so it misses Gradle's `--isolated-projects` command-line flag. It also reports isolation enabled when `org.gradle.isolated-projects=true` is overridden by `--no-isolated-projects`.

This is present in KSP 2.3.11 and current main. The compatible source-registration path is selected by this function in both `KspSubplugin` and `KspAATask`.

A small Groovy build demonstrates the detection error:

```groovy
import com.google.devtools.ksp.gradle.utils.KgpUtilsKt

plugins {
    id 'org.jetbrains.kotlin.jvm' version '2.3.20' apply false
    id 'com.google.devtools.ksp' version '2.3.11' apply false
}

def isolated = KgpUtilsKt.enableProjectIsolationCompatibleCodepath(project)
tasks.register('checkIsolation') {
    inputs.property('isolated', isolated)
    doLast { println('KSP isolation: ' + inputs.properties.isolated) }
}
```

With Gradle 9.7.1:

- `./gradlew checkIsolation --isolated-projects` prints Gradle's isolation banner, but `KSP isolation: false`.
- Put `org.gradle.isolated-projects=true` in `gradle.properties`; `./gradlew checkIsolation --no-isolated-projects` prints `KSP isolation: true`.
- Explicit `ksp.project.isolation.enabled=true` remains a useful opt-in/workaround and should continue to work with Gradle isolation disabled.

I have reproduced both failures in focused TestKit tests against current main. `BuildFeatures.isolatedProjects.active` supplies Gradle's effective feature state, including CLI precedence. It is available since Gradle 8.5, so a version-gated injected helper can use it while retaining the existing property fallback on older Gradle versions.

I am preparing that small fix with CLI/cache-reuse, explicit opt-in, older-Gradle, and actual generated-source compilation coverage. This investigation and the tests are substantially AI-assisted with Codex.

gregjotau:
fix #3188

### DIFF
--- a/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/kgpUtils.kt
+++ b/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/utils/kgpUtils.kt
@@ -1,6 +1,9 @@
 package com.google.devtools.ksp.gradle.utils
 
+import javax.inject.Inject
 import org.gradle.api.Project
+import org.gradle.api.configuration.BuildFeatures
+import org.gradle.util.GradleVersion
 import org.jetbrains.kotlin.gradle.plugin.KotlinBaseApiPlugin
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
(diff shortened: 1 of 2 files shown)


gregjotau:
The current property-only check does not see `--isolated-projects`, and it ignores `--no-isolated-projects` overriding a property. This patch uses injected `BuildFeatures.isolatedProjects.active` on Gradle 8.5+ and keeps the existing property fallback on older Gradle versions. The explicit `ksp.project.isolation.enabled=true` opt-in is preserved even when Gradle isolation is disabled.

Before the implementation change, the focused tests reproduced both wrong values: the CLI enable test returned false and the CLI disable-over-property test returned true. With the change, all six tests pass, with no skips:

- CLI enable and configuration-cache reuse
- CLI disable overriding the Gradle property
- explicit KSP opt-in with isolation disabled
- Gradle 8.4 compatibility before BuildFeatures exists
- the legacy isolation property on Gradle 8.5
- a two-project processor/consumer build compiling actual generated Kotlin classes with only `--isolated-projects`, then reusing its configuration cache

Validation (JDK 17):

```
./gradlew :gradle-plugin:validatePlugins :gradle-plugin:test --tests '*ProjectIsolationTest*'
```

The changed Kotlin files also pass ktfmt 0.63 with the repository's kotlinlang style. Tests consume the build's test Maven publications, using the real plugin classpath and no TestKit classloader injection. No repository-wide test claim is made.

This patch and its tests were substantially prepared and tested with Codex. The draft is available for human review; any required contributor agreement must be completed by the contributor, not the agent.

Mon322812 (approved):

gregjotau:
@Mon322812 Can you just open this and get it merged, I dont care about contribution here and the CLA. I just want it fixed :)

Issue #3140, opened 2026-08-19
`PluginProblemReporter` is not registered as application service


jaschdoc:
As seen in https://github.com/google/ksp/pull/3049 and https://github.com/google/ksp/pull/3138, KSP can crash if `PluginProblemReporter` is transitively used. The fix should be to register it as an application service such that `PluginProblemRepoerter.getInstance()` returns a valid instance.

PR #3195, opened 2026-09-07
On error run type checking to improve diagnostics

### LINKED ISSUE #3196: Improve diagnostics by returning last round dismissible errors when KSP fails.
I'm getting a few reports with the same root cause. The initial problem is a syntax error at the application level. Sometimes we find actual problems with toolchain, code generation or else. Sometimes is pure syntax problems. Ideally we want to improve diagnostics to make it easier for devs to find problems with their code.

We already implemented last round dismissible errors for KAPT and now I added a [typeCheck](https://github.com/google/ksp/pull/3195) function for KSP so we get this error details. Please let me know what you think.

alfonsocv12:
I'm getting a few reports related to Codegeneration throwing, when diving deeper into the issue, most of the time its due to a incorrect syntax at the application level. Showing the dismissible errors when KSP fails will improve diagnostic, reducing the number of application level syntax error reports.

### DIFF
--- a/gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
+++ b/gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
@@ -82,6 +82,102 @@ class GradleCompilationTest(isExperimentalPsiResolution: Boolean) {
     }
 
     @Test
+    fun missingTypeReportedOnFailure() {
+        testRule.setupAppAsJvmApp()
+        testRule.appModule.dependencies.add(
+            module(configuration = "ksp", testRule.processorModule)
+        )
+        testRule.appModule.addSource(
+            "TestFailure.kt",
+            """
+            package com.example
+
+            class TestFailure {
+                val field: UnresolvedClass = UnresolvedClass()
+            }
+            """.trimIndent()
+        )
+        class ErrorReporting(private val logger: KSPLogger) : SymbolProcessor {
+            override fun process(resolver: Resolver): List<KSAnnotated> {
+                logger.error("Fatal KSP processor error!")
+                return emptyList()
+            }
+        }
+
+        class Provider : TestSymbolProcessorProvider({ env -> ErrorReporting(env.logger) })
+
+        testRule.addProvider(Provider::class)
+        val failure = testRule.runner()
+            .withArguments("app:assemble")
+            .buildAndFail()
+        assertThat(failure.output).contains("Fatal KSP processor error!")
+        assertThat(failure.output).contains("TestFailure.kt:4: Unresolved reference \'UnresolvedClass\'")
+    }
+
+    @Test
+    fun onlyLastRoundFilesAreCheckedOnFailure() {
+        testRule.setupAppAsJvmApp()
+        testRule.appModule.dependencies.add(
+            module(configuration = "ksp", testRule.processorModule)
+        )
+        testRule.appModule.addSource(
+            "Round0File.kt",
+            """
+            package com.example
+
+            class Round0File {
+                val round0Field: UnresolvedInRound0 = UnresolvedInRound0()
+            }
+            """.trimIndent()
+        )
+        class MultiRoundProcessor(
+            private val codeGenerator: CodeGenerator,
+            private val logger: KSPLogger,
+        ) : SymbolProcessor {
+            private var round = 0
+
+            override fun process(resolver: Resolver): List<KSAnnotated> {
+                if (round == 0) {
+                    round++
+                    codeGenerator.createNewFile(
+                        Dependencies.ALL_FILES,
+                        "com.example",
+                        "GeneratedInRound0"
+                    ).use { stream ->
+                        stream.writer(Charsets.UTF_8).use {
+                            it.write(
+                                """
+                                package com.example
+
+                                class GeneratedInRound0 {
+                                    val round1Field: UnresolvedInRound1 = UnresolvedInRound1()
+                                }
+                                """.trimIndent()
+                            )
+                        }
+                    }
+                } else {
+                    logger.error("Fatal KSP processor error in round 1!")
+                }
+                return emptyList()
+            }
+        }
+
+        class Provider : TestSymbolProcessorProvider({ env ->
+            MultiRoundProcessor(env.codeGenerator, env.logger)
+        })
+
+        testRule.addProvider(Provider::class)
+        val failure = testRule.runner()
+            .withArguments("app:assemble")
+            .buildAndFail()
+
+        assertThat(failure.output).contains("Fatal KSP processor error in round 1!")
+        assertThat(failure.output).contains("GeneratedInRound0.kt:4: Unresolved reference 'UnresolvedInRound1'.")
+        assertThat(failure.output).doesNotContain("UnresolvedInRound0")
+    }
+
+    @Test
     fun applicationCanAccessGeneratedCode_multiplatform_withConfigCache() {
         testRule.setupAppAsMultiplatformApp(
             """
--- a/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/primary/KMPImplementedIT.kt
+++ b/integration-tests/src/test/kotlin/com/google/devtools/ksp/test/primary/KMPImplementedIT.kt
@@ -4,6 +4,9 @@ import com.google.devtools.ksp.test.fixtures.TemporaryTestProject
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
 import org.gradle.testkit.runner.TaskOutcome
+import org.hamcrest.CoreMatchers.containsString
+import org.hamcrest.CoreMatchers.hasItem
+import org.hamcrest.MatcherAssert.assertThat
 import org.junit.Assert
 import org.junit.Assume
 import org.junit.Ignore
@@ -117,7 +120,9 @@ class KMPImplementedIT(experimentalPsiResolution: Boolean) {
             ":workload-jvm:build"
         ).buildAndFail().let {
             val errors = it.output.lines().filter { it.startsWith("e: [ksp]") }
-            Assert.assertEquals("e: [ksp] java.lang.Exception: Test Exception in process", errors.first())
+
+            assertThat(errors, hasItem(containsString("Baz.kt:4: Unresolved reference 'Foo'.")))
+            assertThat(errors, hasItem("e: [ksp] java.lang.Exception: Test Exception in process"))
         }
         project.restore("workload-jvm/build.gradle.kts")
     }
@@ -239,7 +244,9 @@ class AnnoOnProperty {
             ":workload-js:build"
         ).buildAndFail().let {
             val errors = it.output.lines().filter { it.startsWith("e: [ksp]") }
-            Assert.assertEquals("e: [ksp] java.lang.Exception: Test Exception in process", errors.first())
+
+            assertThat(errors, hasItem(containsString("Baz.kt:4: Unresolved reference 'Foo'.")))
+            assertThat(errors, hasItem("e: [ksp] java.lang.Exception: Test Exception in process"))
         }
         project.restore("workload-js/build.gradle.kts")
     }
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
@@ -493,6 +493,9 @@ class KotlinSymbolProcessing(
 
         val projectDisposable: Disposable = Disposer.newDisposable("StandaloneAnalysisAPISession.project")
         var kotlinCoreProjectEnvironment: KotlinCoreProjectEnvironment? = null
+        var newKSFiles = emptyList<KSFile>()
+        var codeGenerator: CodeGenerator? = null
+
         try {
             val (analysisAPISession, env, modules) =
                 createAASession(projectDisposable)
@@ -515,7 +518,7 @@ class KotlinSymbolProcessing(
             val allKSFiles =
                 prepareAllKSFiles(env, modules, javaFileManager)
             val anyChangesWildcard = AnyChanges(kspConfig.projectBaseDir)
-            val codeGenerator = CodeGeneratorImpl(
+            codeGenerator = CodeGeneratorImpl(
                 kspConfig.classOutputDir,
                 { if (kspConfig is KSPJvmConfig) kspConfig.javaOutputDir else kspConfig.kotlinOutputDir },
                 kspConfig.kotlinOutputDir,
@@ -540,7 +543,7 @@ class KotlinSymbolProcessing(
                 kspConfig.changedClasses,
             )
             var allDirtyKSFiles = incrementalContext.calcDirtyFiles(allKSFiles).toList()
-            var newKSFiles = allDirtyKSFiles
+            newKSFiles = allDirtyKSFiles
 
             val targetPlatform = ResolverAAImpl.ktModule.targetPlatform
             val processorsRegisteredForUpcomingFeatures = mutableSetOf<SymbolProcessor>()
@@ -654,6 +657,7 @@ class KotlinSymbolProcessing(
 
             // Call onError() or finish()
             if (logger.hasError) {
+                runTypeCheck(newKSFiles, project, logger)
                 processors.forEach(SymbolProcessor::onError)
             } else {
                 processors.forEach(SymbolProcessor::finish)
@@ -671,6 +675,10 @@ class KotlinSymbolProcessing(
 
             dropCaches()
             codeGenerator.closeFiles()
+        } catch (t: Throwable) {
+            val ktFiles = getFilesOnCrash(newKSFiles, kotlinCoreProjectEnvironment?.project)
+            runTypeCheck(ktFiles, logger)
+            throw t
         } finally {
             maybeRunInWriteAction {
                 (kotlinCoreProjectEnvironment?.environment?.jarFileSystem as? CoreJarFileSystem)?.clearHandlersCache()
@@ -681,6 +689,49 @@ class KotlinSymbolProcessing(
         }
 
         return if (logger.hasError) ExitCode.PROCESSING_ERROR else ExitCode.OK
+    }
+
+    companion object {
+        fun List<KSFile>.toKtFiles(project: Project): List<KtFile> {
+            val paths: Set<Path> = this.filter { it.origin == Origin.KOTLIN }.map { File(it.filePath).toPath() }.toSet()
+           return getPsiFilesFromPaths<KtFile>(project, paths)
+        }
+
+        /**
+         * Get Last Round files on execution crash
+         *
+         * Initially trys to return the latest newKSFiles (codeGenerated files)
+         * if it doesn't find any fallback into project files from ktModule. If project is null
+         * return an emptyList
+         *
+         * @param newKSFiles Last round KSFile list
+         * @param project The current project execution
+         */
+        @OptIn(KaExperimentalApi::class, KaPlatformInterface::class)
+        fun getFilesOnCrash(newKSFiles: List<KSFile>, project: Project?): List<KtFile> {
+            if (!newKSFiles.isEmpty()) return newKSFiles.toKtFiles(project!!)
+
+            if (project == null) return emptyList()
+
+            val ktModule = try {
+                ResolverAAImpl.ktModule
+            } catch (_: Throwable) {
+                return emptyList()
+            }
+
+            return ktModule.psiRoots.filterIsInstance<KtFile>()
+        }
+
+        fun runTypeCheck(ktFiles: List<KtFile>, logger: KSPLogger) {
+            try {
+                KspDiagnostics.runTypeCheck(ktFiles, logger)
+            } catch (_: Throwable) {
+                // ignore if type check cannot proceed to avoid masking original failure
+            }
+        }
+
+        fun runTypeCheck(newKSFiles: List<KSFile>, project: Project, logger: KSPLogger) =
+            runTypeCheck(newKSFiles.toKtFiles(project), logger)
     }
 }
 
--- a/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
+++ b/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2026 Google LLC
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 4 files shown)


jaschdoc:
Thanks for opening the PR @alfonsocv12! Do you want to fix the failing integration tests, or is it okay with you if I open another PR?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2022 Google LLC
+ * Copyright 2010-2022 JetBrains s.r.o. and Kotlin Programming Language contributors.
```
jaschdoc: nit: Update to 2026

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.jetbrains.kotlin.psi.KtFile
+
+@OptIn(KaExperimentalApi::class, KaPlatformInterface::class)
+object KspDiagnostics {
+
+    fun runTypeCheck(project: Project?, logger: KSPLogger) {
+        if (project == null) return
+        val ktModule = try {
+            ResolverAAImpl.ktModule
+        } catch (_: Throwable) {
```
jaschdoc: Let's pass the `ktModule` as a parameter to the `runTypeCheck` function instead of accessing the resolver here. This also allows us to do less error handling.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,84 @@
...
+                                ""
+                            }
+                            logger.error("$prefix${diagnostic.defaultMessage}", null)
+                        }
+                    }
+                }
+            }
+        } catch (_: Throwable) {
+            // Ignored if type check cannot proceed to avoid masking original failure
+        }
```
jaschdoc: I think this try-catch should be removed, and instead the function should have a doc comment mentioning that the caller must handle any exception thrown by it. Then we can perform the error handling logic at the call site and decide whether to ignore it or throw.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -671,6 +672,9 @@ class KotlinSymbolProcessing(
 
             dropCaches()
             codeGenerator.closeFiles()
+        } catch (t: Throwable) {
+            KspDiagnostics.runTypeCheck(kotlinCoreProjectEnvironment?.project, logger)
+            throw t
```
jaschdoc, line 681: I do not think we should add this catch-branch. If something fails, it should do so immediately. It seems unlikely an error is caused by invalid Kotlin code.

gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
```diff
@@ -81,6 +81,39 @@ class GradleCompilationTest(isExperimentalPsiResolution: Boolean) {
...
+        }
+
+        class Provider : TestSymbolProcessorProvider({ env -> ErrorReporting(env.logger) })
+
+        testRule.addProvider(Provider::class)
+        val failure = testRule.runner()
+            .withArguments("app:assemble")
+            .buildAndFail()
+        assertThat(failure.output).contains("Fatal KSP processor error!")
+        assertThat(failure.output).contains("UnresolvedClass")
```
jaschdoc: Can we assert something more specific than `UnresolvedClass`?

jaschdoc (changes requested):

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
alfonsocv12: I'm thinking about this case [PublinProblemReporter is not register...](https://github.com/google/ksp/issues/3140), the crash was induce by a syntax error (A missing type). 

The crash should be reported but developers can find what is wrong with there code unblocking themself.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
alfonsocv12: Thanks for the suggestion. 

I think theres a bug in this code `ktModule.psiRoots.filterIsInstance<KtFile>().forEach { ktFiles.add(it) }` is going to get me the initial files, instead of the last round is that correct?

To avoid that from happening instead of passing `ktModule` I'm sending the KtFiles, and I'm getting them from `newKSFiles`. WDYT?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
alfonsocv12: Good idea

gradle-plugin/src/test/kotlin/com/google/devtools/ksp/gradle/GradleCompilationTest.kt
alfonsocv12: done

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
alfonsocv12: done

alfonsocv12:
> Thanks for opening the PR @alfonsocv12! Do you want to fix the failing integration tests, or is it okay with you if I open another PR?

If theres nothing else. It should be fixed, I was getting an assertion error with a expected `first` error message, now is the type check message instead [KMPImplementedIT.kt](https://github.com/google/ksp/pull/3195/changes#diff-1416280789e3ce6343a859347f5536709504addd00d263cd447b2405f03df689)

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
jaschdoc: I see. If we do decide to go with this option, let's only catch `InternalKSPException` instead.
1. If it's an exception that KSP does not produce or knows nothing about, it should be throw without doing anything else since we do not know the consequences of running a type check before exiting
2. A throwable should never be caught as it might signal invalid JVM state.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -493,6 +493,9 @@ class KotlinSymbolProcessing(
 
         val projectDisposable: Disposable = Disposer.newDisposable("StandaloneAnalysisAPISession.project")
         var kotlinCoreProjectEnvironment: KotlinCoreProjectEnvironment? = null
+        var newKSFiles = emptyList<KSFile>()
+        var codeGenerator: CodeGenerator? = null
+
```
jaschdoc, line 498: Could you explain why this is necessary?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -682,6 +690,49 @@ class KotlinSymbolProcessing(
...
+        fun getFilesOnCrash(newKSFiles: List<KSFile>, project: Project?): List<KtFile> {
+            if (!newKSFiles.isEmpty()) return newKSFiles.toKtFiles(project!!)
+
+            if (project == null) return emptyList()
+
+            val ktModule = try {
+                ResolverAAImpl.ktModule
+            } catch (_: Throwable) {
+                return emptyList()
+            }
```
jaschdoc, line 720: For the same reasons as above, I do not think we should catch anything here.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -682,6 +690,49 @@ class KotlinSymbolProcessing(
...
+    companion object {
+        fun List<KSFile>.toKtFiles(project: Project): List<KtFile> {
+            val paths: Set<Path> = this.filter { it.origin == Origin.KOTLIN }.map { File(it.filePath).toPath() }.toSet()
+           return getPsiFilesFromPaths<KtFile>(project, paths)
+        }
+
+        /**
+         * Get Last Round files on execution crash
+         *
+         * Initially trys to return the latest newKSFiles (codeGenerated files)
```
jaschdoc, line 703: nit: typo in `trys`, should be `tries`.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -682,6 +690,49 @@ class KotlinSymbolProcessing(
 
         return if (logger.hasError) ExitCode.PROCESSING_ERROR else ExitCode.OK
     }
+
+    companion object {
+        fun List<KSFile>.toKtFiles(project: Project): List<KtFile> {
+            val paths: Set<Path> = this.filter { it.origin == Origin.KOTLIN }.map { File(it.filePath).toPath() }.toSet()
```
jaschdoc, line 696: Should we also include Java files here, i.e., with `origin == Origin.JAVA`?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
jaschdoc: One more thing, the PluginProblemReporter issue was resolved. I think it's fair to not report any diagnostics in a fatal crash scenario.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -682,6 +690,49 @@ class KotlinSymbolProcessing(
...
+
+            return ktModule.psiRoots.filterIsInstance<KtFile>()
+        }
+
+        fun runTypeCheck(ktFiles: List<KtFile>, logger: KSPLogger) {
+            try {
+                KspDiagnostics.runTypeCheck(ktFiles, logger)
+            } catch (_: Throwable) {
+                // ignore if type check cannot proceed to avoid masking original failure
+            }
```
jaschdoc, line 730: Same as above.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,65 @@
...
+                        val psi = diagnostic.psi
+                        val fileLoc = psi.containingFile?.virtualFile?.path ?: psi.containingFile?.name
+                        val line = psi.containingFile?.viewProvider?.document?.getLineNumber(psi.textOffset)?.plus(1)
+                        val prefix = if (fileLoc != null && line != null) {
+                            "$fileLoc:$line: "
+                        } else if (fileLoc != null) {
+                            "$fileLoc: "
+                        } else {
+                            ""
+                        }
```
jaschdoc, line 58: Nit: Could you extract this into a helper function in `KspDiagnostics`, e.g., `readableFileLocation` or something like that?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
jaschdoc: Yes, believe that's correct.

jaschdoc (commented):
Thank you for the changes. I'm requesting some more changes as I do not think we should catch exceptions and run the type check after an exception has been raised (at least not to begin with). We can see if it becomes necessary later, but for now it introduces too much of a risk IMO.
Additionally, I'm curious, why are we only interested in last-round files and not both user-created files, i.e., not generated by any processor, and the very last round of processing? That way the user get diagnostics if there are problems with their code as well as any resulting transitive errors.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
alfonsocv12: It was needed for the catch but not needed now that is remove

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
alfonsocv12: I think you're right. Remove the typeCheck on Crash for now

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/primary/KMPImplementedIT.kt
```diff
@@ -117,7 +120,9 @@ class KMPImplementedIT(experimentalPsiResolution: Boolean) {
             ":workload-jvm:build"
         ).buildAndFail().let {
             val errors = it.output.lines().filter { it.startsWith("e: [ksp]") }
-            Assert.assertEquals("e: [ksp] java.lang.Exception: Test Exception in process", errors.first())
+
+            assertThat(errors, hasItem(containsString("Baz.kt:4: Unresolved reference 'Foo'.")))
```
jaschdoc, line 124: Is this assertion wrong? Should it be `assertThat(errors, hasItem("e: [ksp] Baz.kt:4: Unresolved reference 'Foo'.")`?

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/primary/KMPImplementedIT.kt
```diff
@@ -239,7 +244,9 @@ class AnnoOnProperty {
             ":workload-js:build"
         ).buildAndFail().let {
             val errors = it.output.lines().filter { it.startsWith("e: [ksp]") }
-            Assert.assertEquals("e: [ksp] java.lang.Exception: Test Exception in process", errors.first())
+
+            assertThat(errors, hasItem(containsString("Baz.kt:4: Unresolved reference 'Foo'.")))
```
jaschdoc, line 248: Same thing here as the other comment.

jaschdoc (commented):
Thanks for updating the PR. There are still failing integration tests. It looks like either the assertions are causing the test failures, or the test failure is a real defect. I know that the assertions are checking for substrings but let's just assert the full expected output for the diagnostic line so we know that its shape is obviously correct.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
alfonsocv12: Done

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/primary/KMPImplementedIT.kt
alfonsocv12: done

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/primary/KMPImplementedIT.kt
alfonsocv12: done

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
alfonsocv12: Sorry for the delay it was a bit more complex than I though to add Java support initially I try to use Javac/Turbine but the current sources types from kotlin files uses in java files are always unresolved until full compilation.

The final approach was using PSI wdyt?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,134 @@
...
+
+            messages.forEach { logger.error(it, null) }
+        }
+    }
+}
+
+class UnresolveReferenceCollector(
+    private val messages: MutableSet<String>
+) : JavaRecursiveElementWalkingVisitor() {
+    // Javadoc references are not resolved by the compiler, don't descend inot them.
```
jaschdoc, line 90: Nit: typo in `inot`, should be `into`

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -682,6 +684,20 @@ class KotlinSymbolProcessing(
 
         return if (logger.hasError) ExitCode.PROCESSING_ERROR else ExitCode.OK
     }
+
+    companion object {
+        fun List<KSFile>.toKtFiles(project: Project): List<KtFile> = getPsiFilesFromPaths<KtFile>(project, this.toPathSet())
+
+        fun List<KSFile>.toPsiJavaFiles(project: Project): List<PsiJavaFile> = getPsiFilesFromPaths<PsiJavaFile>(project, this.toPathSet())
+
+        fun List<KSFile>.toPathSet(): Set<Path> = this.map {File(it.filePath).toPath() }.toSet()
```
jaschdoc, line 693: Nit: Formatting of the lambda is off.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KotlinSymbolProcessing.kt
```diff
@@ -682,6 +684,20 @@ class KotlinSymbolProcessing(
...
+
+        fun runTypeCheck(newKSFiles: List<KSFile>, project: Project, logger: KSPLogger) {
+            val (kotlinKSFiles, javaKSFiles) = newKSFiles
+                .partition { it.origin == Origin.KOTLIN }
```
jaschdoc, line 697: Nit: This could just be one line.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,134 @@
...
+     * @param logger If analysis finds type errors it's going to log them as errors
+     */
+    fun runTypeCheck(psiJavaFiles: Collection<PsiJavaFile>, ktFiles: Collection<KtFile>, logger: KSPLogger) {
+        if (psiJavaFiles.isEmpty() && ktFiles.isEmpty()) return
+
+        runJava(psiJavaFiles, logger)
+        runKotlin(ktFiles, logger)
+    }
+
+    fun runKotlin(ktFiles: Collection<KtFile>, logger: KSPLogger) {
```
jaschdoc, line 58: Nit: This could be private and be called `runKotlinTypeCheck`. This same comment applies to the `runJava` function.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Run typeChecking on a list of Kotlin Files
+     *
+     * Some of the APIs in use could throw exceptions so you need to handler errors accordingly.
+     *
+     * @param ktFiles List of files to run typecheck
+     * @param logger If analysis finds type errors it's going to log them as errors
+     */
+    fun runTypeCheck(psiJavaFiles: Collection<PsiJavaFile>, ktFiles: Collection<KtFile>, logger: KSPLogger) {
```
jaschdoc, line 51: Nit: I think it's better to have a "pure" function that returns a list of messages that can then be logged by the caller.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,134 @@
...
+            } else {
+                file.accept(UnresolveReferenceCollector(messages))
+            }
+
+            messages.forEach { logger.error(it, null) }
+        }
+    }
+}
+
+class UnresolveReferenceCollector(
```
jaschdoc, line 87: Nit: We should probably write some brief documentation for this class.

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,134 @@
...
+        super.visitReferenceElement(reference)
+        reference.collectIfUnresolved()
+    }
+
+    override fun visitReferenceExpression(expression: PsiReferenceExpression) {
+        super.visitReferenceExpression(expression)
+        expression.collectIfUnresolved()
+    }
+
+    private fun PsiJavaCodeReferenceElement.isResolved(): Boolean =
```
jaschdoc, line 103: Nit: Isn't the name `isResolvable` more appropriate?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,134 @@
...
+        this is PsiMethodReferenceExpression || resolve() != null || hasUnresolvedQualifier() ||
+            this.getName() == null
+
+    private fun PsiJavaCodeReferenceElement.getName(): String? = referenceName ?: text
```
jaschdoc, line 107: Is the `text` property the text that occurs in the source, i.e., just the referenced name? That may either be resolvable or not but it seems that regardless of that the text is always available. Could you clarify if this is the case or not?

kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/impl/KspDiagnostics.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    private fun PsiJavaCodeReferenceElement.collectIfUnresolved() = if (!this.isResolved()) {
+        messages.add((referenceNameElement ?: this).withLocation("Unresolved reference '${this.getName()}'"))
+    } else null
+
+    private fun PsiJavaCodeReferenceElement.hasUnresolvedQualifier(): Boolean =
+        when (val qualifier = qualifier) {
+            null -> false
+            is PsiJavaCodeReferenceElement -> qualifier.resolve() == null
+            is PsiExpression -> qualifier.type == null
```
jaschdoc, line 117: This suggests to me that it is not only collecting unresolved references but also checking whether a type can be assigned to the expression in which case it is also a type error (which may be caused by an unresolvable reference).

jaschdoc (approved):
Hey thanks for updating the PR again. I have left some a few questions and a lot of nitpicks. I'm approving the PR and will clean up the minor details afterwards, but I'm still curious about the questions I raised. Thanks again!

PR #3204, opened 2026-09-11
Fix backing field related JVM modifiers in `Resolver.effectiveJavaModifiers`

### LINKED ISSUE #3155: Backing field modifiers might be incorrect
For the upcoming backing fields feature, it should be double checked what the correct modifiers are.
- Do they always inherit the same JVM modifiers as the owning property?
- They are always final in Kotlin but do they have other modifiers?
- What about Java fields? KSP models Java fields as properties, but some of the features seem to apply to the field instead of the property.

jaschdoc:
Closes https://github.com/google/ksp/issues/3155


jaschdoc:
Superseded by https://github.com/google/ksp/pull/3228

PR #3210, opened 2026-09-15
KSP2: reuse processor classloaders across modules within a build

hfmehmed:
`KspAAWorkerAction` constructs a fresh `URLClassLoader` over the processor
classpath for every module and closes it when that module finishes. The same
processor jars are therefore loaded and compiled once per module.

This PR adds an opt-in cache of processor classloaders keyed by the KSP classpath
plus the processor classpath. It lives alongside the existing isolated classloader
cache and is cleared by the same build service, so its lifetime is still bounded
by the build.

### Results

Measured on the [Metro 500-module Dagger/KSP benchmark][1], ABI-change scenario,
via `gradle-profiler` with 3 warm-ups and 6 measured iterations. The same ksp binary is used - one of them has a flag set to enable the cache and the other represents the baseline (current state)

| | Median | Mean | Range |
|---|---:|---:|---:|
| Cache OFF | 85.58s | 86.75s | 84.52 – 91.52s |
| Cache ON | 57.74s | 57.69s | 55.30 – 59.97s |
| **Delta** | **−27.84s (−32.5%)** | **−29.07s (−33.5%)** | — |

**Overlap between the two states: NO** - 24.55s difference between the slowest cache-ON build (59.97s) and the fastest cache-OFF build (84.52s).

Screenshot 2026-09-15 at 18 05 26

[1]: https://zacsweers.github.io/metro/latest/performance/


gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/IsolatedClassLoaderCacheBuildService.kt
```diff
@@ -25,15 +25,29 @@ import java.util.concurrent.ConcurrentHashMap
...
 
+    /**
+     * Cache of processor classloaders, keyed by the KSP classpath plus the processor classpath.
+     *
+     * In large multi-module builds the same processor jars are used by many modules. Creating (and
+     * discarding) a [URLClassLoader] per module forces those classes to be re-loaded, re-verified
+     * and re-JITted for every module, which is pure overhead. Reusing the loader keeps the JIT
+     * profile warm across modules.
+     *
+     * This is opt-in via the `ksp.classloader.cache.processors` Gradle property, because reusing a
```
jaschdoc, line 36: Can we document this property in the README as well?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/IsolatedClassLoaderCacheBuildService.kt
```diff
@@ -25,15 +25,29 @@ import java.util.concurrent.ConcurrentHashMap
...
+     * loader also extends the lifetime of any static state held by processors.
+     */
+    val processorCache = ConcurrentHashMap<String, URLClassLoader>()
+
     fun clear() {
-        cache.values.forEach { classLoader ->
+        (cache.values + processorCache.values).forEach { classLoader ->
```
jaschdoc: Nit: maybe we can extract a variable here? `val classLoaders = cache.values + processorCache.values`

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -743,13 +759,23 @@ abstract class KspAAWorkerAction : WorkAction<KspAAWorkParameter> {
         // TODO: support incremental processing.
         gradleCfg.outputBaseDir.get().asFile.deleteRecursively()
 
-        val processorClassloader = URLClassLoader(
+        val cacheProcessorLoader = gradleCfg.cacheProcessorClassloader.get()
+        val processorKey = key + "|processors|" +
```
jaschdoc: Is the `|processors|` string just a delimiter here or what is its purpose?

integration-tests/src/test/resources/incremental-annotation-argument-class-references/processor/src/main/kotlin/TestProcessor.kt
```diff
@@ -28,6 +28,14 @@ import com.google.devtools.ksp.symbol.KSNode
...
 
 class TestProcessor(val environment: SymbolProcessorEnvironment) : SymbolProcessor {
+    init {
+        // Identifies the classloader this processor was loaded from. Modules that share a cached
+        // processor classloader report the same id; modules with their own loader report different
+        // ids. Deliberately uses a prefix distinct from [log]'s so that tests asserting on
+        // "[TestProcessor]" lines are unaffected.
+        println("[KSPLoaderId] ${System.identityHashCode(TestProcessor::class.java.classLoader)}")
+    }
+
```
jaschdoc, line 38: Should we make this behavior toggle-able such that it only prints this information given a flag? WDYT?

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
```diff
@@ -743,13 +759,23 @@ abstract class KspAAWorkerAction : WorkAction<KspAAWorkParameter> {
...
             gradleCfg.processorClasspath.files.map { it.toURI().toURL() }.toTypedArray(),
             isolatedClassLoader
         )
+        val processorClassloader = if (cacheProcessorLoader) {
+            IsolatedClassLoaderCache.processorCache.computeIfAbsent(processorKey) {
+                newProcessorClassLoader()
+            }
+        } else {
+            newProcessorClassLoader()
+        }
```
jaschdoc: Nit: Should we refactor this if-else expression into a function and lift the `newProcessorClassLoader` function as well? Maybe rename it to `createProcessorClassLoader`?

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/secondary/IncrementalAnnotationArgumentClassReferences.kt
```diff
@@ -91,4 +92,43 @@ class IncrementalAnnotationArgumentClassReferences(experimentalPsiResolution: Bo
             )
         }
     }
+
+    /**
+     * Both `:upstream` and `:downstream` apply the same processor, so they resolve an identical
+     * processor classpath. Without the cache each gets its own processor classloader; with the
+     * cache enabled they must share one.
+     *
+     * Also asserts that enabling the cache does not change what the processor does.
```
jaschdoc, line 101: Personally, this sounds like it should be two separate tests, but I understand that for performance reasons, it much faster to just compare the outputs so we only run two builds instead of four. Should we add a brief comment about that as well?

jaschdoc (commented):
Mostly nits and questions. Looks like a nice optimization!

integration-tests/src/test/resources/incremental-annotation-argument-class-references/processor/src/main/kotlin/TestProcessor.kt
hfmehmed: I am not sure it is worth it. This will only affect test logs which should be mostly fine.

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
hfmehmed: Yes its a delimeter. I wanted to make that kspCP=[a,b] + procCP=[c] can't collide with kspCP=[a] + procCP=[b,c] for example

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/IsolatedClassLoaderCacheBuildService.kt
hfmehmed: Yes good point. Done

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
hfmehmed: Done

integration-tests/src/test/kotlin/com/google/devtools/ksp/test/secondary/IncrementalAnnotationArgumentClassReferences.kt
hfmehmed: Added a comment

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/IsolatedClassLoaderCacheBuildService.kt
hfmehmed: Done

gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt
jaschdoc: Good point. I understand.

jaschdoc (approved):
Thanks, looks good!

PR #3213, opened 2026-09-17
Fix performance of `KSClassDeclaration.getAllSuperTypes`

jaschdoc:
This change brings down the complexity from what looks like exponential to a polynomial time function by maintaining a visited set. Additionally also guards cyclic type aliases (although this cannot occur, we do not want non-termination if it were to happen).


api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -241,29 +266,28 @@ fun KSClassDeclaration.getAllSuperTypes(): Sequence<KSType> {
...
-                        else ->
-                            throw InternalKSPException(
-                                "Unhandled super type kind",
-                                it.location,
-                                it.javaClass,
-                            )
-                    }
-                }
-        )
-        .distinct()
```
troelsbjerre, line 266: The "fixed" implementation no longer returns a unique set of supertypes, as is evident in the test expectation as well. This is likely to break clients.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -241,29 +266,28 @@ fun KSClassDeclaration.getAllSuperTypes(): Sequence<KSType> {
...
+    val resolved = superTypes.map { it.resolve() }.toList()
```
troelsbjerre: This `.toList()` breaks the sequence laziness. The purpose of it is to make resolved repeatable, but I think it would be more idiomatic to write a version using `sequence {/*hierarchy traversal an uniqueness check */}` that is truly lazy.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -176,12 +176,25 @@ fun KSNode.validate(
...
-    return if (resolvedType is KSTypeAlias) {
-        resolvedType.findActualType()
-    } else {
-        resolvedType as KSClassDeclaration
+fun KSTypeAlias.findActualType(): KSClassDeclaration = expandTypeAlias(mutableSetOf())
+
+private fun KSTypeAlias.expandTypeAlias(visited: MutableSet<KSDeclaration>): KSClassDeclaration {
+    if (!visited.add(this)) {
+        throw InternalKSPException(
+            "Circular type alias declaration: '${qualifiedName?.asString()}'",
```
troelsbjerre, line 184: No test covering this

kotlin-analysis-api/testData/diamondInheritance.kt
```diff
@@ -0,0 +1,177 @@
...
+// L25a
+// L25b
+// L26a
+// L26b
+// L27a
+// L27b
+// L28a
+// L28b
+// L29a
+// L29b
```
troelsbjerre: Do not emit duplicate supertypes

troelsbjerre (changes requested):

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
jaschdoc: This cannot happen in source code, but theoretically if you handcrafted the Kotlin metadata then it could.

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
jaschdoc: Fixed

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
jaschdoc: Fixed

kotlin-analysis-api/testData/diamondInheritance.kt
jaschdoc: Fixed

api/src/main/kotlin/com/google/devtools/ksp/utils.kt
```diff
@@ -225,10 +238,22 @@ internal fun KSDeclaration.isKotlinBackingField(): Boolean =
  * get all super types for a class declaration Calling [getAllSuperTypes] requires type resolution
  * therefore is expensive and should be avoided if possible.
  */
-fun KSClassDeclaration.getAllSuperTypes(): Sequence<KSType> {
+fun KSClassDeclaration.getAllSuperTypes(): Sequence<KSType> =
+    getAllSuperTypes(mutableSetOf(), mutableSetOf())
```
troelsbjerre: In `main`, we have that `KSClassDeclaration.getAllSuperTypes()` is repeatable. If that is a required property, this line could be changed to
```
        Sequence { getAllSuperTypes(mutableSetOf(), mutableSetOf()).iterator() }                                        
```

jaschdoc: Could you clarify what you mean by repeatable? Should it be a [`lazyMemoizedSequence`](https://github.com/google/ksp/blob/41d6ce378a1a73930b4bf38b65d2e956c8171f79/kotlin-analysis-api/src/main/kotlin/com/google/devtools/ksp/common/PsiUtils.kt#L99)?

